PR #25, opened 2023-04-12
Readme cleanup

gabe-l-hart:
## Description

This PR partially cleans up the README and removes misleading examples

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,26 +1,26 @@
 # Caikit
 
-Caikit is an AI toolkit that enables users to manage AI models through a set of developer friendly APIs. It provides a consistent format for AI model deployment, consumption and evolution.
+Caikit is an AI toolkit that enables users to manage models through a set of developer friendly APIs. It provides a consistent format for creating and using AI models against a wide variety of data domains and tasks.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,128 +1,51 @@
...
 
-Caikit provides an abstraction layer for application developers where they can consume an AI model through APIs independent of understanding the data form of the model. In other words, the input and output to the model is in a format which is easily programmable and does not require data transformations. This facilitates the model and the application to evolve independently of each other.
+Caikit provides an abstraction layer for application developers where they can consume a AI models through APIs independent of understanding the data form of the model. In other words, the input and output to the model is in a format which is easily programmable and does not require data transformations. This facilitates the model and the application to evolve independently of each other.
```
evaline-ju, line 21: find-replace typo? I think the original with singular references made sense here or we should pluralize all?
```suggestion
Caikit provides an abstraction layer for application developers where they can consume an AI model through APIs independent of understanding the data form of the model. In other words, the input and output to the model is in a format which is easily programmable and does not require data transformations. This facilitates the model and the application to evolve independently of each other.
```

PR #28, opened 2023-04-13
:bug: Flit uses classifiers for license

evaline-ju:
Only `classifiers` are parsed - we will keep `license` in case pep changes


joerunde:
For reference here is the PEP about fixing license metadata: https://peps.python.org/pep-0639/

Looks like they want to deprecate both the `license` field and licenses in `classifiers` and replace with `License-Expression` and `License-File` fields

joerunde (approved):

Issue #8, opened 2023-04-11
Support runtime service generation config for `modules` and `task_types`


aluu317:
## Is your feature request related to a problem? Please describe.

ServicePackageFactory currently does not respect the modules and task types requested in `config.yml`.

## Describe the solution you'd like

Add support in the service factory to respect the configs for:
```
service_generation:
  module_types:
    included: ["blocks", "workflows"]
  modules:
    excluded: []
  task_types:
    excluded: []
```
`modules` and `task_types` could also have an include list as well that excludes everything else

## Additional context

Add any other context about the feature request here.

aluu317:
Closing this issue as the PR was merged. @tharapalanivel as discussed outside of this Issue, we need to revisit the service-level configs for inclusion/exclusion more and can open another Issue if needed!

PR #32, opened 2023-04-13
Allow ServicePackageFactory to read config's lists of exclusions and inclusions

aluu317:
Issue https://github.com/caikit/caikit/issues/8

This PR allows for users to specify an `included` and `excluded` fields in `service_generation` that ServicePackageFactory will respect.
Example:

```
service_generation:
    enabled: true
    modules:
        included: ["00110203-baad-beef-0809-0a0b02dd0e0f"],
        excluded: ["00af2203-0405-0607-0263-0a0b02dd0c2f"]
    task_types:
        included: ["other_task"]
        excluded: ["sample_task"]
```

- If a module happens to be on both `included` and `excluded` lists, it's excluded
- If `included` and `excluded` field is `[]` or `null`, it equates to no-op

### DIFF
--- a/caikit/runtime/service_factory.py
+++ b/caikit/runtime/service_factory.py
@@ -261,8 +261,8 @@ class ServicePackageFactory:
         )
         included_modules = (
             config_parser.service_generation
-            and config_parser.service_generation.modules
-            and config_parser.service_generation.modules.included
+            and config_parser.service_generation.module_guids
+            and config_parser.service_generation.module_guids.included
         )
 
         # Check config for any exclusions
@@ -273,8 +273,8 @@ class ServicePackageFactory:
         )
         excluded_modules = (
             config_parser.service_generation
-            and config_parser.service_generation.modules
-            and config_parser.service_generation.modules.excluded
+            and config_parser.service_generation.module_guids
+            and config_parser.service_generation.module_guids.excluded
         )
 
         for ck_module in modules:
--- a/caikit/runtime/service_factory.py
+++ b/caikit/runtime/service_factory.py
@@ -15,7 +15,7 @@
 # Standard
 from enum import Enum
 from types import ModuleType
-from typing import Callable, Dict, List, Type
+from typing import Callable, Dict, List, Set, Type
 import dataclasses
 import inspect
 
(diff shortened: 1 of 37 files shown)


evaline-ju (commented):
I think we would want to update the defaults in `config.yml` to account for the added keys here?

caikit/runtime/service_factory.py
```diff
@@ -242,6 +243,73 @@ def get_service_package(
             )
 
     # Implementation details for pure python service packages #
+    @staticmethod
+    def _get_and_filter_modules(config_parser: ConfigParser, lib: str):
```
evaline-ju, line 247: missing return type annotation?

tests/runtime/test_service_factory.py
```diff
@@ -165,3 +167,130 @@ def test_get_service_proto_module():
...
+        assert len(clean_modules) == 1
+        assert "sample_task" not in str(clean_modules)
+
+
+def test_get_and_filter_modules_respects_excluded_modules():
+    assert "InnerBlock" in str(MODULE_LIST)
+    with temp_config_parser(
+        {
+            "service_generation": {
+                "modules": {"excluded": ["00110203-baad-beef-0809-0a0b02dd0e0f"]}
```
evaline-ju, line 197: general usability q as someone who hasn't looked at runtime for a while - are users generally expected to know the guids here?  or how are other config readers supposed to know what module each guid refers to i.e. reading this - how ""00110203-baad-beef-0809-0a0b02dd0e0f" is `InnerBlock` without grepping?

joerunde: A quick fix here might be to change `modules` to `module_guids` so that it's clear these are lists of guids, and leave this open to extension later to add like 
```
module_paths:
    included: ["my_lib.blocks.my_task.some_partial_name*"]
    # or...
    excluded: ["*ESAHierachical*", "my_lib.blocks.some_task.internal.*"]
```

aluu317: Updated to use `module_guids`. Yea I agree with @evaline-ju . Guids arent super readable for usability, I'm unsure of why we chose that when we first started with runtime...

joerunde: Probably because we needed to get things out the door ⏩⏩⏩ and the simplest option was to match guids instead of doing anything more complex

joerunde (approved):
lgtm, thanks @aluu317 !

PR #34, opened 2023-04-17
:fire: Remove performance runner and iterators

evaline-ju:
This version of the performance runner has not been validated on various domains that caikit will intend to eventually support. At this time, we want to remove this code from being maintained.


alex-jw-brooks (approved):
Looks good to me! Ship it!

joerunde (approved):
nice test coverage improvement 👍

Issue #17, opened 2023-04-12
Consolidate config


joerunde:
## Description

As a library extending caikit (or somebody deploying a container of that library) I would like a single file defining all of my config.

(Instead of one file for core config and one for runtime config)

## Discussion

The `caikit.core.config` and `caikit.runtime.config` packages developed separately, and now could use merging.
It would be great to have 
- only a single package deal with reading config
- only a single config yml baked into the caikit release
- only a single config file required to be written by the extension library
- the same `CONFIG_FILES` sources override from the runtime config package so that configs can be easily supplied by files at runime

The `environment` overrides (e.g. test.config.name vs. prod.config.name, _not_ environment variable overrides) could be considered for deletion, it's unclear if they are actually useful

## Acceptance Criteria

- [ ] Unit tests cover new/changed code
- [ ] Examples build against new/changed code
- [ ] READMEs are updated
- [ ] Type of [semantic version](https://semver.org/) change is identified

PR #36, opened 2023-04-17
:fire: remove strict rpc mode

joerunde:
Doing some config cleanup / feature pruning in support of https://github.com/caikit/caikit/issues/17

Before we provide one single config I wanted to remove as much as possible from the existing ones. This feature was added for one of the very first iterations of caikit to raise exceptions when models were called incorrectly. We have better ways of handling this now that services are generated from `caikit` itself instead of being run by hand-written proto files


evaline-ju (approved):
LGTM

PR #38, opened 2023-04-17
🔥  remove grpc sleep interval

joerunde:
legacy config that wasn't deleted when we deleted the server's busy loop


evaline-ju (approved):
LGTM

PR #40, opened 2023-04-17
⬆️ expand protobuf range even further

prashantgupta24:
To add compatibility against `tensorflow` versions in which they specify `protobuf - required: >=3.9.2,<3.20`


joerunde (approved):
nice!

PR #42, opened 2023-04-17
:fire: Remove test model and make comments more library-generic

evaline-ju:
- Make comments more library generic
- Remove a test model and update one test that references it


gabe-l-hart (approved):
Looks good!

Issue #13, opened 2023-04-12
Document initial architectural decisions


joerunde:
## Is your feature request related to a problem? Please describe.

During the initial implementation of caikit we iterated through quite a few design choices, and landed on some important ones after trial and error. We should document those for reference.

## Describe the solution you'd like

A few [ADRs](https://adr.github.io/) about key decisions that have already been made

One or more documents about general architectural principles that we consider when making architectural decisions, to guide future work 

## Describe alternatives you've considered

Not documenting lmao

## Additional context

Possible decisions to consider:
- Using protobuf as the base backend for all data models, open to extension
- JTD as the single source of schema definition for data models
- Runtime and interfaces co-located with `caikit`
- `train` (classmethod) instead of `fit` (instance method)
- `load` and `save` semantics
- distributed backends

joerunde:
ADR buckets:

## Module things
- Module type signatures
- Tasks
- Module backends
- module base class hierarchy (blocks, workflows, resources)

## Model? Things
- Model Encapsulation
- Model Versioning

## Library meta-architecture stuff
- config
- logging
- monorepo

## Data model
- dataclasses as data models
- data model concept
- data model backends
- common protobuf backend

somewhere in between is:
- Data Stream + Data Stream Source

## Runtime
- Service Generation
- Global Handlers
- Runtime as an actor in external system (model runtime, model train)

gkumbhat:
I'll start with 3 today and could take on 2 next.

PR #45, opened 2023-04-18
:memo: Data model ADR

evaline-ju:
For #13

### DIFF
--- a/docs/adrs/010-data-model-definition.md
+++ b/docs/adrs/010-data-model-definition.md
@@ -22,4 +22,4 @@ Accepted
 - Caikit users may face one extra level of indirection if they want to access mathematical representations of their data models results (e.g. arrays).
 - Caikit will provide the same data model per AI task [TODO: link to task ADR].
 - Data models will be backed by a common backend [TODO: link to backends, protobuf ADRs].
-- Caikit model authors will have to translate model inputs and outputs to work with data models. This has implications on usability and stability.
+- AI model authors will have to translate model inputs and outputs to work with data models. This has implications on usability and stability.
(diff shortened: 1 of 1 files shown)


markstur (commented):
Just wondering... I tend to think of the data models as both input and/or output to Predict (I'm ignoring train for now).  I like using simpler types as input when possible.  So does it make sense to clarify if "data of concrete models" is meant to imply inference output and/or training input...etc?  It's a little vague to me right now.  Maybe the word "results" should be used more. Just curious if that fits this ADR.

evaline-ju:
@markstur Thanks for the input. I've tried to rephrase to be a bit clearer. The wording was likely too vague in an attempt at generality, but concretely I was imagining we could use data models for any of inference input (say a raw document structure) or output (classification result) or training input (any sort of training record). I tend to associate the word “result” with “output,” which would be a bit constraining.

docs/adrs/010-data-model-definition.md
```diff
@@ -20,5 +20,6 @@ Accepted
 
 - Caikit users will be able to access model information through data models
 - Caikit users may face one extra level of indirection if they want to access mathematical representations of their data models results (e.g. arrays).
-- Caikit will provide the same data model per task [TODO: link to task ADR]. 
+- Caikit will provide the same data model per AI task [TODO: link to task ADR].
 - Data models will be backed by a common backend [TODO: link to backends, protobuf ADRs].
+- Caikit model authors will have to translate model inputs and outputs to work with data models. This has implications on usability and stability.
```
markstur, line 25: I think s/Caikit model authors/Caikit module authors/

evaline-ju: I was trying to go for some consistency with the README personas provided here: https://github.com/caikit/caikit#user-profiles switched s/Caikit/AI/

hickeyma: I think model makes more sense as we are looking at the profile from the high level go AI models.

joerunde (approved):

Issue #15, opened 2023-04-12
Get to 100% test coverage


joerunde:
## Describe the bug

We don't have 100% line coverage in unit tests.
We really like having 100% test coverage, so that we can ensure all future changes at least have tests that invoke all lines of code.

We've already caught bugs in some modules by requiring 100% coverage, and keeping it once you have it is not hard

## Platform

every platform

## Sample Code

`pytest tests --cov --cov-fail-under=100`

## Expected behavior

passing tests 😎 

## Observed behavior

~80% coverage

## Additional context

Coverage reports can be found in the `htmlcov` directory after running tests

```
tox -e 3.9
open htmlcov/index.html
```

PR #46, opened 2023-04-19
test: Add more sig_parsing tests

aluu317:
For Issue #15 

What I assumed:
- We do not need "syntax_doc" and "raw_document" as KNOWN_ARG_TYPES
- I added "SampleOutputType" and "OtherOutputType" as KNOWN_OUTPUT_TYPES

This PR added more tests to increase the coverage for parsers.py

Before: 69%
Screen Shot 2023-04-19 at 11 26 29 AM
After this PR: 100%
Screen Shot 2023-04-19 at 11 33 02 AM

### DIFF
--- a/caikit/runtime/service_generation/signature_parsing/parsers.py
+++ b/caikit/runtime/service_generation/signature_parsing/parsers.py
@@ -33,11 +33,6 @@ KNOWN_ARG_TYPES = {
     "producer_id": "ProducerId",
 }
 
-KNOWN_OUTPUT_TYPES = {
-    "SampleTaskPredict": "SampleOutputType",
-    "OtherTaskPredict": "OtherOutputType",
-}
-
 
 @alog.logged_function(log.debug2)
 def get_output_type_name(
@@ -68,14 +63,11 @@ def get_output_type_name(
     if type_from_docstring:
         return type_from_docstring
 
-    # Check based on naming conventions and then known output types
-    module_parts = module_class.__module__.split(".")
-    log.debug3("Parent module parts for %s: %s", module_class.__name__, module_parts)
-
-    # TODO: this part needs a test (or consider deleting and say user needs to specify output type)
-    class_name = _snake_to_camel(module_parts[2]) + "Predict"
-    return _get_dm_type_from_name(class_name) or _get_dm_type_from_name(
-        KNOWN_OUTPUT_TYPES.get(class_name)
+    # If we get here, it means no annotation or docstring for type was provided
+    log.warning(
+        "Could not deduct output type from function %s for module class %s.",
+        fn_signature,
+        module_class.__name__,
     )
 
 
--- a/caikit/runtime/service_generation/signature_parsing/parsers.py
+++ b/caikit/runtime/service_generation/signature_parsing/parsers.py
@@ -30,12 +30,13 @@ log = alog.use_channel("SIG-PARSING")
 
 # Constants ##################################
 KNOWN_ARG_TYPES = {
-    "syntax_doc": "SyntaxPrediction",
     "producer_id": "ProducerId",
-    "raw_document": "RawDocument",
 }
 
-KNOWN_OUTPUT_TYPES = {}
+KNOWN_OUTPUT_TYPES = {
+    "SampleTaskPredict": "SampleOutputType",
+    "OtherTaskPredict": "OtherOutputType",
+}
 
 
 @alog.logged_function(log.debug2)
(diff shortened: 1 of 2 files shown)


caikit/runtime/service_generation/signature_parsing/parsers.py
```diff
@@ -79,7 +73,7 @@ def get_output_type_name(
     log.debug3("Parent module parts for %s: %s", module_class.__name__, module_parts)
 
     # TODO: this part needs a test (or consider deleting and say user needs to specify output type)
-    class_name = _snake_to_camel(module_parts[2]) + "Prediction"
+    class_name = _snake_to_camel(module_parts[2]) + "Predict"
```
joerunde, line 76: 🤔 , so changing this here would actually affect any users relying on this functionality.

I think I'm leaning towards just cutting it entirely, this was one of those hacks to get up and running but users should be specifying return types at least in the docstring or as a type annotation.

caikit/runtime/service_generation/signature_parsing/parsers.py
```diff
@@ -30,12 +30,13 @@
...
-    "syntax_doc": "SyntaxPrediction",
     "producer_id": "ProducerId",
-    "raw_document": "RawDocument",
 }
 
-KNOWN_OUTPUT_TYPES = {}
+KNOWN_OUTPUT_TYPES = {
+    "SampleTaskPredict": "SampleOutputType",
+    "OtherTaskPredict": "OtherOutputType",
+}
```
joerunde, line 39: Let's make the call on keeping this at all, and if so it seems like a good time to make it configurable so that we don't leak test fixture names into the source code

caikit/runtime/service_generation/signature_parsing/parsers.py
```diff
@@ -60,12 +60,6 @@ def get_output_type_name(
                     fn_signature.return_annotation,
                 )
                 return module_class
-
-            log.debug(
-                "Adding %s to list of candidate type names to search for concrete types:",
-                fn_signature.return_annotation,
-            )
-            extra_candidate_names.append(fn_signature.return_annotation)
```
joerunde, line 68: Oh shoot did we delete the section where we used `extra_candidate_names`? oops!

I think we'll want a test that does like:
```
def run(stuff) -> "SomeDataModelType":
```
as opposed to
```
def run(stuff) -> SomeDataModelType:
```
and make sure we can still pick up the name from a string

joerunde: (But I would also be open to opening a quick lil bug issue about it if that test fails, and fixing in another PR)

tests/runtime/service_generation/signature_parsing/test_parsers.py
```diff
@@ -102,6 +126,59 @@ def _run(self, foo):
...
+        pass
+
+    assert get_argument_types(_run_with_docstring)["input_1"] == Optional[str]
+
+    def _run_with_known_dm_type(sample_input_type):
+        pass
+
+    assert (
+        get_argument_types(_run_with_known_dm_type)["sample_input_type"]
+        == sample_lib.data_model.SampleInputType
```
joerunde, line 152: wait... how does this actually work?

I'm a bit confused on where we configure that the name `sample_input_type` -> `dm.SampleInputType`. Is this from `_get_dm_type_from_name`?

tests/runtime/service_generation/signature_parsing/test_parsers.py
```diff
@@ -102,6 +126,59 @@ def _run(self, foo):
     assert get_argument_types(_run)["foo"] == sample_lib.data_model.SampleInputType
 
 
+def test_get_args_with_no_annotation():
+    """Check that we get arguments with known arg type supplied"""
```
joerunde, line 130: copy-paste docstring from below?

tests/runtime/service_generation/signature_parsing/test_parsers.py
aluu317: Yes. We do `sample_input_type` converted snake to camel case `SampleInputType` and then get_dm_type_from_name . Gabe magic

tests/runtime/service_generation/signature_parsing/test_parsers.py
aluu317: you got me :grin:

tests/runtime/service_generation/signature_parsing/test_parsers.py
joerunde: 🪄🪄🪄

joerunde (approved):
awwwww yiss

PR #48, opened 2023-04-19
:fire: Remove extension utils

evaline-ju:
The extension utils here support caikit extensions like plugins to an existing caikit library, whereas going forward we likely will only want "extension" support in the sense that an extension library is derived from a caikit library with additional/new implementations.

Supports #17 since in the course of making the config consolidated we also wanted to make it more robust. The extension utils depend on the `lib_config` version that a library has to provide, whereas this information should be more directly available by an imported library.


joerunde (approved):
🔥🔥🔥

Issue #47, opened 2023-04-19
Union of return types throws an exception


prashantgupta24:
## Describe the bug

If I have a run function defined like this:
```
def run(self, sample_input: SampleInputType) -> Union[OtherOutputType, str]:
```
I get an exception:
```
AttributeError: '_SpecialForm' object has no attribute 'get_proto_class'
```

## Platform

Please provide details about the environment you are using, including the following:

- Interpreter version: 3.9
- Library version: 0.1.3

## Sample Code

Another example:
```
  def run(self, sample_input: SampleInputType) -> str:
```
gives me:
```
AttributeError: type object 'str' has no attribute 'get_proto_class'
```
**Note: This has been deemed invalid since we do not want to support return of primitive types through modules.**

## Expected behavior

The introspection logic should iterate through the `Union` defined and pick the `data model` type from it.

## Observed behavior

```
self = typing.Union[sample_lib.data_model.sample.OtherOutputType, str], attr = 'get_proto_class'

    def __getattr__(self, attr):
        # We are careful for copy and pickle.
        # Also for simplicity we don't relay any dunder names
        if '__origin__' in self.__dict__ and not _is_dunder(attr):
>           return getattr(self.__origin__, attr)
E           AttributeError: '_SpecialForm' object has no attribute 'get_proto_class'
```

## Additional context

While returning the `return_type` from `serializers.py`, we should perhaps call the `py_type_to_proto_type` method (that existed before) to pick the `data model` from the `Union` type - or think about reusing some code from it!

joerunde:
@prashantgupta24 we should definitely _not_ call `py_type_to_proto_type`! All the conversions to proto types are now handled by `@dataobject` instead. We're trying to get the introspection bits down to worrying about python types only.

Sounds good re: picking the data model type out of a union, though users would need to be very careful about models that can potentially return non-data-model things, since those will 🤮 if invoked via the runtime

prashantgupta24:
@joerunde yep definitely not the full function as is, but a lot of code is reusable!

PR #49, opened 2023-04-19
🐛 Handle Union for output type

prashantgupta24:
The introspection logic for inference now runs through the Union defined and picks the data model type from it.

fix https://github.com/caikit/caikit/issues/47


joerunde:
Yeah we really need to rename `serializers.py` and refactor that to be more meaningful :/
And of course add tests because right now those are only tested via the service factory which is awful to set up test contexts for.

Time to open a story...

caikit/runtime/service_generation/primitives.py
```diff
@@ -82,6 +82,36 @@ def to_primitive_signature(
     return primitives
 
 
+def to_output_dm_type(arg_type: type) -> str:
```
joerunde, line 85: Should this be `to_output_dm_type(arg_type: Type) -> Type:`?
Looks like it takes a type and returns a type

tests/fixtures/sample_lib/blocks/other_task/other_implementation.py
```diff
@@ -16,7 +19,7 @@ def __init__(self, batch_size=64, learning_rate=0.0015):
         self.batch_size = batch_size
         self.learning_rate = learning_rate
 
-    def run(self, sample_input: SampleInputType) -> OtherOutputType:
+    def run(self, sample_input: SampleInputType) -> Union[OtherOutputType, str]:
```
joerunde, line 22: super good enough for ow, definitely need some dang tests to follow on the `serializers` that are now no longer serializers

joerunde (approved):
Thanks @prashantgupta24 ! Just a couple little comments, this looks 🌶️🌶️🌶️

caikit/runtime/service_generation/primitives.py
joerunde: Could even throw one of those "complete but long and awful" names on here like `extract_data_model_type_from_union` 🤔

caikit/runtime/service_generation/primitives.py
```diff
@@ -82,6 +82,36 @@ def to_primitive_signature(
...
+        dm_types = [
+            arg
+            for arg in typing_args
+            if inspect.isclass(arg) and issubclass(arg, DataBase)
+        ]
+        if dm_types:
+            log.debug2(
+                "Found data model types in Union: [%s], taking first one", dm_types
+            )
+            return to_output_dm_type(dm_types[0])
```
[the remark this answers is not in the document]
prashantgupta24: Hmm. Yeah I thought about that, I left it in case we want to expand on this later on with more types of arguments. I can make it easy if that makes better sense

Issue #51, opened 2023-04-20
Union within a Train function argument throws an exception


prashantgupta24:
## Describe the bug
If we have a `train` function defined like so:
```
def train(
        cls,
        sample_input: Union[SampleInputType, str],
    )
```
I see an exception:
```
ValueError: Invalid input schema, cannot handle unions yet: typing.Union[sample_lib.data_model.sample.SampleInputType, str]
```

## Platform

Please provide details about the environment you are using, including the following:

- Interpreter version: 3.9
- Library version: 0.1.4

## Sample Code

Already provided

## Expected behavior

The introspection needs to pick the `primitive` or `primitive_data_model_types` from the Union.

## Observed behavior

What you see happening (error messages, stack traces, etc...)

PR #52, opened 2023-04-20
🐛 handle union within train func args

prashantgupta24:
The introspection now picks the `primitive` or `primitive_data_model_types` from the `Union` as an arg to `.train` functions.

fix https://github.com/caikit/caikit/issues/51

### DIFF
--- a/caikit/runtime/service_generation/primitives.py
+++ b/caikit/runtime/service_generation/primitives.py
@@ -45,9 +45,11 @@ def to_primitive_signature(
     primitives = {}
     log.debug("Building primitive signature for %s", signature)
     for arg, arg_type in signature.items():
-        primitives[arg] = extract_primitive_type_from_union(
+        primitive_arg_type = extract_primitive_type_from_union(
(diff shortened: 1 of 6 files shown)


caikit/runtime/service_generation/primitives.py
```diff
@@ -45,43 +45,47 @@ def to_primitive_signature(
...
-                        log.debug(
-                            "Just picking first primitive type %s in union",
-                            union_primitives[0],
-                        )
-                        primitives[arg] = union_primitives[0]
-
-            else:
-                primitives[arg] = arg_type
-        else:
-            log.debug("Skipping non-primitive argument [%s], type [%s]", arg, arg_type)
```
prashantgupta24, line 80: this refactor is changing the way we returned the value of the `primitives` dict though. Earlier, if there was no primitive, we returned an empty `{}.` But now it will have the name of the `arg: None`. Thinking...

joerunde (approved):
sweeeet!

PR #71, opened 2023-04-26
:arrow_up: unpin prometheus_client

joerunde:
See https://github.com/caikit/caikit/pull/63


prashantgupta24 (approved):

hickeyma:
@joerunde Is prometheus still used?

joerunde:
@hickeyma technically, yeah.

IIRC even if we move to opentelemetry, we would still need to import any packages that we use as exporters. So it's likely that the `prometheus_client` package will remain as a dependency, though `py-grpc-prometheus` could be nixxed

PR #72, opened 2023-04-26
test: Docstrings module tests

aluu317:
This PR is part of Issue #15

- refactors the 3 methods in `docstrings.py` that use 3rd party `docstring_parser.parse` [function](https://github.com/rr-/docstring_parser/blob/master/docstring_parser/parser.py#L23) to instead of checking for None, handle the `ParseError` exception. 
- adjusts the tests to mock the exception raised. I don't think we ever get this exception because we call `docstring_parser.parse` with Auto style, we always get a Docstring object back, albeit a Docstring object with empty params.
- add more tests to increase docstring test coverage:
Screen Shot 2023-04-26 at 3 18 01 PM

**Updated:**
- When handling `ParseError` exception, this PR adds a raise for `CaikitRuntimeException` and lets the callers of those 3 methods in `parsers.py` decide what to do with that exception. In this case, these callers log a warning that the docstring was not parseable and continue their type deduction task without docstring.
- Also figured I'd add tests to cover the rest of `module_signature.py` file to bring the coverage of `signature_parsing` module to 100%
Screen Shot 2023-04-27 at 4 19 53 PM

### DIFF
--- a/caikit/core/data_model/base.py
+++ b/caikit/core/data_model/base.py
@@ -256,7 +256,7 @@ class _DataBaseMetaClass(type):
     def _make_property_getter(mcs, field):
         """This helper creates an @property attribute getter for the given field
 
-        NOTE: This needs to live as a standalone funciton in order for the given
+        NOTE: This needs to live as a standalone function in order for the given
             field name to be properly bound to the closure for the attrs
         """
         private_name = f"_{field}"
--- a/caikit/runtime/service_generation/signature_parsing/docstrings.py
+++ b/caikit/runtime/service_generation/signature_parsing/docstrings.py
@@ -21,12 +21,14 @@ import sys
 
 # Third Party
 import docstring_parser
+import grpc
 
 # First Party
 import alog
 
 # Local
 from caikit.core.data_model.base import DataBase
+from caikit.runtime.types.caikit_runtime_exception import CaikitRuntimeException
 import caikit.core
 
 log = alog.use_channel("DOCSTRINGS")
@@ -44,9 +46,11 @@ def get_return_type(fn: Callable) -> Optional[Type]:
     """
     try:
         docstring = docstring_parser.parse(fn.__doc__)
-    except docstring_parser.ParseError:
-        log.warning("Failed to parse the docstring for %s", fn.__name__)
-        return None
+    except Exception as exc:
+        raise CaikitRuntimeException(
+            grpc.StatusCode.INVALID_ARGUMENT,
+            f"ParseError when parsing docstring for function: {fn.__name__}",
+        ) from exc
 
     type_names, desc_names = _get_candidate_type_names_from_docstring(docstring.returns)
 
@@ -71,9 +75,11 @@ def is_optional(fn: Callable, arg_name: str) -> bool:
     """
     try:
         docstring = docstring_parser.parse(fn.__doc__)
-    except docstring_parser.ParseError:
-        log.warning("Failed to parse the docstring for %s", fn.__name__)
-        return None
+    except Exception as exc:
+        raise CaikitRuntimeException(
+            grpc.StatusCode.INVALID_ARGUMENT,
+            f"ParseError when parsing docstring for function: {fn.__name__}",
+        ) from exc
 
     ds_param = [param for param in docstring.params if param.arg_name == arg_name]
     if ds_param:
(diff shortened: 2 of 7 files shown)


joerunde:
Thanks @aluu317!

I think I'm just a bit lost on what's going on here though:
> refactors the 3 methods in docstrings.py that use 3rd party docstring_parser.parse [function](https://github.com/rr-/docstring_parser/blob/master/docstring_parser/parser.py#L23) to instead of checking for None, handle the ParseError exception.

Is there a case where we don't get a `ParseError` exception back but we do get a `None` back? I know the test coverage was pretty bad before, but I would want to make sure that we still get visible logging and correct handling when that case happens.

> adjusts the tests to mock the exception raised. I don't think we ever get this exception because we call docstring_parser.parse with Auto style, we always get a Docstring object back, albeit a Docstring object with empty params.

I think this may answer my question from above, does this mean that the `Docstring object with empty params` is the case that was previously checked for by `if not docstring` instead of catching the parser error?

caikit/runtime/service_generation/signature_parsing/docstrings.py
```diff
@@ -68,10 +69,11 @@ def is_optional(fn: Callable, arg_name: str) -> bool:
...
             e.g. "raw_document"
     """
-    docstring = docstring_parser.parse(fn.__doc__)
-    if not docstring:
+    try:
+        docstring = docstring_parser.parse(fn.__doc__)
+    except docstring_parser.ParseError:
         log.warning("Failed to parse the docstring for %s", fn.__name__)
-        return False
+        return None
```
joerunde, line 76: nit: this function returns a `bool`, should this still be `False`? Or is there a semantic difference between `False` and `None` that we're checking for later?

tests/runtime/service_generation/signature_parsing/test_docstrings.py
```diff
@@ -45,3 +85,67 @@ def test_extract_nested_type():
...
+        Returns:
+            None
+        """
+        pass
+
+    assert is_optional(_fn, "some_input") == True
+
+    # test if docstring_parser.parse throws an exception, we return None
+    with patch("docstring_parser.parse", side_effect=ParseError("mocked error")):
+        assert is_optional(_fn, "some_input") == None
```
joerunde, line 112: Nice, love to see it!

Do we also have test cases for places where the docstring parser fails but not with an exception? Like if there's no docstring defined at all or if it's an empty string?

caikit/core/data_model/base.py
```diff
@@ -256,7 +256,7 @@ def __new__(mcs, name, bases, attrs):
     def _make_property_getter(mcs, field):
         """This helper creates an @property attribute getter for the given field
 
-        NOTE: This needs to live as a standalone funciton in order for the given
+        NOTE: This needs to live as a standalone function in order for the given
```
joerunde, line 259: I always read these as "Funkytown!" and get a little chuckle

tests/runtime/service_generation/signature_parsing/test_docstrings.py
```diff
@@ -45,3 +85,67 @@ def test_extract_nested_type():
...
+    with patch("docstring_parser.parse", side_effect=ParseError("mocked error")):
+        assert get_arg_type(_fn, "some_input") == None
+
+
+def test_get_return_type_corner_case_with_exception():
+    def _fn():
+        pass
+
+    # test get_return_type works on empty docstring
+    assert get_return_type(_fn) == None
```
joerunde, line 147: ah, I see this test for no docstring. This should cover the `get_arg_types` case as well, right?

joerunde (approved):
Looks 🌶️! 

It sounds like we have some future design discussing to do about enforcing or otherwise handling styled docstrings vs. unstyled ones?

Could be cool to detect that a docstring is, in fact, `Google` or `numpy` styled and skip all the extra "yadda yadda yadda" processing 🤔

aluu317:
To clarify my Description above (I did a bad job explaining):
3rd party docstring_parser.parse [function](https://github.com/rr-/docstring_parser/blob/master/docstring_parser/parser.py#L23) does not return None. This line [here](https://github.com/rr-/docstring_parser/blob/master/docstring_parser/parser.py#L43) tells me that if there's no Docstring parsed, a ParseError had occurred and the exception was raised. So in our `docstrings.py` code, instead of checking for None, I chose to handle the ParseError instead.

caikit/runtime/service_generation/signature_parsing/docstrings.py
aluu317: I considered throwing an exception here and let the caller of this function handle that... I could do that

Issue #30, opened 2023-04-13
Remove project structure requirements/assumptions


gabe-l-hart:
## Is your feature request related to a problem? Please describe.

Currently, there are several places in the code that make assumptions about the directory/module structure for the derived library. We should identify these areas and make them more generic to allow arbitrary project structures.

## Describe the solution you'd like

- [x] Identify all places where module structure assumptions are made
- [x] Fix them!

joerunde:
@gabe-l-hart @tharapalanivel @prashantgupta24 So I think everything left for this issue is covered by #90 _except_ for the training RPC names. Right now those smash the whole directory structure for a module together to create a unique rpc name.

We may actually want an option to keep that around for backwards compatibility reasons, but we should probably also think of a better way to have unique training RPC names that aren't as verbose and don't break if you move a module to a different file.

Thoughts on repurposing this story for that specific problem?

joerunde:
@evaline-ju is auditing right now, but we should be extremely close, if not done with this

joerunde:
Closing as "very likely completed"

PR #73, opened 2023-04-26
✨ Task decorator

prashantgupta24:
For #76 and #30

This PR provides `@task` and `@taskgroup` decorators, which authors can use to explicitly specify the task that a model belongs to.

Very open to feedback, we followed the `@module` pattern of throwing a decorator on a class which a user could theoretically tweak the behavior of. The `Task` classes are responsible for providing validators on inference input / output types.

This will unblock:
- Massive simplification in the runtime service generation, since we won't have to guess/infer the task types of modules
- Module verification at module-decoration-time rather than runtime-boot-time, which is a pain for our users
- And will remove (most) structural requirements to users' code bases, since we won't have to use the python module paths when reasoning about models

This is currently ~a **breaking**~ not a breaking change, ~since it requires a `task` parameter~ it will require a task parameter in the future for all modules that don't either:
- have `base_module` set or
- have a parent class already with a `task`


markstur (commented):
If I understand correctly, as a caikit consumer this does nothing for me right now. Task-level APIs sounds like a good thing(?).  If this is released before task APIs release with it, probably should add some support for undefined tasks so that this can be non-breaking. Otherwise I could see this getting in the way of any important non-breaking fixes going out.

joerunde:
@markstur to clarify, the runtime already generates an inference service with task level apis, but the tasks are just inferred from the python module structure and require some interesting `runtime.service_generation.primitive_data_model_types` configuration in order to work at all. The user needs to put every data model type in that list that they want exposed via the task apis, and any model with a data model parameter not in that list is not included, which is a little 🤮 

You are also correct though that this doesn't really do anything until it's hooked up to either the:
- service generation module, so that we can generate task-level APIs without requiring specific python module structures nor the `runtime.service_generation.primitive_data_model_types` config
- module decorator's input validation, to allow users to get feedback when a module's `.run` function doesn't match the task

I could be down for making the `task` bit optional for this change, and requiring it once we come back around and hook up the api generation with it

markstur (commented):
❤️

caikit/core/task.py
```diff
@@ -0,0 +1,180 @@
...
+    @classmethod
+    def validate_task_inputs(cls) -> bool:
+        pass
+
+    @classmethod
+    def get_input_type_set(cls) -> Set[ProtoableInputTypes]:
+        raise NotImplementedError("This is implemented by the @taskgroup decorator!")
+
+
+class TaskBase:
```
gabe-l-hart, line 45: NIT: Needs a docstring

caikit/core/task.py
```diff
@@ -0,0 +1,180 @@
...
+    @classmethod
+    def validate_run_signature(cls) -> bool:
+        # TODO: implement
```
gabe-l-hart, line 48: Worth doing now or defer to another PR?

caikit/core/module.py
```diff
@@ -1248,6 +1260,30 @@ def decorator(cls_):
...
+                    )
+                if tasks:
+                    parent_task = tasks.pop()
+                    if task and task != parent_task:
+                        raise TypeError(
+                            f"Class {cls_} has task {task} but superclass has task {parent_task}"
+                        )
+                    cls_.TASK_CLASS = parent_task
+                else:
+                    if not task:
```
gabe-l-hart, line 1281: Maybe stick a `warning` (or at least a `debug`) log statement here so that authors whose blocks don't have a `task` can easily identify them

caikit/core/task.py
```diff
@@ -0,0 +1,180 @@
...
+        pass
+
+    @classmethod
+    def get_required_inputs(cls) -> Dict[str, ValidInputTypes]:
+        raise NotImplementedError("This is implemented by the @task decorator!")
```
gabe-l-hart, line 53: I'm assuming the reason this isn't using `abc` is that the derived tasks are dynamically created with ad-hoc `setattr` calls below? We could pobably get around that, though I'm also not clear that `@abc.abstractmethod` and `@classmethod` play nicely together despite what the documentation says.

caikit/core/task.py
```diff
@@ -0,0 +1,180 @@
...
+        raise TypeError("task_group must be a TaskGroup class")
+
+    for parameter_name, input_type in required_inputs.items():
+        if input_type not in task_group.get_input_type_set():
+            raise TypeError(
+                f"Task parameter {parameter_name} has type {input_type} not in task_group: "
+                f"{task_group.__name__}. Valid types are: {task_group.get_input_type_set()}"
+            )
+
+    def get_required_inputs(_):
```
gabe-l-hart, line 125: These inline-declared functions should have docstrings so that when a user does `help` on a class wrapped with `@task`, the docs show up nicely

caikit/core/task.py
```diff
@@ -0,0 +1,180 @@
...
+        return required_inputs
+
+    def get_output_type(_):
+        return output_type
+
+    def get_task_group(_):
+        return task_group
+
+    def decorator(cls: Type[TaskBase]) -> Type[TaskBase]:
+        if not isinstance(cls, type) or not issubclass(cls, TaskBase):
```
gabe-l-hart, line 135: NIT: Use `error`

caikit/core/task.py
```diff
@@ -0,0 +1,180 @@
...
+
+
+class TaskBase:
+    @classmethod
+    def validate_run_signature(cls) -> bool:
+        # TODO: implement
+        pass
+
+    @classmethod
+    def get_required_inputs(cls) -> Dict[str, ValidInputTypes]:
```
gabe-l-hart, line 52: Even though these methods will always be overwritten, they should have docstrings so that `help(TaskBase.get_required_inputs)` works and gives useful info about what the method does in the abstract.

gabe-l-hart (changes requested):
This is a _really_ nice solution to the task inference issue. It also solves the need for "domain primitives" quite nicely.

I've got a bunch of small coding comments/suggestions. Since this is going to be a primary abstraction in the public API, I think making sure the `TaskBase` and `TaskGroupBase` classes are well documented with docstrings is really important!

caikit/core/module.py
```diff
@@ -1211,9 +1216,16 @@ class as a valid {module_type} with caikit.core
...
                 name = base_module_class.MODULE_NAME
+                task = base_module_class.TASK_CLASS
                 backend_module_impl = True
 
             error.type_check("<COR54118928E>", str, id=id, name=name, version=version)
+            # Technically `task` can still be None here:
+            # We'll check when the decorator runs to see if the class extends one that
+            # already has a task
+            if task is not None:
+                if not isinstance(task, type) or not issubclass(task, TaskBase):
```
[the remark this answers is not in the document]
joerunde: Yeah, I'm pretty sure that `error.type_check` needs to be modified a bit to support this case where the input is itself a type. I left that as an afterthought but if we're good with the overall approach here we can slam it in

caikit/core/task.py
joerunde: Deferring I think- that'd be part of hooking this up to better service generation and doesn't need to block getting the apis out

caikit/core/module.py
joerunde: 🌶️

caikit/core/task.py
joerunde: Yeah, I slapped abc on at first but then you get "unimplemented abstract method" linter errors on classes that users decorate. There might be a way around that but I didn't take the time to look closer

caikit/core/task.py
```diff
@@ -0,0 +1,180 @@
...
+
+
+def taskgroup(
+    input_types: Set[ProtoableInputTypes],
+) -> Callable[[Type[TaskGroupBase]], Type[TaskGroupBase]]:
+    """The decorator for AI Task Groups"""
+
+    def type_check(x: type) -> bool:
+        return (
+            x == int
```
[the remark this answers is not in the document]
joerunde: yup, definitely part of extending the error handler to do proper type checks on type input

joerunde (approved):

PR #74, opened 2023-04-26
:pushpin::heavy_minus_sign: Add dep upper bounds and remove wheel

evaline-ju:
Pin dependency upper bounds to prevent potential breaking changes, `wheel` is not needed beyond test time

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,10 +14,10 @@ classifiers=[
     "License :: OSI Approved :: Apache Software License"
 ]
 dependencies = [
-    "alchemy-config>=1.1.1",
-    "alchemy-logging>=1.0.4",
+    "alchemy-config>=1.1.1,<1.2.0",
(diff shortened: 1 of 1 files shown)


hickeyma (commented):
@evaline-ju Will setting the upper bound make it susceptible to missing a vulnerability update?

evaline-ju:
> @evaline-ju Will setting the upper bound make it susceptible to missing a vulnerability update?

@hickeyma yes potentially - I was initially doing a pretty conservative minor upper bound for alog and aconfig since caikit was broken temporarily with an untested edge case with a minor bump in `aconfig`, but have now bumped to a `<` major version to be more flexible. I have kept the narrower band in `docstring-parser` since it is pre `1.0`. I am also hoping tools like dependabot will help notify of updates.

joerunde (approved):

PR #77, opened 2023-04-27
:sparkles: Add train process exit code handling

gkumbhat:
### Changes
- Add subprocess exit code handling for training to allow capturing OOM type failures that kills the subprocess

### DIFF
--- a/caikit/runtime/servicers/global_train_servicer.py
+++ b/caikit/runtime/servicers/global_train_servicer.py
@@ -360,6 +360,19 @@ class GlobalTrainServicer:
         proc.start()
         proc.join()
 
+        if proc.exitcode is not None:
+            # If process exited with a non-zero exit code
+            if proc.exitcode == 137:
+                raise CaikitRuntimeException(
+                    grpc.StatusCode.RESOURCE_EXHAUSTED,
+                    "Training process died with OOM error!"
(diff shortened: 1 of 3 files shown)


caikit/runtime/servicers/global_train_servicer.py
```diff
@@ -360,6 +360,19 @@ def _train_and_save_model_subproc(cls, *args, **kwargs):
         proc.start()
         proc.join()
 
+        if proc.exitcode is not None:
+            # If process exited with a non-zero exit code
+            if proc.exitcode == 137:
```
gabe-l-hart, line 365: This feels pretty platform specific. Is there any platform-agnostic way to detect different exit codes? Since the process is `python` itself, I would think maybe `sys` would be able to do something like this? Just a guess

gkumbhat: There doesn't seem to be a straight forward way to get platform independent OOM exit code.  `sys` doesn't seem to have declaration of exit codes, `os` have some, but not for OOM case and neither in `signal` :/

gabe-l-hart (approved):
Nice. In the absence of a platform-independent representation of the OSS exit code, I think this is a good way to handle it with a constant. Thanks!

PR #78, opened 2023-04-27
:truck: Rename grpc server params

evaline-ju:
This is a public API and the shortened names are more difficult to understand


joerunde (approved):
#agree

alanbraz:
much better making it more readable!

PR #81, opened 2023-04-28
:memo: Write the first cakepop

joerunde:
This is a rough stab at a `PEP` but for `caikit`: a `Cake POP`!

After some long discussions with @prashantgupta24 about the deficiencies of both @blocks and implicit task definitions, we have a design to share that we think would make this all a whole lot more usable


docs/POPs/001-domains-tasks-and-couplers.md
```diff
@@ -0,0 +1,174 @@
...
+```python
+@caikit.task(
+    domain=caikit.interfaces.nlp.NLP,
+    output_type=caikit.interfaces.nlp.SentimentPrediction
+)
+class Sentiment(caikit.TaskBase):
+    pass
+```
+
+A new class TaskBase will provide a `validate_output` method which will raise if the `CaikitSignature` does return the correct type for the task.
```
prashantgupta24, line 154:
```suggestion
A new class TaskBase will provide a `validate_output` method which will raise if the `CaikitSignature` does not return the correct type for the task.
```

hickeyma (changes requested):
I'm not sure of this name for the process. Maybe something like Caikit Improvement Process (CIP) might read better and be more informative?
The first doc (001) should be explaining the process and not the first improvement/enhancement.

PR #83, opened 2023-05-01
:bug: add second argument to interrupt()

tjohnson31415:
The `signal.signal` interface has a second argument:

> The handler is called with two arguments: the signal number and the current stack frame

REF: https://docs.python.org/3/library/signal.html#signal.signal

Without this change, sending a signal to the server process results in a stack trace with:
```
TypeError: interrupt() takes 2 positional arguments but 3 were given
```


evaline-ju (approved):

gabe-l-hart (approved):
Looks good! Thanks Travis

PR #85, opened 2023-05-02
♻️ warning message for missing train service

prashantgupta24:
2 changes:
1. `grpc_server.py`:
    Doesn't make sense to have `log.error` for this statement since we still go forward with the initialization of the grpc server even if train service doesn't exist?
2. `import_util.py`:
    Doesn't make sense to `log.error` _and_ throw an exception?

### DIFF
--- a/caikit/runtime/utils/import_util.py
+++ b/caikit/runtime/utils/import_util.py
@@ -164,7 +164,7 @@ def get_dynamic_module(module_name: str, module_dir: str = None) -> ModuleType:
     spec = importlib.util.find_spec(module_path)
     if not spec:
         message = "Unable to find spec for module: %s" % (module_path)
-        log.error("<RUN11991313E>", message)
+        log.warning("<RUN11991313E>", message)
         raise CaikitRuntimeException(StatusCode.NOT_FOUND, message)
     # Found spec - import the library
     if module_dir:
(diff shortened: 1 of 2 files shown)


caikit/runtime/utils/import_util.py
```diff
@@ -164,7 +164,7 @@ def get_dynamic_module(module_name: str, module_dir: str = None) -> ModuleType:
     spec = importlib.util.find_spec(module_path)
     if not spec:
         message = "Unable to find spec for module: %s" % (module_path)
-        log.error("<RUN11991313E>", message)
+        log.warning("<RUN11991313E>", message)
```
joerunde, line 167: 🤔, okay here's some food for thought: 

I know we're not super consistent but we should try to follow:
- Warning when something goes wrong but we can recover or fall back to something else to complete the operation
- Error when something goes wrong that we can't recover from and the operation cannot be completed
- Fatal when something goes wrong that the application can't recover from, and crashes

(Also a little nit that the last letter of the log code should match the log level)

This particular piece of code is kinda buried deep in a few different operations so it's hard to tell what's going on. If this fails when starting a servicer and the app crashes, then it's a fatal error. But if we attempt to import something and move on if it's not there, it'd just be a warning.

I could see just removing this log altogether and letting the caller decide how bad this error is when the exception is caught?

But also- I bet our exception handling hygiene is not the best everywhere and I would not bet money that this is even caught in all code paths. I'd nit on the exception on the next line that `CaikitRuntimeException`s were meant for raising errors during the processing of a runtime request, and I'm not sure this particular piece of code is ever invoked during a request.

/shrug

joerunde (approved):
Thanks for improving opsviz @prashantgupta24!

Left some thoughts, I like all the improvements we can get

caikit/runtime/utils/import_util.py
gabe-l-hart: Totally agree with your assessment @joerunde

aluu317 (approved):
LGTM!

Issue #82, opened 2023-05-01
Support dataclasses as schema for data model


evaline-ju:
## Description

As a Caikit model author, I want to be able to define my data model classes with dataclasses, so that I can have a Python-friendly schema to define my data models.

## Discussion

We want to use the new features in [py-to-proto](https://github.com/IBM/py-to-proto), formerly jtd-to-proto.

## Acceptance Criteria

- [ ] Users are able to define data models with dataclasses
- [ ] Service generation uses dataclasses
- [ ] Unit tests cover new/changed code
- [ ] Examples build against new/changed code

PR #86, opened 2023-05-02
Dataclass to proto

gabe-l-hart:
## Description

**This PR is a major API breaking change!!**

This PR overhauls the implementation of `@dataobject` to primarily use [the native `@dataclass`](https://docs.python.org/3/library/dataclasses.html) representation of "plain old data" and [the native `Enum`](https://docs.python.org/3/library/enum.html) to represent enumerations.

In the process of this overhaul, pure JTD can still be used with the explicit `schema=` kwarg, but the custom shorthand has been removed.

## Discussion

Through a lot of discussion, the team has decided that the [JTD-based implementation of `@dataobject` is not ideal](https://github.com/caikit/caikit/issues/82). The main problems with this are:

1. There are corner-cases of `proto` that cannot be represented by JTD (non-sequential enums, enum aliases, non-sequential field numbers, oneofs with primitive members).
2. JTD is less well known than `proto` or `@dataclass` for python developers, so while the learning curve is not steep, it's less straightforward to find supporting materials
3. The custom "shorthand" syntax on top of JTD, while cleaner, is fully custom and not something that is supported with public documentation beyond `caikit` itself.

This PR is built on the changes from `jtd-to-proto` migrating to `py-to-proto` ([PR to add `dataclass_to_proto`](https://github.com/IBM/py-to-proto/pull/46)) and updates the implementation of `@dataobject` accordingly.

## API Breaking Changes

* The "shorthand" version of JTD schemas that could be used with `@dataobject` has been fully removed (e.g. `{"foo": int}` versus `{"properties": {"foo": {"type": "int64"}}}`
* In order to use the long-form JTD schema version of `@dataobject`, you must explicitly use the `schema=` keyword arg
* The system for enums is completely overhauled:
    * Declarations of `@dataobject` enums are indicated by deriving from `enum.Enum`
    * Enums imported from compiled `proto`s are converted into `enum.Enum` classes instead of `EnumBase`
    * `caikit.core.data_model.enums.EnumBase` has been completely removed
    * Since enums are no longer `munch.Munch` objects, they no longer directly expose `toYAML` and `toJSON` functions. Instead, each has a `to_munch` function that will lazily perform a one-time munch conversion which can then be used to call `toYAML`/`toJSON`
        * The `toDict` function has also been renamed `to_dict` to be more pythonic
    * Since all enums are now standard python `enum.Enum`, all enum attributes must call `.values` to dereference to the integer value

## Outstanding TODOs

* Add an ADR or all of this
* Improve `oneof` support so that [this test works](https://github.com/gabe-l-hart/caikit/blob/DataclassToProto/tests/core/data_model/test_dataobject.py#L424)


caikit/core/data_model/dataobject.py
```diff
@@ -18,99 +18,64 @@
...
-        schema={"foo": str, "bar": int},
-    )
+    @dataobject("foo.bar")
+    @dataclass
     class MyDataObject:
         '''My Custom Data Object'''
+        foo: str
+        bar: int
 
     NOTE: The wrapped class must NOT inherit directly from DataBase. That
```
joerunde, line 73: Just a question: Is there any way to get that to work?

This snippet makes a static type checker barf but runtime type checking passes:
```python
@dataobject
@dataclass
class MyDataObject:
    foo: int

def my_func(data_model: DataBase):
    assert isinstance(data_model, DataBase)

my_func(MyDataObject())
```

joerunde (approved):
lpgdgtm

too much diff to step closely through but this looks super hotttt

gabe-l-hart:
There will be a few follow-on PRs to this one to improve usability and eventually strip out the `"schema"` flavor of `@dataobject`, but given that this is in a stable place, let's merge as is and iterate.

PR #89, opened 2023-05-03
:fire: delete leftover xxPrediction logic

### LINKED ISSUE #31: Allow multiple output types (names) in caikit runtime

Currently `caikit.runtime`  deduces return type based on request type name `<task-type>Request` -> `<task-type>Prediction`, which is implemented [here](https://github.com/caikit/caikit/blob/main/caikit/runtime/service_generation/signature_parsing/parsers.py#L86). However, a user may need to have different output type names other than `<task-type>/Prediction`. 

One possible solution might be for output pattern types to be made configurable via `caikit.runtime.config` and allowing list of output "suffixes".

Renaming all my output types for `.run` function with `Prediction` suffixes, however the term "prediction" doesn't represent all the domains of AI problems.

## Additional context

joerunde:
Closes #31 

As a fun sidequest, the recent bump to prometheus_client 0.16 bricked our tests for our custom throughput metric.

This removes the throughput metric too and drops a little comment in the code where it was pulled out. We probably didn't want to be in the business of writing server-side throughput metrics anyway, since those would be impossible to correctly aggregate across pods


evaline-ju (approved):
LGTM - thanks for the cleanup!

evaline-ju (approved):
LGTM!

PR #91, opened 2023-05-04
Data object base

gabe-l-hart:
## Description

The [discussion on the `dataclass_to_proto` PR](https://github.com/caikit/caikit/pull/86#discussion_r1183077788) brought up a really interesting point that injecting base classes with a decorator is confusing at best and an anti-pattern at worst.

This PR attempts to fix that by having a new base class for all data model classes that use `@dataobject`: `DataObjectBase`.

## Details

The crux of this PR is lifting the logic for parsing the protobuf descriptor out of the `__new__` function on `_DataBaseMetaClass` and allowing it to be called from within `@dataobject` once the proto class has been generated. This required some fancy footwork with a new `metaclass`/base class pair: `_DataObjectBaseMetaClass`/`DataObjectBase`. The only logic that couldn't be hoisted out of `_DataBaseMetaClass.__new__` was the initialization of the class attributes, and in particular `__slots__` which needs to know the names of the fields that will eventually be valid on an instance of the class. To get around this, `_DataObjectBaseMetaClass` uses a special "forward declaration" attribute that is popped out of the class attrs by `_DataBaseMetaClass` which contains the names of the fields discovered from the `__annotations__` that will be used as input to the `@dataclass` call.

## Fallout

* With this change, all uses of `@dataobject` must inherit from `DataObjectBase` OR follow the special hackery to use `schema`
* If using a JTD `schema`, the class must also use annotations to declare the fields so that the forward-declaration stuff doesn't explode.
    * This is _only_ done to avoid having this PR encapsulate the changes needed to support `oneofs` robustly. These semantics are needed in `DataStreamSource`, so once `oneof` is properly supported, we can move that implementation off of `schema` and fully remove the JTD style of `@dataobject`
* The clause in `dataobject` that auto-created a special `"protobufs"` module was removed as unnecessary and unused (or so I thought)
* The removal of the special `"protobufs"` auto-gen module triggered issues in the special logic to find a precompiled protobuf in `_DataBaseMetaClass.__new__`. Rather than further patch around this, I (with input from @joerunde) decided to fully remove this "automagic" logic and instead require that anyone using precompiled protos explicitly declare the correlation between the data model class and the proto class by setting `_proto_class` as a class attribute on the data model class pointing to the proto class.

### DIFF
--- a/caikit/core/data_model/base.py
+++ b/caikit/core/data_model/base.py
@@ -155,15 +155,6 @@ class _DataBaseMetaClass(type):
             [f"_{field}" for field in fields] + list(private_slots) + ["_backend"]
         )
 
-        # add properties that use the underlying backend
-        for field in fields:
-            attrs[field] = mcs._make_property_getter(field)
-
-        # If there is not already an __init__ function defined, make one
-        current_init = attrs.get("__init__")
-        if current_init is None or current_init is DataBase.__init__:
-            attrs["__init__"] = mcs._make_init(fields)
-
         # Set fields class variable for reference
         # these are valuable for validating attributes and
         # also for recursively converting to and from protobufs
@@ -179,21 +170,18 @@ class _DataBaseMetaClass(type):
         attrs["_fields_primitive"] = _fields_primitive
         attrs["_fields_primitive_repeated"] = _fields_primitive_repeated
         attrs["_proto_class"] = proto_class
+
         instance = super().__new__(mcs, name, bases, attrs)
 
         # If there's a valid proto class, perform proto descriptor parsing
         if proto_class is not None:
             mcs.parse_proto_descriptor(instance)
 
-        # Update the global class and proto registries
-        if name not in ["DataBase", "DataObjectBase"]:
-            mcs.class_registry[instance.full_name] = instance
-
         # Return the constructed class instance
         return instance
 
-    @staticmethod
-    def parse_proto_descriptor(cls):
+    @classmethod
+    def parse_proto_descriptor(mcs, cls):
         """Encapsulate the logic for parsing the protobuf descriptor here. This
         allows the parsing to be done as a post-process after metaclass
         initialization
@@ -279,6 +267,20 @@ class _DataBaseMetaClass(type):
         cls._fields_primitive_repeated = frozenset(fields_repeated).intersection(
             _fields_primitive_all
         )
+
+        # Update the global class and proto registries
+        # NOTE: Explicitly not respecting metaclass inheritance so single
+        #   registry shared for all
+        _DataBaseMetaClass.class_registry[cls.full_name] = cls
+
+        # Add properties that use the underlying backend
+        for field in cls.fields:
+            setattr(cls, field, mcs._make_property_getter(field))
+
+        # If there is not already an __init__ function defined, make one
+        current_init = cls.__init__
+        if current_init is None or current_init is DataBase.__init__:
+            setattr(cls, "__init__", mcs._make_init(cls.fields))
 
     @classmethod
     def _make_property_getter(mcs, field):
--- a/caikit/core/data_model/base.py
+++ b/caikit/core/data_model/base.py
@@ -54,6 +54,11 @@ class _DataBaseMetaClass(type):
     # attribute does exist.
     _MISSING_ATTRIBUTE = "missing attribute"
 
+    # Special attribute used to communicate that the proto fields are forward
+    # declared and will be populated after the metaclass has completed
+    # construction.
+    _FWD_DECL_FIELDS = "__fwd_decl_fields__"
+
     def __new__(mcs, name, bases, attrs):
         """When constructing a new data model class, we set the 'fields' class variable from the
         protobufs descriptor and then set the '__slots__' magic class attribute to fields.  This
@@ -82,7 +87,7 @@ class _DataBaseMetaClass(type):
         _fields_map = ()
         proto_class = None
         full_name = name
-        if name != "DataBase":
+        if name not in ["DataBase", "DataObjectBase"]:
             # Look for the proto class. There are two places it could be:
             #
             # 1. The "protobufs" module in caikit.core (injected via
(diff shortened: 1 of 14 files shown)


caikit/core/data_model/base.py
```diff
@@ -305,7 +295,7 @@ def __init__(self, *args, **kwargs):
             if num_args + num_kwargs > num_fields:
                 error(
                     "<COR71444420E>",
-                    ValueError(f"Too many arguments given. Args are: {fields}"),
+                    TypeError(f"Too many arguments given. Args are: {fields}"),
```
gabe-l-hart, line 298: I changed these to match standard python error semantics

caikit/core/data_model/base.py
```diff
@@ -82,176 +86,162 @@ def __new__(mcs, name, bases, attrs):
         _fields_map = ()
         proto_class = None
         full_name = name
-        if name != "DataBase":
-            # Look for the proto class. There are two places it could be:
```
gabe-l-hart, line 86: Getting rid of this logic is a big step towards removing reliance on project structure!

caikit/core/data_model/base.py
```diff
@@ -82,176 +86,162 @@ def __new__(mcs, name, bases, attrs):
...
-                    f"library protobufs"
+                error.value_check(
+                    "<COR49310991E>",
+                    fields is not None,
+                    name,
+                    msg="No proto class found for {}",
                 )
 
-            # all fields
-            fields += tuple(proto_class.DESCRIPTOR.fields_by_name)
```
gabe-l-hart, line 133: All of this logic moved to `parse_proto_descriptor` below so that it could be lazily invoked in `@dataobject`

caikit/core/data_model/dataobject.py
```diff
@@ -54,8 +52,43 @@
...
+
+        # Get the annotations that will go into the dataclass
+        if name != "DataObjectBase":
+            raw_field_names = attrs.get("__annotations__")
+            if raw_field_names is None:
+                raise TypeError(
+                    "All DataObjectBase classes must follow dataclass syntax"
+                )
+
+            # TODO: Sort out oneof field names
```
gabe-l-hart, line 76: This will be part of the next PR to properly handle `oneof` everywhere

caikit/core/data_model/dataobject.py
```diff
@@ -54,8 +52,43 @@
...
+            field_names = list(raw_field_names.keys())
+
+            # Add the forward declaration
+            attrs[_DataBaseMetaClass._FWD_DECL_FIELDS] = field_names
+
+        # Delegate to the base metaclass
+        return super().__new__(mcs, name, bases, attrs)
+
+
+class DataObjectBase(DataBase, metaclass=_DataObjectBaseMetaClass):
```
gabe-l-hart, line 86: This base class serves no purpose other than to bind `DataBase` and `_DataObjectBaseMetaClass` together. In the future, we may need real logic here, but we should try hard to avoid it since we want all classes declared using `@dataobject` to cleanly line up with those that use `DataBase` directly

caikit/core/data_model/dataobject.py
```diff
@@ -88,9 +121,9 @@ def decorator(cls: Type) -> Type[DataBase]:
         # Make sure that the wrapped class does NOT inherit from DataBase
         error.value_check(
             "<COR95184230E>",
-            not issubclass(cls, DataBase),
-            "{} should not directly inherit from DataBase when using @schema",
```
gabe-l-hart, line 92: This was a fun learning experience for me. The `value_check` signature is strange and requires that `msg` be given as a kwarg while the `format` args come before the message. 🤷

caikit/core/data_model/dataobject.py
```diff
@@ -150,27 +183,14 @@ def decorator(cls: Type) -> Type[DataBase]:
...
-            wrapper_class = _make_data_model_class(proto_class, cls)
+            setattr(cls, "_proto_class", proto_class)
+            cls = _make_data_model_class(proto_class, cls)
         else:
             enums.import_enum(proto_class, cls)
             setattr(cls, "_proto_enum", proto_class)
-            wrapper_class = cls
-
-        # Attach the proto class to the protobufs module
-        parent_mod_name = getattr(cls, "__module__", "").rpartition(".")[0]
```
gabe-l-hart, line 160: This was all in service of keeping the expected module structure for libraries with a "protobufs" module full of precompiled protos. Since we're moving away from that, this needed to go!

caikit/core/data_model/dataobject.py
```diff
@@ -215,9 +235,11 @@ class _DataobjectConverter(DataclassConverter):
     def get_concrete_type(self, entry: Any) -> Any:
         """Also include data model classes and enums as concrete types"""
         unwrapped = self._resolve_wrapped_type(entry)
-        if (isinstance(unwrapped, type) and issubclass(unwrapped, DataBase)) or hasattr(
-            unwrapped, "_proto_enum"
-        ):
+        if (
```
gabe-l-hart, line 238: This was interesting. Now that the class being decorated is _already_ a derived `DataBase` class, this was triggering to detect it as an already-converted "concrete type reference." The check on `unwrapped._proto_class is not None` eliminates this case and indicates that this class has yet to be converted.

caikit/core/data_model/dataobject.py
```diff
@@ -270,31 +292,32 @@ def _get_all_enums(
     return all_enums
 
 
-def _make_data_model_class(proto_class, wrapped_cls):
-    wrapper_cls = _DataBaseMetaClass(
```
gabe-l-hart, line 274: Very nice to be able to get rid of this dynamically declared shim class!

caikit/core/data_model/dataobject.py
```diff
@@ -270,31 +292,32 @@ def _get_all_enums(
...
-        wrapped_cls.__name__,
-        tuple([DataBase, wrapped_cls]),
-        {"_proto_class": proto_class, **wrapped_cls.__dict__},
-    )
-    update_wrapper(wrapper_cls, wrapped_cls, updated=())
+def _make_data_model_class(proto_class, cls):
+    if issubclass(cls, DataObjectBase):
+        _DataBaseMetaClass.parse_proto_descriptor(cls)
```
gabe-l-hart, line 297: This is where the lazy proto parsing happens now that the protobuf class has been created with `dataclass_to_proto`

caikit/core/data_model/dataobject.py
```diff
@@ -270,31 +292,32 @@ def _get_all_enums(
...
         nested_message_name = nested_message_descriptor.name
         nested_proto_class = getattr(proto_class, nested_message_name)
         setattr(
-            wrapper_cls,
+            cls,
             nested_message_name,
             _make_data_model_class(
                 nested_proto_class,
-                types.new_class(nested_message_name),
+                _DataBaseMetaClass.__new__(
```
gabe-l-hart, line 308: There might be a cleaner way to do this. The doc on `types.new_class` says that its whole reason to exist is to create a new type using the right `metaclass`, but I couldn't easily find a way to give it `attrs`, so I wend the manual route.

caikit/runtime/service_factory.py
```diff
@@ -350,7 +354,12 @@ def _create_request_message_types(
                 continue
 
             decorator = dataobject(package=package_name)
-            cls_ = type(task.request.name, (object,), attrs)
+            cls_ = _DataObjectBaseMetaClass.__new__(
```
gabe-l-hart, line 357: Same point here about maybe wanting a cleaner way to dynamically create a new `DataObjectBase` derived class on the fly

caikit/runtime/service_generation/data_stream_source.py
```diff
@@ -218,43 +222,57 @@ def make_data_stream_source(data_element_type: Type) -> Type[DataBase]:
...
-            },
-            package="caikit_data_model.runtime",
-        )(
-            type(
-                cls_name,
-                (DataStreamSourceBase,),
-                {"ELEMENT_TYPE": data_element_type},
+            }
+        }
+        # TODO: Once full oneof support is implemented, this can move to using
```
gabe-l-hart, line 255: Gross! Will get this done soon...

caikit/runtime/service_generation/data_stream_source.py
```diff
@@ -263,14 +281,14 @@ def make_data_stream_source(data_element_type: Type) -> Type[DataBase]:
...
-            sys.modules[__name__],
+            sys.modules[data_object.__module__],
             cls_name,
             data_object,
         )
 
         # Add an init that sequences the initialization so that
         # DataStreamSourceBase is initialized after DataBase
-        orig_init = data_object.__init__
+        orig_init = _DataObjectBaseMetaClass._make_init(data_object.fields)
```
gabe-l-hart, line 291: This is needed because the logic to detect whether or not there's a "custom __init__" fails since the class inherits from `DataStreamSourceBase`. Rather than make that logic more complicated, I added this hack.

caikit/core/data_model/dataobject.py
joerunde: @gabe-l-hart yup! Most of our current usages of it are broken and don't create a message which is super fun.

If we're up for changing that I'd spin up a quick PR to make it match how logging works, like it was here before

prashantgupta24: ohh! This is going to cause changes in a lot of value checks! Let me create a PR for this on the side

caikit/core/data_model/base.py
```diff
@@ -82,176 +86,162 @@ def __new__(mcs, name, bases, attrs):
...
         attrs["_fields_primitive_repeated"] = _fields_primitive_repeated
-        attrs["_fields_map"] = _fields_map
         attrs["_proto_class"] = proto_class
+
         instance = super().__new__(mcs, name, bases, attrs)
 
-        # Update the global class and proto registries
-        if name != "DataBase":
-            mcs.class_registry[full_name] = instance
+        # If there's a valid proto class, perform proto descriptor parsing
```
joerunde, line 136: if there's not then this happens later, right?

caikit/core/data_model/dataobject.py
joerunde: Any chance that context can be thrown into a comment?

caikit/core/data_model/dataobject.py
prashantgupta24: Interesting! Yeah we saw this issue with BAM debugging too! Added a failed test here: https://github.com/caikit/caikit/pull/94. Can work on the fix as well

joerunde (approved):
sweet, let's get this merged to unblock everybody waiting on the next round of breaking apis

PR #93, opened 2023-05-04
:bug: check for socket file first

### LINKED ISSUE #92: Can we not E0503...os_error:"No such file or directory all the time?  Is it just me?
## Describe the bug

I get an error, but I don't think it's a real error.

E0503 17:25:08.082902000 140704599909184 chttp2_server.cc:1045]        UNKNOWN:No address added out of total 1 resolved for 'unix:///tmp/mmesh/grpc.sock' {created_time:"2023-05-03T17:25:08.082045-07:00", children:[UNKNOWN:Unable to configure socket {fd:6, created_time:"2023-05-03T17:25:08.080064-07:00", children:[UNKNOWN:No such file or directory {created_time:"2023-05-03T17:25:08.078961-07:00", errno:2, os_error:"No such file or directory", syscall:"bind"}]}]}

I'm looking for cleaner output for demo, but really logging this as a bug because bogus but serious looking error message can cause everybody that has trouble getting started to have to deal with triage of this one.

Unless it is just me.  Then what might I be doing wrong?

## Platform

- Interpreter version: 3.9
- Library version: 0.2.0

## Sample Code

from caikit.runtime import grpc_server
grpc_server.main()

## Expected behavior

Library users should not see bogus errors that look like real errors.  These get in the way every time someone has to troubleshoot something.

## Observed behavior

see above

joerunde:
Closes #92 

This checks if the socket path exists before trying to bind to it. Should reduce the amount of error logs when sockets aren't being used

### DIFF
--- a/caikit/runtime/grpc_server.py
+++ b/caikit/runtime/grpc_server.py
@@ -148,21 +148,24 @@ class RuntimeGRPCServer:
         health_pb2_grpc.add_HealthServicer_to_server(health_servicer, self.server)
 
         # Listen on a unix socket as well for model mesh.
-        try:
-            self.server.add_insecure_port(
-                f"unix://{self.config.runtime.unix_socket_path}"
-            )
-            log.info(
-                "<RUN10001011I>",
-                "Caikit Runtime is communicating through address: unix://%s",
-                self.config.runtime.unix_socket_path,
-            )
-        except RuntimeError:
-            log.info(
-                "<RUN10001100I>",
-                "Binding failed for: unix://%s",
-                self.config.runtime.unix_socket_path,
-            )
+        if self.config.runtime.unix_socket_path and os.path.exists(
+            self.config.runtime.unix_socket_path
+        ):
+            try:
+                self.server.add_insecure_port(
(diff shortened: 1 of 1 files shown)


evaline-ju (approved):
LGTM!

caikit/runtime/grpc_server.py
```diff
@@ -148,21 +148,24 @@ def __init__(
...
-                "Caikit Runtime is communicating through address: unix://%s",
-                self.config.runtime.unix_socket_path,
-            )
-        except RuntimeError:
-            log.info(
-                "<RUN10001100I>",
-                "Binding failed for: unix://%s",
-                self.config.runtime.unix_socket_path,
-            )
+        if self.config.runtime.unix_socket_path and os.path.exists(
```
markstur, line 151: I did a quick test by copying this change into my lib and it works, thanks!

I also looked a bit for possible None reference and this seems safe (self.config.runtime.port is used earlier).

markstur (approved):
LGTM thank you!

Not sure what my "Approve" button does here, but I'll try it.

PR #94, opened 2023-05-04
🐛 fix value checks

prashantgupta24:


caikit/core/toolkit/errors/error_handler.py
```diff
@@ -273,7 +273,7 @@ def type_check_all(self, log_code, *types, allow_none=False, **variables):
                         ),
                     )
 
-    def value_check(self, log_code, condition, *args, msg=""):
+    def value_check(self, log_code, condition, msg, *args):
```
prashantgupta24, line 276: I couldn't image a scenario in which msg is an empty string so removed the default value for it.

gabe-l-hart: I think the idea was if you just didn't want a message, you didn't need one

gabe-l-hart (approved):
It might be worth adding the default value for `msg` back in, but otherwise, this is a nice cleanup

prashantgupta24:
> It might be worth adding the default value for `msg` back in, but otherwise, this is a nice cleanup

Adding a default to the `msg` arg fails the linter - `keyword argument before variable positional arguments list in the definition of value_check function `

gabe-l-hart:
>Adding a default to the msg arg fails the linter - keyword argument before variable positional arguments list in the definition of value_check function 

Ah! that actually makes so much sense and explains why it was done the way it was 🤦. You _could_ accomplish this by not having a standalone `msg` arg and instead just taking `*args` and if it's length 1, it's a message, with anything in `args[1:]` taken as format expansions.

prashantgupta24:
> > Adding a default to the msg arg fails the linter - keyword argument before variable positional arguments list in the definition of value_check function
> 
> Ah! that actually makes so much sense and explains why it was done the way it was 🤦. You _could_ accomplish this by not having a standalone `msg` arg and instead just taking `*args` and if it's length 1, it's a message, with anything in `args[1:]` taken as format expansions.

Great idea, tried that out

caikit/core/toolkit/errors/error_handler.py
```diff
@@ -291,19 +291,25 @@ def value_check(self, log_code, condition, *args, msg=""):
...
+                format syntax by passing additional arguments.  This is the preferred method for
+                performing string interpolation on `msg` so that it is only done if an error
+                condition is encountered.
+
         """
         if not get_config().enable_error_checks:
             return
 
+        if len(args) == 1:
+            msg = "{}"
```
gabe-l-hart, line 308: I would thing this would look more like the following:

```py
if not condition:
    msg = (
        "" if not args else (
            args[0] if len(args) == 1 else args[0].format(args[1:])
        )
    )
```

This way, all the `if` conditions are only evaluated after the check on `condition`. Also, no extra formatting is done (even "empty" formatting is not free).

One advantage to the way you have it written is that `args[0]` could be something besides a string and it would be stringified by `msg.format`. Not sure that was intentional, though?

gabe-l-hart (commented):
One more thought about how `msg` and interpolation work

caikit/core/toolkit/errors/error_handler.py
prashantgupta24: Nope, wasn't my intention to stringify non-string params (do we need to?)

And yep incorporating all checks within the condition makes total sense!

prashantgupta24: Had to tweak the above code to `format(*args[1:])` for unpacking the args

prashantgupta24: Although this still fails if the user provides 2 args, neither of which are supposed to be strings.
```
bad_value = 1.1
self.error.value_check("<CCC>", 0.0 <= bad_value <= 1.0, bad_value, "<1.0")
```
results in an error since we try to format assuming `bad_value` as `msg`.

prashantgupta24: We can additionally check if the type of arg is `str` before trying to `.format` with it?

gabe-l-hart (approved):

PR #95, opened 2023-05-04
FixEnumDeref: When converting an Enum value to a serialized format, d…

gabe-l-hart:
…ereference it


joerunde (approved):
lpgtm, just needs fmt and lint?

PR #96, opened 2023-05-05
♻️ fix example, switch to DataObjectBase

prashantgupta24:
Switch to the new `DataObjectBase` for the `text-sentiment` example


evaline-ju (approved):
LGTM - thanks for updating!

PR #97, opened 2023-05-08
✨ 💥   Add shared train/load backends

joerunde:
This PR is the beginning of @gabe-l-hart and @tjohnson31415 's vision of the future of backends

This is a **breaking configuration change**

This:
- Adds `SharedLoadBackend` and `SharedTrainBackend` base classes and the `local` implementations of them
- Consolidates the `module_backends.priority` and `module_backends.config` configurations as a list of objects that include the backend type and its config
- Splits the backend priorities between training and loading, `module_backends.load_priority` and `module_backends.train_priority`
- Allows multiple backends of the same type to be registered under a different name and with different configurations
- Deletes the `module_backends.enabled` flag that was only used by the runtime. (It was buggy anyway, since the core loader would run backend configuration even if `module_backends.enabled` was False)

This does not yet implement a `caikit.train` method that can use an appropriate shared train backend similar to how `caikit.load` does.

As a followup design issue, the split between "shared" and the original "not shared" backends is pretty awkward. We'd like to explore if the original backends are really required or if they can be posed as shared backends as well. This would probably mean walking back the original "distributed module" pattern where multiple backend implementations are registered against the same guid as a base module

### DIFF
--- a/caikit/core/model_manager.py
+++ b/caikit/core/model_manager.py
@@ -187,6 +187,12 @@ class ModelManager:
                             module_path,
                             load_backend.name,
                         )
+                        error.type_check(
+                            "<COR76726077E>",
+                            ModuleBase,
+                            model=model,
+                        )
+
                         loaded_model = model
                         model.set_load_backend(load_backend)
                         break
--- a/caikit/core/model_manager.py
+++ b/caikit/core/model_manager.py
@@ -178,6 +178,7 @@ class ModelManager:
             # directly. If not, parse the module config and look to see if there is
             # a version of the module available for the given backend
             loaded_model = None
+            log.debug("Available load backends: %s", configured_load_backends)
             for load_backend in configured_load_backends:
                 # If this is a shared loader, try loading the model directly
                 if isinstance(load_backend, SharedLoadBackendBase):
--- a/caikit/core/model_manager.py
+++ b/caikit/core/model_manager.py
@@ -155,6 +155,8 @@ class ModelManager:
 
             # Using the module_path as a key, look for an instance preloaded in the
             # singleton cache if desired
+            # 🌶🌶🌶 This doesn't work for nested blocks
+            # TODO: think about bringing back the `unique_hash` or `tracking_id`
             if singleton_entry := (
                 load_singleton and self.singleton_module_cache.get(module_path)
             ):
--- a/caikit/core/model_manager.py
+++ b/caikit/core/model_manager.py
@@ -166,7 +166,7 @@ class ModelManager:
             if not configured_load_backends:
                 log.info(
                     "<COR56759744I>",
-                    "No backend configured! Trying to configure using default config file.",
+                    "No backends configured! Configuring backends with current configuration",
                 )
                 module_backend_config.configure()
                 configured_load_backends = (
--- a/caikit/core/model_manager.py
+++ b/caikit/core/model_manager.py
@@ -148,7 +148,9 @@ class ModelManager:
 
             # Using the module_path as a key, look for an instance preloaded in the
             # singleton cache if desired
-            if singleton_entry := self.singleton_module_cache.get(module_path):
+            if singleton_entry := (
+                load_singleton and self.singleton_module_cache.get(module_path)
+            ):
                 log.debug("Found %s in the singleton cache", module_path)
                 return singleton_entry
 
--- a/caikit/core/model_manager.py
+++ b/caikit/core/model_manager.py
@@ -21,6 +21,7 @@ from contextlib import contextmanager
 from io import BytesIO
 from threading import Lock
 from typing import Union
+import errno
 import os
 import tempfile
 import zipfile
@@ -141,6 +142,12 @@ class ModelManager:
             subclass of blocks.base.BlockBase
                 Model object that is loaded, configured, and ready for prediction.
         """
+        # Short-circuit the loading process if the path does not exist
+        if not os.path.exists(module_path):
+            raise FileNotFoundError(
+                errno.ENOENT, os.strerror(errno.ENOENT), module_path
+            )
+
         # If this is a singleton load, the entire body of this function needs to
         # be locked to avoid concurrent loads on the same model. Otherwise, we
         # can freely load in parallel.
@@ -199,12 +206,11 @@ class ModelManager:
                         loaded_model = model
                         model.set_load_backend(load_backend)
                         break
-                    else:
-                        log.debug3(
-                            "Could not load %s with loader %s",
-                            module_path,
-                            load_backend.name,
-                        )
+                    log.debug3(
+                        "Could not load %s with loader %s",
+                        module_path,
+                        load_backend.name,
+                    )
 
                 # If this is not a shared loader, look for an implementation of the
                 # model's module that works with this backend
--- a/caikit/core/model_manager.py
+++ b/caikit/core/model_manager.py
@@ -34,11 +34,12 @@ from .module import (
     _MODULE_TYPES,
     MODULE_BACKEND_REGISTRY,
     MODULE_REGISTRY,
-    SUPPORTED_LOAD_BACKENDS_VAR_NAME,
     ModuleBase,
     ModuleConfig,
 )
-from .module_backends import SharedLoadBackendBase, backend_types
+from .module_backends import backend_types
+from .module_backends.base import SharedLoadBackendBase
+from .module_type import SUPPORTED_LOAD_BACKENDS_VAR_NAME
 from .toolkit.errors import error_handler
 from caikit.config import get_config
 
--- a/caikit/core/model_manager.py
+++ b/caikit/core/model_manager.py
@@ -17,7 +17,9 @@
 """
 
 # Standard
+from contextlib import contextmanager
 from io import BytesIO
+from threading import Lock
 from typing import Union
 import os
 import tempfile
@@ -27,7 +29,7 @@ import zipfile
 import alog
 
 # Local
-from . import module_backend_config as backend_config
+from . import module_backend_config
 from .module import (
     _MODULE_TYPES,
     MODULE_BACKEND_REGISTRY,
@@ -36,7 +38,7 @@ from .module import (
     ModuleBase,
     ModuleConfig,
 )
-from .module_backends import backend_types
+from .module_backends import SharedLoadBackendBase, backend_types
 from .toolkit.errors import error_handler
 from caikit.config import get_config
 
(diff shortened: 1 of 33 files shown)


evaline-ju (commented):
Due to the size of this PR, definitely just glossing over all the moved code in `module.py`

caikit/core/module_type.py
```diff
@@ -0,0 +1,308 @@
...
+
+                # Verify UUID and add this block to the module and block registries
+                global MODULE_REGISTRY
+                current_class = MODULE_REGISTRY.get(cls_.MODULE_ID)
+                if not backend_module_impl:
+                    if current_class is not None:
+                        error(
+                            "<COR30607646E>",
+                            RuntimeError(
+                                "BLOCK_ID `{}` conflicts for classes `{}` and `{}`".format(
```
evaline-ju, line 217: module?

tests/core/test_model_manager.py
```diff
@@ -419,27 +392,26 @@ def test_fall_back_to_local(reset_globals):
...
-        assert isinstance(model, NonDistributedBlock)
+    assert isinstance(model, NonDistributedBlock)
 
 
 def test_no_local_if_disabled(reset_globals):
     """Make sure that if LOCAL is disabled and a given module doesn't have any
     registered backends, loading fails.
     """
     _ = setup_saved_model(MockBackend)
+    # 🌶️ TODO: remove the `disable_local` flag
```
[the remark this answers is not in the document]
gabe-l-hart: I think the goal is to eventually remove all "special" handling for `LOCAL` and just have the default values in `config.yml` pre-populated with `LOCAL`. Could certainly argue either way, though

caikit/core/model_manager.py
```diff
@@ -137,150 +142,153 @@ def _load_from_dir(self, module_path, load_singleton, *args, **kwargs):
...
-                log.debug(
-                    "%s backend implementation found for backend [%s]",
-                    backend_impl.__name__,
-                    backend,
+        # If this is a singleton load, the entire body of this function needs to
+        # be locked to avoid concurrent loads on the same model. Otherwise, we
+        # can freely load in parallel.
+        with self.singleton_lock(load_singleton):
+
+            # Using the module_path as a key, look for an instance preloaded in the
```
gabe-l-hart, line 156: I know we talked about reviving the `unique_id` content hash for the case where the same content is being loaded from multiple paths. Just noting that here for posterity

tests/core/test_model_manager.py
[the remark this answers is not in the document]
joerunde: right, needs more context.

The idea was to get `aconfig` to handle lists so that we could set `LOCAL` in the base config and allow users to merge more priorities to the list ahead of `LOCAL`, instead of handling it separately in code here

caikit/core/module_backends/base.py
```diff
@@ -54,3 +59,67 @@ def start(self):
...
+class SharedTrainBackendBase(BackendBase, abc.ABC):
+    """Interface for a backend that can perform train on any given module
+
+    A Shared backend is one that treats the given module as a black box and
+    delegates the execution of that module's functionality to an alternate
+    execution engine.
+    """
+
+    @abc.abstractmethod
+    def train(self, module_class: Type[ModuleBase], *args, **kwargs) -> ModuleBase:
```
gkumbhat, line 73: Just thinking out the train use-case:
1. Would train data massaging / packaging also be responsibility of train backend impl class? Like in case of training in a torch cluster, if it requires data in torch data loaders, would the conversion from potentially stream in `caikit.runtime` to torch data loader be responsibility of impl of `SharedTrainBackend`?
2. Would we always be running `caikit.module` based training code in the backend? Or are we going to support non-caikit module training in backend as well?
3. Does error / exception handling responsibility of train function? Considering below `TODO` around async training, since `train` here can return future, would we need to expose another semantics for checking on the status, fetching error etc?
4.  How would the `.save` be handled and the case of automatic loading of the model, like the `auto_load_trained_model` case?

gkumbhat (commented):
Thanks @gabe-l-hart , @joerunde . Left some questions

caikit/core/module.py
```diff
@@ -510,7 +729,7 @@ def _extract_pred_set(self, dataset, *args, preprocess_func=None, **kwargs):
         )
 
     @staticmethod
-    def load_evaluation_dataset(dataset_path):
+    def _load_evaluation_dataset(dataset_path):
```
gabe-l-hart, line 732: @alex-jw-brooks I'm pretty sure you owned this originally. To the best of my knowledge, this function is an implementation detail that's only used in the default implementation of `evaluate_quality`. Is that right?

alex-jw-brooks: Hey @gabe-l-hart! As far as I'm aware, that is correct

caikit/core/module_backends/base.py
gabe-l-hart: >Would train data massaging / packaging also be responsibility of train backend impl class? Like in case of training in a torch cluster, if it requires data in torch data loaders, would the conversion from potentially stream in caikit.runtime to torch data loader be responsibility of impl of SharedTrainBackend?

I have a hard time imagining a `SharedTrainBackend` that wraps running in a `torch` cluster since that would only work for module implementations that are backed by `torch`. I would expect such usages to be non-shared backends that require backend-specific implementations for a remote `torch` cluster. Realistically, the data conversions would probably be very similar across models, so the backend could expose any utility functions needed to make this code reusable.

In general, though, for a truly `Shared` train backend, yes, any data massaging would be the responsibility of the backend itself. For example, a shared train backend that delegates to a remote instance of `caikit.runtime` would need to serialize the python args into a train request `grpc` call.

>Would we always be running caikit.module based training code in the backend? Or are we going to support non-caikit module training in backend as well?

Good question. I think the baseline is that we would always have any code encapsulated as a `caikit` module. It could be totally fair game to have a generic module that allows for arbitrary training against a given platform, though. Again, something like this would probably not be a `Shared` backend, though, since the only generic interface that a `SharedTrainBackend` can rely on is `ModuleBase`.

>Does error / exception handling responsibility of train function? Considering below TODO around async training, since train here can return future, would we need to expose another semantics for checking on the status, fetching error etc?

Currently this does not return a future, but we absolutely will need to have careful error handling semantics down the line. Since the goal of a `SharedTrainBackend` is to encapsulate platform-specific concerns for a given backend platform, I think we will likely want to define a set of standard training error flows and make sure that all `SharedTrainBackend` implementations follow them.

>How would the .save be handled and the case of automatic loading of the model, like the auto_load_trained_model case?

This will be the crux of the `orchestration` design once we get there. We'll need to make sure that the train and load backends are able to interoperate cleanly to enable `auto_load_trained_model` semantics, even when the model is being trained and loaded on divergent backends. Certainly some design work to do here to figure out how to validate configurations and perform the model handoff from training to inference.

>From runtime perspective, can a shared backend in itself do distribution of training (internal distribution) and manage its own cluster?

By definition, a `SharedTrainBackend` is performing `external` distribution (the module is the atomic unit being distributed). Hypothetically a given shared backend _could_ also be used for `internal` distribution (distribution of parts of the workload inside a given module). For example, I could imagine a generic `ray` backend that holds the connection to a cluster and can perform distribution of modules as `remote` (`actor`, `job`?) primitives as a `SharedTrainBackend`, but also allows individual `ray`-based module implementations to grab the client and perform internal distribution.

As for "and manage its own cluster," this is an interesting question. My gut feeling is that or anything that will be running in a truly "remote" sense, the deployment topology should live outside the backend implementation to avoid mixing runtime and deploy-time concerns. That said, you could certainly imagine something like how [`caikit-tgis-backend` runs its subprocess](https://github.com/caikit/caikit-tgis-backend/blob/main/caikit_tgis_backend/tgis_backend.py#L82) for cluster-based backends that want to be able to operate in "build your own" mode.

tests/core/test_module.py
```diff
@@ -474,13 +402,13 @@ class DummyLocal(caikit.core.blocks.base.BlockBase):
...
         pass
 
     @caikit.core.blocks.block(
         backend_type=backend_types.FOO, base_module=DummyLocal.MODULE_ID
     )
-    class DummyFoo(caikit.core.ModuleBase):
+    class DummyFoo(caikit.core.blocks.base.BlockBase):
         pass
 
     assert DummyLocal in list(caikit.core.MODULE_REGISTRY.values())
```
gabe-l-hart, line 414: I think we need to add a new test to exercise the ability to disambiguate multiple instances of the same backend by name. I'll add that now

tests/core/helpers.py
```diff
@@ -39,6 +50,38 @@ def stop(self):
...
+    def load(self, model_path: str, *args, **kwargs) -> Optional[ModuleBase]:
+        # allow config.model_type to control whether this loader barfs
+        if "model_type" in self.config and "model_type" in kwargs:
+            if self.config["model_type"] != kwargs["model_type"]:
+                # Don't load in this loader
+                return None
+        # use the "Local" loader to actually load the model
+        model = LocalBackend(name="test").load(model_path)
+        # Let tests know which loader was used
+        setattr(model, "loader_name", self.name)
```
gabe-l-hart, line 69: Wouldn't the call to `set_load_backend` take care of this?

tests/core/test_model_manager.py
```diff
@@ -529,3 +503,111 @@ def load(self, *args, **kwargs):
...
+                "load_priority": [
+                    {"type": TestLoader.backend_type, "name": loader_name},
+                    {"type": backend_types.LOCAL},
+                ],
+            }
+        }
+    ):
+        configure()
+        model = caikit.core.load(good_model_path)
+        assert model.loader_name == loader_name
```
gabe-l-hart, line 542: I think this could be `model.get_load_backend.name == loader_name` and skip that extra `setattr`

gabe-l-hart (commented):
@joerunde a couple of tiny NITs on the tests using `TestLoader`, but otherwise, a huge thanks for getting this fleshed out!

caikit/core/module_backends/base.py
gkumbhat: Thanks Gabe.  

I think we would need to document above definitions as "parameters" for designing shared backends, so that not everything ends up being one-off shared backends and the difference between shared and non-shared backend is evident.

gabe-l-hart: 100% agree. We probably need a pretty robust ADR around this. I can start writing one up today, but we need to sort out numbering at some point @joerunde

gabe-l-hart (approved):
From my view, this looks ready to go. Then again, I wrote a lot of it, so let's get signoff from one other maintainer before hitting merge!

tests/core/test_model_manager.py
joerunde: hot

joerunde:
I can't sign off because I opened it 🙃

gabe-l-hart:
@joerunde @evaline-ju I think verbal signoff is plenty! If you're good, let's merge it.

PR #98, opened 2023-05-09
🐛 handle ctrl chars in datastream

prashantgupta24:
Handle escaped control characters in the server while loading training data.

The current code handles control characters as part of the string, for example `"name \t ": "John"` works well. But introducing new line characters as part of a `jsonl` file is what's causing problems at the moment.

### DIFF
--- a/caikit/core/data_model/streams/data_stream.py
+++ b/caikit/core/data_model/streams/data_stream.py
@@ -176,7 +176,7 @@ class DataStream(Generic[T]):
 
                 try:
                     for line in lines:
-                        if line != b"\n":  # ignore new line characters
+                        if line.strip():  # ignore empty lines
                             yield json.loads(line)
                 except json.JSONDecodeError as e:
                     error(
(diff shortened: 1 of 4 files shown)


caikit/core/data_model/streams/data_stream.py
```diff
@@ -176,7 +176,8 @@ def generator_func():
 
                 try:
                     for line in lines:
-                        yield json.loads(line)
+                        if line != b"\n":  # ignore new line characters
```
prashantgupta24, line 179: there has to be a better way to do this?

gabe-l-hart: Can you just do `if not line.strip(): continue`?

prashantgupta24: Since we open the file in binary mode (I'm still not sure I understand why), `line.strip` gives me `b''` which is why I wasn't able to use that and I had to resort to checking `line != b"\n"` explicitly.

Issue #53, opened 2023-04-21
Rename/refactor `serializers`


joerunde:
## Is your feature request related to a problem? Please describe.

We previously had these protobuf serializers in the service_generation package which would write protobuf files.

We offloaded the protobuf serialization support to `jtd_to_proto`, but kept these classes around and hijacked them to be the containers to hold data for our conversions from python functions to both 
1. `@dataobject` classes for the request messages
1. RPC methods for the service definition

## Describe the solution you'd like

These "serializers" (TaskPredictRPC, ModuleTrainRPC) should be renamed, and should probably own the logic to do like:
```
def to_data_model() -> Type[DataBase]:
    # return the data model class describing the request message

def to_service_json() -> Dict:
    # return some json snippet for the service deinition
```

currently that logic is a big mess in the service factory

## Describe alternatives you've considered

/shrug

## Additional context

We'd like to remove any and all `protobuf` logic from the caikit code

PR #99, opened 2023-05-10
Refactor serializers (and service factory)

aluu317:
This PR:
- Moves the logic for the below from `service_factory` into `serializers`:
```
def create_request_data_model() -> Type[DataBase]:
    # return the data model class describing the request message

def create_rpc_json() -> Dict:
    # return json snippet for the service definition of this RPC
```
- Adds a few more tests for `service_factory` test coverage

Closes https://github.com/caikit/caikit/issues/53

### DIFF
--- a/caikit/runtime/service_factory.py
+++ b/caikit/runtime/service_factory.py
@@ -15,8 +15,7 @@
 # Standard
 from enum import Enum
 from types import ModuleType
-from typing import Callable, Dict, List, Optional, Set, Type
-import copy
+from typing import Callable, Set, Type
 import dataclasses
 import inspect
 
@@ -26,10 +25,6 @@ import google.protobuf.service
 import grpc
 
 # First Party
-from py_to_proto.dataclass_to_proto import (  # NOTE: Imported from here for compatibility
-    Annotated,
-    FieldNumber,
-)
 from py_to_proto.json_to_service import (
     json_to_service,
     service_descriptor_to_client_stub,
@@ -41,12 +36,6 @@ import alog
 
 # Local
 from caikit import get_config
-from caikit.core.data_model.base import DataBase
-from caikit.core.data_model.dataobject import (
-    DataObjectBase,
-    _DataObjectBaseMetaClass,
-    dataobject,
-)
 from caikit.core.module import ModuleBase
 from caikit.interfaces.runtime.data_model import (
     TrainingInfoRequest,
@@ -54,10 +43,7 @@ from caikit.interfaces.runtime.data_model import (
 )
 from caikit.runtime import service_generation
 from caikit.runtime.service_generation.core_module_helpers import get_module_info
-from caikit.runtime.service_generation.serializers import (
-    RPCSerializerBase,
-    snake_to_upper_camel,
-)
+from caikit.runtime.service_generation.serializers import snake_to_upper_camel
 from caikit.runtime.types.caikit_runtime_exception import CaikitRuntimeException
 from caikit.runtime.utils import import_util
 import caikit.core
@@ -216,18 +202,13 @@ class ServicePackageFactory:
                 task_rpc_list = service_generation.create_training_rpcs(clean_modules)
                 service_name = f"{ai_domain_name}TrainingService"
 
-            for rpc in task_rpc_list:
-                if rpc.return_type is None:
-                    # TODO: need to hook up the excluded tasks / modules configs and add some
-                    # more handling here to ensure good RPCs generated
-                    log.info("Skipping rpc %s, no return type on method!", rpc.name)
             task_rpc_list = [
                 rpc for rpc in task_rpc_list if rpc.return_type is not None
             ]
 
-            request_data_models = cls._create_request_message_types(
-                task_rpc_list, package_name
-            )
+            request_data_models = []
+            for rpc in task_rpc_list:
+                request_data_models.append(rpc.create_request_data_model(package_name))
 
             client_module = ModuleType(
                 "ClientMessages",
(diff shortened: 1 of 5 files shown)


caikit/runtime/service_factory.py
```diff
@@ -216,18 +202,13 @@ def get_service_package(
...
             task_rpc_list = [
                 rpc for rpc in task_rpc_list if rpc.return_type is not None
             ]
 
-            request_data_models = cls._create_request_message_types(
-                task_rpc_list, package_name
-            )
+            request_data_models = []
+            for rpc in task_rpc_list:
+                request_data_models.append(rpc.create_request_data_model(package_name))
```
joerunde, line 211:
```
request_data_models = [
    rpc.create_request_data_model(package_name) for rpc in task_rpc_list
]
```
?

caikit/runtime/service_factory.py
```diff
@@ -238,7 +219,10 @@ def get_service_package(
                 # We need the message class that data model serializes to
                 setattr(client_module, dm_class.__name__, type(dm_class().to_proto()))
 
-            service_json = cls._create_service_json(task_rpc_list, package_name)
+            rpc_jsons = []
+            for rpc in task_rpc_list:
+                rpc_jsons.append(rpc.create_rpc_json(package_name))
```
joerunde, line 224: could probably replace with a list comprehension as well

caikit/runtime/service_factory.py
```diff
@@ -323,71 +307,6 @@ def _get_and_filter_modules(
...
-    @staticmethod
-    def _create_service_json(
-        rpcs_list: List[RPCSerializerBase], package_name: str
-    ) -> Dict:
-        """Make a json service def out of some rpc defs"""
-        rpc_jsons = []
-        for task in rpcs_list:
-            # The return type should be a "data model" object.
-            # We can take advantage of the fact that all of these should contain a private
-            # `proto_class` field ...and use that to yoink the fully qualified name of the
```
joerunde, line 377: hah, that was a pretty old comment

caikit/runtime/service_factory.py
joerunde: oh man, just re-reading this code and the logic for creating the `package_name` is gnarly lol.

Gotta think about that one more, maybe when we introduce `@task`s they should come with a package name?

caikit/runtime/service_generation/create_service.py
```diff
@@ -55,6 +55,7 @@ def create_inference_rpcs(modules: List[Type[ModuleBase]]) -> List[RPCSerializer
 
     rpcs = []
     # Inference specific logic:
+    # Remove non-primitive modules (including modules that return None type)
```
joerunde, line 58: cannot wait to not have to do this 🙃

caikit/runtime/service_generation/serializers.py
```diff
@@ -50,6 +61,52 @@ def module_list(self) -> List[Type[ModuleBase]]:
     def request(self) -> "_RequestMessage":
         """Return the internal representation of the request message type for this RPC"""
 
+    def create_request_data_model(self, package_name: str) -> Type[DataBase]:
+        """Dynamically create data model for this class's input RPC"""
```
joerunde, line 65: 🤔 
`Dynamically create data model for this RPC's request message`?

There might also be a fun edge case here where the class is created multiple times if this method is called multiple times. I don't know if that would be a problem or not, might be fun to at least throw in a test that makes sure nothing blows up, if one doesn't already exist

joerunde: (I'm guessing it's okay to do since this class would be stored in some `locals` that will be different on each invocation)

tests/runtime/test_service_factory.py
```diff
@@ -63,6 +63,18 @@ def compiled_training_service() -> ServicePackage:
     )
 
 
+def test_service_package_raises_for_compiled_training_management():
+    with pytest.raises(CaikitRuntimeException) as e:
+        ServicePackageFactory.get_service_package(
+            ServicePackageFactory.ServiceType.TRAINING_MANAGEMENT,
+            ServicePackageFactory.ServiceSource.COMPILED,
+        )
+    assert (
```
joerunde, line 72: something cool I found the other day is that `pytest.raises` has a `match` keyword that takes regexes to match on the exception message so that you don't have to do this!

```
with pytest.raises(CaikitRuntimeException, match="Not allowed to get Training Management services from compiled packages"):
```

tests/runtime/test_service_factory.py
```diff
@@ -313,6 +323,5 @@ def test_get_and_filter_modules_respects_included_task_types_and_excluded_module
     ) as cfg:
 
         clean_modules = ServicePackageFactory._get_and_filter_modules(cfg, "sample_lib")
-        assert len(clean_modules) == 4
```
joerunde, line 316: nice, good to see tests more robust against test fixture changes 🌶️

joerunde:
Looks great!

We should get some better names in for these, maybe a starting point is to rename:
`serializers.py` -> `rpcs.py`
`RPCSerializerBase` -> `CaikitRPC`
?

caikit/runtime/service_factory.py
aluu317: Now that you mention. I think this logic is redundant:
```
ai_domain_name = snake_to_upper_camel(lib.replace("caikit_", ""))
```
if our `lib` is sample_lib. This might have been more for our internal use case.

I think we need to revisit this when `task decorator` is introduced

joerunde (approved):
j'approuve, these improvements are 🌭🌭🌭

PR #100, opened 2023-05-10
Shared backends adr

gabe-l-hart:
## Description

This PR introduces two ADRs related to the [shared backend PR changes](https://github.com/caikit/caikit/pull/97)


docs/adrs/019-loader-stack.md
```diff
@@ -0,0 +1,20 @@
+# ADR 019: Loader Stack
+
+When loading a model in a non-local backend, there are several ways to map a given set of model artifacts to a concrete `ModuleBase` derived class that can perform the `load` operation. With the introduction of [shared load backends](./018-shared-backends.md), we could either require that _any_ shared backend be able to load _all_ models, or we could require that _some_ shared backend be able to load _any_ model. This ADR proposes that we do the later in order to support a mixed set of loaders that can combine to support the set of modules a given usage requires.
```
evaline-ju, line 3:
```suggestion
When loading a model in a non-local backend, there are several ways to map a given set of model artifacts to a concrete `ModuleBase` derived class that can perform the `load` operation. With the introduction of [shared load backends](./018-shared-backends.md), we could either require that _any_ shared backend be able to load _all_ models, or we could require that _some_ shared backend be able to load _any_ model. This ADR proposes that we do the latter in order to support a mixed set of loaders that can combine to support the set of modules a given usage requires.
```

evaline-ju (approved):

PR #104, opened 2023-05-11
Add train cancel handler

gkumbhat:
### Changes
- Add training job executor base class with 2 implementations:
    1. Handling training via threads
    2. Handling training via sub-process
- Add `event` concept to notify when the training is cancelled and needs to be terminated
- Add handling of cancellation of training request for both types of executors

### Diagram of process / thread handling

image

### Some design considerations
1. Didn't use `ProcessPoolExecutor` similar to `Threadpool` executor since `ProcessPoolExecutors` do not allow `fork`
2. `ThreadPoolExecutors` provide some niceties in API and some ease of use, so use `ThreadPoolExecutor` instead of custom Thread  
3. Currently avoided creating custom futures and pool executors as that all might need to change with SharedBackend concept
4. To handle subprocess termination in cases where OS kills the process along with potential cancellation from client side, I have added a mechanism that along with subprocess, we will kick off a parallel thread that will handle cancellation in parallel to `worker.join()`  allowing us to kill the subprocess in parallel if requested

### TODOs
1. Event creation for different executors is a bit manual currently. We'll work on moving that part as a function of the executor


gkumbhat:
The conflict is easy to resolve, but rebase / merging with main seems to be opening whole can of errors, so leaving it as is for now.

caikit/runtime/work_management/train_executors.py
```diff
@@ -0,0 +1,275 @@
...
+            # Save it
+            with alog.ContextTimer(
+                log.debug,
+                "Done saving %s to %s in: ",
+                module_class.__name__,
+                model_path,
+            ):
+                model.save(model_path)
+
+        except Exception as ex:
```
gabe-l-hart, line 83: This can go away

caikit/runtime/work_management/train_executors.py
```diff
@@ -0,0 +1,275 @@
...
+
+        self._worker = self._ErrorCaptureProcess(
+            target=TrainSaveExecutorBase.train_and_save,
+        )
+
+        self.__proc_id = self._worker.pid
+        self.__event = event
+
+    def __del__(self):
+        self.cancel()
```
gabe-l-hart, line 220: Let's add a docstring here to explicitly say that this is NOT how cancelation should be invoked, but that it's here to make sure cleanup happens

caikit/runtime/model_management/training_manager.py
```diff
@@ -40,6 +40,12 @@ def get_training_status(self, model_id: str) -> TrainingStatus:
 
         raise RuntimeError("Unexpected error")
 
+    def cancel_training(self, model_id: str):
```
gkumbhat, line 43: This isn't needed. Will get this removed

caikit/runtime/work_management/train_executors.py
```diff
@@ -0,0 +1,275 @@
...
+            with alog.ContextTimer(
+                log.debug,
+                "Done saving %s to %s in: ",
+                module_class.__name__,
+                model_path,
+            ):
+                model.save(model_path)
+
+        except Exception as ex:
+            raise ex
```
gkumbhat, line 84: Can be removed

caikit/runtime/servicers/global_train_servicer.py
```diff
@@ -209,26 +225,37 @@ def run_training_job(
...
+            runnable_executor=target,
             kwargs=kwargs,
             model_name=model_name,
             model_path=model_path,
         )
+
         self.training_map[training_id] = thread_future
 
+        # Add callback to register cancellation of training
+        def cancel_future(*args, **kwargs):
```
gabe-l-hart, line 252: I think it's a little misleading to call this `cancel_future` since it will _always_ be called on termination. It would be good to make the semantics clear here that this is a termination callback that will perform cancelation if termination happens in an abnormal way

gabe-l-hart (changes requested):
A few little nit/doc requests, but this looks great. This will likely change a bit once full shared train backend support is added, but let's get this in now and iterate

caikit/runtime/servicers/global_train_servicer.py
```diff
@@ -44,17 +43,19 @@
     snake_to_upper_camel,
     validate_data_model,
 )
+from caikit.runtime.work_management.train_executors import (
+    LocalTrainSaveExecutor,
```
gabe-l-hart, line 47: Once I've got the shared train backend work in place, I think these will move over to `caikit.core` as local shared train backends. Nothing to do now, just noting it.

caikit/runtime/servicers/global_train_servicer.py
```diff
@@ -114,7 +121,7 @@ def Train(self, request, *_, **__) -> TrainingJob:
 
         try:
             with alog.ContextLog(log.debug, outer_scope_name):
-                # BlocksSampleTaskSampleModuleTrainRequest
+                # ModulesSampleTaskSampleModuleTrainRequest
```
gabe-l-hart, line 124: This looks like dead code?

caikit/runtime/servicers/global_train_servicer.py
```diff
@@ -162,6 +172,15 @@ def Train(self, request, *_, **__) -> TrainingJob:
...
             raise e
 
+        except concurrent.futures.CancelledError as err:
+            log_dict = {
+                "log_code": "<RUN71530128W>",
+                "message": err.message,
+                "error_id": err.id,
+            }
+            log.warning({**log_dict, **e.metadata})
+            raise err
```
gabe-l-hart, line 182: Do we need to wrap this in `CaikitRuntimeException` to put a grpc error code on it?

caikit/runtime/servicers/global_train_servicer.py
```diff
@@ -98,13 +103,15 @@ def __init__(self, training_service: ServicePackage):
         )
         super()
 
-    def Train(self, request, *_, **__) -> TrainingJob:
+    def Train(self, request, context, *_, wait=False, **__) -> TrainingJob:
```
gabe-l-hart, line 106: When would we ever pass `wait` to this function? This is the rpc handler, so afaik we never actually call it directly?

gabe-l-hart (commented):
A couple of small NITs/questions around error handling

caikit/runtime/servicers/global_train_servicer.py
gkumbhat: This one is getting called for test to keep the connection open: https://github.com/gkumbhat/caikit/blob/add_train_cancel_handler/tests/runtime/servicers/test_global_train_servicer_impl.py#L438

caikit/runtime/servicers/global_train_servicer.py
gkumbhat: I think this is an example of how we get to the module name itself?

caikit/runtime/servicers/global_train_servicer.py
gkumbhat: actually we don't need this anymore. I'll remove it

caikit/runtime/servicers/global_train_servicer.py
gabe-l-hart: I think this needs to get removed per slack conversation

gabe-l-hart (changes requested):
Last fix should be to remove the `wait` from `Train`. So close!

gabe-l-hart (approved):
Ship it!

PR #111, opened 2023-05-15
✅ Add missing union typing import

alex-jw-brooks:
Adds `Union` import for type hints in one of tests


alex-jw-brooks:
Nice, thanks!

PR #112, opened 2023-05-15
Tests for enums

gabe-l-hart:
## Description

This PR encapsulates the Dependabot bump to `munch` (#108) and adds unit tests for `caikit.core.data_model.enums` to ensure that things work as expected.

There is one small functionality change to the global `import_enums` function. In my opinion, we should look at removing this function entirely as part of the effort to remove the reliance on project structure, but that wasn't the goal of this PR. For this PR, I just swapped to trying to find the `proto_enum` on the `"protobufs"` "module" from the `current_globals` rather than `caikit.core.data_model.protobufs`. The old implementation relied on all proto enums being pulled in from compiled protos using [`import_protobufs`](https://github.com/caikit/caikit/blob/main/caikit/core/data_model/protobufs/__init__.py#L27) which we're actively moving away from.


evaline-ju (approved):
LGTM - thanks for adding!

evaline-ju (approved):
LGTM - thanks for adding!

PR #114, opened 2023-05-16
:bug: install graphviz/dot

joerunde:
PR checks are having issues because:
- pydeps needs `dot` installed
- some weird bug with 1.12.5 crashes :(

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -37,7 +37,7 @@ commands = pylint caikit
 
 [testenv:imports]
 description = enforce internal import rules
-deps = pydeps
+deps = pydeps==1.12.3
 commands = ./scripts/check_deps.sh
 allowlist_externals = ./scripts/check_deps.sh
 skip_install = True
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -37,7 +37,7 @@ commands = pylint caikit
 
 [testenv:imports]
 description = enforce internal import rules
-deps = pydeps
+deps = pydeps==1.12.3
```
gabe-l-hart, line 40: Version range?

gabe-l-hart (approved):
Looks good. Might be better with a version range?

tox.ini
joerunde: maaaybe, but this thing literally released half an hour ago and bricked the builds lol. This version works and has the functionality we need so I'd kinda rather just stick to it for stability for this tiny lil check

joerunde:
The check lives!

PR #116, opened 2023-05-16
Task validation at decoration time

### LINKED ISSUE #115: Validate task modules at decoration time

When a `@module` is decorated with a `task=SomeTask`, the task class' validation should be run to ensure that the module is correct. 

The module decoration should fail to give good feedback to users that their module is incorrect.

We've been deferring this to the runtime service generation, which causes a whole lot of problems when modules are built and only later when the runtime doesn't work does it turn out that the module requires changes.

## Additional context

This would generally make users' code fail to import if their `run` functions are ill-specified for the given task

joerunde:
This PR updates the `@module` decorator to provide decoration-time task validation if a task is specified.

This contains some structural updates:
- `CaikitMethodSignature`s and the rest of the `signature_parsing` package have moved from the runtime to the core
- Modules have their `train` and `run` signatures inspected at decoration-time and those `CaikitMethodSignatures` are stored on the module class
- The inspection code no longer raises `CaikitRuntimeException`s, as it's now in the core (and was never in the path of a runtime request anyway)
- A bunch of the `SampleTask`s that were put onto mock modules for testing were deleted, because the modules didn't actually implement the task

This PR does not yet update the service generation code, that should greatly simplify once we can:
- roll through all modules and filter by ones with tasks
- group by the explicit task
- use the caikit method signatures that have already been added to the modules

Closes #115


caikit/core/module_type.py
```diff
@@ -238,12 +239,16 @@ def decorator(cls_):
...
                 else:
-                    if not task:
-                        # TODO: raise once @tasks are integrated into service generation
-                        # raise TypeError(f"task argument is required for @{cls}")
-                        pass
                     cls_.TASK_CLASS = task
 
+                # Parse the `train` and `run` signatures
+                cls_.RUN_SIGNATURE = CaikitMethodSignature(cls_, "run")
+                cls_.TRAIN_SIGNATURE = CaikitMethodSignature(cls_, "train")
```
prashantgupta24, line 246: this is 🌶️ ! I'm assuming we can use these signatures to send exact parameter types when making `run` or `train` calls through the servicers?

joerunde: We sure can!

And I still think the servicers should get a copy of the `ServicePackage` that they're serving anyway to have more 🌶️🌶️🌶️ info about what to do. It could be the case that somebody passes a model id to the incorrect task, and the servicer won't know that if it doesn't have a handle to the caikit rpc object that it's supposed to be serving, to see if it aligns with what the module says its signature is

caikit/core/module_type.py
gabe-l-hart, line 1: Since this file is fully removed in #119, what sequence are you imagining for these two PRs?

evaline-ju: Since this was the smaller and earlier PR, we were hoping to get this one merged first. Like most of the current functionality here, this additional functionality would move to the new `modules/decorator.py` in #119

tests/core/test_task.py
```diff
@@ -86,3 +86,84 @@ class Stuff(caikit.core.ModuleBase):
         pass
 
     assert Stuff.TASK_CLASS is None
+
```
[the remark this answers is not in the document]
joerunde: Technically every single test in the suite because `SampleModule` has `SampleTask` in its decorator!

We could go add an explicit example though

tests/core/module_backends/test_module_backend_config.py
```diff
@@ -192,7 +192,7 @@ def test_multiple_module_same_backend_configures(reset_globals):
     can override backend configurations"""
     # Register backend type
 
-    @module(id="foo", name="dummy base", version="0.0.1", task=SampleTask)
```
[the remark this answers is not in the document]
joerunde: yup!
We had toyed with requiring all `blocks` to have a `task`, but decided against it. These were all invalid tasks from that mistake

caikit/core/signature_parsing/docstrings.py
```diff
@@ -46,11 +45,9 @@ def get_return_type(fn: Callable) -> Optional[Type]:
...
     try:
         docstring = docstring_parser.parse(fn.__doc__)
-    except Exception as exc:
-        raise CaikitRuntimeException(
-            grpc.StatusCode.INVALID_ARGUMENT,
-            f"ParseError when parsing docstring for function: {fn.__name__}",
-        ) from exc
+    except docstring_parser.ParseError as e:
+        log.warning("Could not parse docstring: %s fn.__doc__ ", fn.__doc__, exc_info=e)
+        return None
```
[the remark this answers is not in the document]
joerunde: This is only used within the signature parsing package, and previously the control flow in places that called this was
```
try:
    return_type = get_return_type(...)
except CaikitRuntimeException:
    log something
    return_type = None
```
so this should just be eliminating the extra exception in the loop, and IME throwing exceptions for cases where in most contexts you can safely ignore the error isn't a good idea anyway.

Hidden here though is also a logical error in our handling of return types, where we assume `None` means "we couldn't find a return type" which precludes the possibility of correctly parsing a return type of `None`, but I wasn't gonna handle that now

PR #122, opened 2023-05-17
Full oneof support

gabe-l-hart:
## Description

This PR adds full support for `oneof` semantics to the data model. Specifically, it makes the following true:

1. Any class derived from `DataBase` (including those derived from `DataObjectBase`) will internally store the current value of a oneof under the name of the oneof and not the name of the currently set field within the oneof
2. The `__init__` to any class derived from `DataBase` can take a keyword argument with the name of the `oneof` OR the name of one of the fields, but not both (and not multiple fields)
3. When calling `getattr` on a data model instance with the name of a `oneof`, it will return whatever value is set if any is
4. When calling `getattr` on a data model instance with the name of a `oneof` field, it will return the value of the oneof if (and only if) that field is the one set in the oneof
5. Every object has a `which_oneof` function which can be used to determine which field in a oneof is currently set (if any)
6. When the value is set using the name of the `oneof` rather than the name of a field, the `which_oneof` information is lazily introspected from the type of the value the first time `which_oneof` is called (this includes attribute access by field name)
    * **NOTE**: This may cause issues with `oneofs` where the fields are not disambiguated by type! In this case, the first matching field is chosen


caikit/core/data_model/data_backends/dict_backend.py
```diff
@@ -63,7 +63,10 @@ def get_attribute(self, data_model_class: Type[DataBase], name: str) -> Any:
         #   being used correctly.
 
         # Make sure the name is a valid field on the given class
-        if name not in data_model_class.fields:
+        if (
```
gabe-l-hart, line 66: In general, anywhere that we assume a field needed to be in `cls.fields` it's now valid to check `cls._fields_oneofs_map` to allow the names of oneofs through.

caikit/core/data_model/dataobject.py
```diff
@@ -67,16 +67,11 @@ class will be the raw input representation of a @dataclass
...
         if name != "DataObjectBase":
-            raw_field_names = attrs.get("__annotations__")
-            if raw_field_names is None:
+            field_names = attrs.get("__annotations__")
+            if field_names is None:
                 raise TypeError(
                     "All DataObjectBase classes must follow dataclass syntax"
                 )
-
-            # TODO: Sort out oneof field names
```
gabe-l-hart, line 76: This ended up being accomplished through guaranteeing that all `DataBase` derived classes stored oneof data by the name of the oneof rather than the name of the field, so nothing more was needed here.

caikit/core/data_model/dataobject.py
```diff
@@ -129,47 +124,27 @@ def decorator(cls: Type) -> Type[DataBase]:
         # Add the package to the kwargs
         kwargs.setdefault("package", package)
 
-        # If there's a schema in the keyword args, use jtd_to_proto
-        schema = kwargs.pop("schema", None)
```
gabe-l-hart, line 133: `schema` is gone!!!!!

caikit/runtime/service_generation/data_stream_source.py
```diff
@@ -216,65 +204,92 @@ def make_data_stream_source(data_element_type: Type) -> Type[DataBase]:
     if data_element_type not in _DATA_STREAM_SOURCE_TYPES:
         cls_name = _make_data_stream_source_type_name(data_element_type)
         log.debug("Creating DataStreamSource[%s] -> %s", data_element_type, cls_name)
-        element_type = (
-            data_element_type.get_proto_class().DESCRIPTOR
-            if isinstance(data_element_type, type)
-            and issubclass(data_element_type, DataBase)
-            else _NATIVE_TYPE_TO_JTD[data_element_type]
+
+        # Set up the "sub classes." In python, this is the same as creating a
```
gabe-l-hart, line 208: I can't say that this is really _prettier_ than the nasty `schema` stuff, but it's at least consistent now, so this change enabled fully removing the JTD-style of `@dataobject`.

caikit/runtime/service_generation/data_stream_source.py
```diff
@@ -288,10 +303,15 @@ def make_data_stream_source(data_element_type: Type) -> Type[DataBase]:
 
         # Add an init that sequences the initialization so that
         # DataStreamSourceBase is initialized after DataBase
-        orig_init = _DataObjectBaseMetaClass._make_init(data_object.fields)
+        orig_init = _make_oneof_init(data_object)
```
gabe-l-hart, line 306: We always know this will have a `oneof`, so just use the oneof init directly

caikit/runtime/service_generation/data_stream_source.py
```diff
@@ -288,10 +303,15 @@ def make_data_stream_source(data_element_type: Type) -> Type[DataBase]:
...
 
         def __init__(self, *args, **kwargs):
-            orig_init(self, *args, **kwargs)
+            try:
+                orig_init(self, *args, **kwargs)
+            except TypeError as err:
+                raise CaikitRuntimeException(
```
gabe-l-hart, line 312: Since the `oneof_init` now does duplicate arg checking, the `CaikitRuntimeException` needed to be hoisted to here

tests/core/data_model/test_base.py
```diff
@@ -161,6 +164,86 @@ def __init__(self, foo):
...
+    # pylint: disable=duplicate-code
+    with temp_data_model(
+        make_proto_def(
+            {
+                "ThingOne": {
+                    "foo": str,
+                    "bar": int,
+                }
+            },
+            mock_compiled=True,
```
gabe-l-hart, line 180: This makes the test simulate using a pre-compiled protobuf that uses `DataBase` directly without `@dataobject`.

tests/core/data_model/test_dataobject.py
```diff
@@ -113,22 +113,6 @@ class Foo(DataObjectBase):
     assert inst.bar is None
 
 
-def test_dataobject_jtd():
```
gabe-l-hart, line 116: Bye bye `schema`!

tests/core/data_model/test_dataobject.py
```diff
@@ -454,47 +365,68 @@ class Bar(DataObjectBase):
             Annotated[Bar, FieldNumber(2), OneofField("bar")],
         ]
 
-        def __getattr__(self, name):
```
gabe-l-hart, line 457: All of this custom stuff that needed to be added to support `oneofs` is now implicit

tests/data_model_helpers.py
```diff
@@ -147,24 +147,61 @@ def _get_proto_val_name(field_val) -> str:
     raise RuntimeError(f"Invalid field type specifier: {field_val}")
 
 
-def make_proto_def(message_specs: Dict[str, dict], pkg_suffix: str = None) -> str:
+def make_proto_def(
```
gabe-l-hart, line 150: Changes in this file support creating temporary data model modules that simulate using precompiled protobufs rather than using `@dataobject`

caikit/core/data_model/base.py
```diff
@@ -70,29 +76,43 @@ def __new__(mcs, name, bases, attrs):
         protobufs, it can be named in the tuple class variable _private_slots and will
         automatically be added to __slots__.
         """
-        # Get all fields in protobufs with same name as class,
-        # except for DataBase, which has no matching protobufs
-        fields = ()
 
-        # Protobufs fields can be divided into these categories, which are used to automatically
```
gabe-l-hart, line 77: I rearranged things a bit so that it was less confusing to keep track of the pre-defined empty variables and the names they were assigned to in `attrs`

caikit/core/data_model/base.py
```diff
@@ -70,29 +76,43 @@ def __new__(mcs, name, bases, attrs):
...
-        _fields_enum_repeated = ()
-        _fields_primitive = ()
-        _fields_primitive_repeated = ()
-        _fields_map = ()
+        # Protobufs fields can be divided into these categories, which are used
+        # to automatically determine appropriate behavior in a number of methods
+        attrs["full_name"] = name
+        attrs["fields_enum_map"] = {}
+        attrs["fields_enum_rev"] = {}
+        attrs["_fields_oneofs_map"] = {}
```
gabe-l-hart, line 85: Two new detail attrs here:

* `_fields_oneofs_map`: Maps from `oneof` name to the set of fields in that oneof
* `_fields_to_oneof`: Inverse of `_fields_oneofs_map`

caikit/core/data_model/base.py
```diff
@@ -70,29 +76,43 @@ def __new__(mcs, name, bases, attrs):
...
             # descriptor
             proto_class = attrs.get("_proto_class")
             if proto_class is not None:
-                fields = tuple(proto_class.DESCRIPTOR.fields_by_name)
+                all_oneof_fields = [
+                    field.name
+                    for oneof in proto_class.DESCRIPTOR.oneofs
+                    for field in oneof.fields
+                ]
+                fields = tuple(
```
gabe-l-hart, line 109: This logic here is responsible for ensuring that when using precompiled protos, data is stored using the oneof name and not the individual field name. This is needed so that the right `__slots__` are defined!

caikit/core/data_model/base.py
```diff
@@ -171,6 +181,21 @@ def parse_proto_descriptor(mcs, cls):
             if field.enum_type is not None
         }
 
+        # mapping of all oneofs and the fields that are part of them
+        # NOTE: protobuf makes an interesting use of oneof to wrap types that
```
gabe-l-hart, line 185: This was important. For _all_ of this oneof logic, I'm excluding oneofs that are used only to wrap `optional` fields.

caikit/core/data_model/base.py
```diff
@@ -236,28 +263,36 @@ def parse_proto_descriptor(mcs, cls):
...
+        # existed in old_fields for supporting oneofs
         # see https://github.com/caikit/caikit/pull/107 for details
         for field in set(cls.fields + tuple(old_fields)):
-            setattr(cls, field, mcs._make_property_getter(field))
+
+            # If the field is the name of a field within a oneof and it was not
+            # in the old fields, the data is held under the oneof's name if this
+            # is the set value for the oneof
+            if oneof_name := cls._fields_to_oneof.get(field):
+                setattr(cls, field, mcs._make_property_getter(field, oneof_name))
```
gabe-l-hart, line 275: This extra argument results in a special getter that will first check `which_oneof` before returning the value of the attribute held under the name of the oneof

caikit/core/data_model/base.py
```diff
@@ -344,10 +415,22 @@ class DataBase(metaclass=_DataBaseMetaClass):
     """
 
     def __setattr__(self, name, val):
-        """If attempting to set one of the named fields, instead set the
-        "private" version of the attribute.
+        """Handle attribute setting for oneofs and named fields with delegation
+        to backends as needed
         """
-        if name in self.__class__.fields:
+        # If setting a oneof directly, remove any oneof information
```
gabe-l-hart, line 421: This is needed so that the next time `which_oneof` is called, it will redo the lazy introspection by type

tests/core/data_model/test_dataobject.py
```diff
@@ -454,47 +365,68 @@ class Bar(DataObjectBase):
...
+    assert foo2.bar is foo2.data_stream
+    assert foo2.which_oneof("data_stream") == "bar"
+
+    # Construct with positional oneof name
+    foo2 = BazObj(BazObj.Foo(data=["some", "foo"]))
+    assert foo2.data_stream.data == ["some", "foo"]
+    assert foo2.bar is None
+    assert foo2.foo is foo2.data_stream
+    assert foo2.which_oneof("data_stream") == "foo"
+
```
prashantgupta24, line 422: This is amazing! The amount of flexibility to define a oneof is 🌶️

prashantgupta24: Is it worth adding an empty initialization test case here?

prashantgupta24:
```suggestion
    foo3 = BazObj()
    assert foo3.foo is None
    assert foo3.bar is None
    assert foo3.data_stream is None
    assert foo3.which_oneof("data_stream") == None
```

caikit/core/data_model/base.py
```diff
@@ -171,6 +181,21 @@ def parse_proto_descriptor(mcs, cls):
...
         }
 
+        # mapping of all oneofs and the fields that are part of them
+        # NOTE: protobuf makes an interesting use of oneof to wrap types that
+        #   should be explicitly optional. We don't want to consider these
+        #   oneofs in the general oneof handling.
+        cls._fields_oneofs_map = {
+            oneof_name: [field.name for field in oneof.fields]
+            for oneof_name, oneof in cls._proto_class.DESCRIPTOR.oneofs_by_name.items()
+            if len(oneof.fields) != 1 or oneof.name != f"_{oneof.fields[0].name}"
```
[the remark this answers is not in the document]
gabe-l-hart: Yep, this is just trying to identify the entries in `oneofs` that are only there for `optional` fields and removing them from the list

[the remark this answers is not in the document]
prashantgupta24: To make sure I understood this correctly - 
```
message Foo {
    int32 bar = 1;
    optional int32 baz = 2;
}
``` 
becomes
```
message Foo {
    int32 bar = 1;
    oneof baz {
        int32 baz = 2;
    }
}
```
because then `oneof.name` is `baz`, which is the same as `oneof.fields[0].name`?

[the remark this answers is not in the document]
gabe-l-hart: This is close, but the name of the oneof is `_baz` not `baz`

tests/core/data_model/test_dataobject.py
```diff
@@ -454,47 +365,73 @@ class Bar(DataObjectBase):
...
+    assert foo3.bar is None
+    assert foo3.data_stream is None
+    assert foo3.which_oneof("data_stream") == None
+    # Invalid constructors
+    with pytest.raises(TypeError):
+        BazObj(BazObj.Foo(), foo=BazObj.Foo())
+    with pytest.raises(TypeError):
+        BazObj(data_stream=BazObj.Foo(), foo=BazObj.Foo())
+    with pytest.raises(TypeError):
+        BazObj(foo=BazObj.Foo(), bar=BazObj.Bar())
```
prashantgupta24, line 436:
```suggestion
def test_dataobject_with_same_type_of_oneof():
    """Make sure that using a Union to create a oneof with the same types works as expected"""

    @dataobject
    class Foo(DataObjectBase):
        foo: Union[
            Annotated[bool, FieldNumber(10), OneofField("foo_bool1")],
            Annotated[bool, FieldNumber(20), OneofField("foo_bool2")],
        ]

    # if the fields are of the same type, then by default the first one is set
    foo1 = Foo(True)
    assert foo1.which_oneof("foo") == "foo_bool1"
    assert foo1.foo_bool1
    assert foo1.foo_bool2 == None

    # unless set explicitly
    foo2 = Foo(foo_bool2=True)
    assert foo2.which_oneof("foo") == "foo_bool2"
    assert foo2.foo_bool1 == None
    assert foo2.foo_bool2
```

tests/core/data_model/test_dataobject.py
```diff
@@ -454,47 +365,73 @@ class Bar(DataObjectBase):
...
```
prashantgupta24, line 435: Worth adding another test with oneofs of the same type? not sure if that's a common scenario

prashantgupta24 (approved):
Looks great from what I understand! Added an optional `same_type_of_oneof` test we can include

tests/core/data_model/test_dataobject.py
gabe-l-hart: Yep! Thanks for adding this

PR #123, opened 2023-05-17
✨ support unions in function signatures

prashantgupta24:
This PR enables `Union` type arguments in `run` and `train` functions.

## Changes
- Removes the `extract_primitive_type_from_union`, instead use `is_primitive_type` and not worry about extraction anymore since we support `Union` now.
- If generating protobuf for field with `Union`, make sure to increment field numbers by the number of args within the `Union`
- Handle `Union` types in `build_caikit_library_request_dict`
    - Make sure to set the name of the field to the name of the `oneof`
    - Make sure to set the value accordingly as well
- Change tests to use `oneof` fields in places where arg type is a `Union`. 

## Example:
Sample `train` function:
```
def train(self, input: Union[bool, str], val2: str):
```
Proto file generated:
```
 string model_name = 1;
 string val2 = 4;

  /*-- oneofs --*/
  oneof input {
    bool inputbool = 2;
    string inputstr = 3;
  }
}
```

Review comments:
- [x] Address `Union[supported_type, unsupported_type]` -> flatten it to the supported type


caikit/runtime/utils/servicer_util.py
```diff
@@ -233,18 +233,39 @@ def build_caikit_library_request_dict(
...
             [field.name for field in request.DESCRIPTOR.fields],
         )
         for field in request.DESCRIPTOR.fields:
-
-            log.debug2("processing field: %s", field.name)
+            field_name = field.name
+            field_value = None
+            log.debug2("processing field: %s", field_name)
+            # check for oneofs
+            if field.containing_oneof:
```
prashantgupta24, line 247: need to revisit this block of many ifs to make it better

caikit/runtime/service_generation/primitives.py
```diff
@@ -45,52 +45,12 @@ def to_primitive_signature(
     primitives = {}
     log.debug("Building primitive signature for %s", signature)
     for arg, arg_type in signature.items():
-        primitive_arg_type = extract_primitive_type_from_union(
-            primitive_data_model_types, arg_type
-        )
-        if primitive_arg_type:
-            primitives[arg] = primitive_arg_type
+        if is_primitive_type(arg_type, primitive_data_model_types):
```
prashantgupta24, line 48: Not sure if the function name `to_primitive_signature` holds true anymore - but don't want to change a lot since other PRs are touching the same stuff too

caikit/runtime/service_generation/rpcs.py
```diff
@@ -322,14 +321,17 @@ def __init__(
...
-        for _, (item_name, p) in enumerate(params.items()):
+        for _, (item_name, typ) in enumerate(params.items()):
             if item_name in existing_fields:
                 # if field existed previously, get the original number from there
                 num = existing_fields[item_name]
             else:
                 num = last_used_number + 1
-                last_used_number += 1
-            self.triples.append((p, item_name, num))
+                if get_origin(typ) == Union:
```
gabe-l-hart, line 330: NIT: I think this could be `is` instead of `==`

caikit/runtime/utils/servicer_util.py
```diff
@@ -238,13 +238,43 @@ def build_caikit_library_request_dict(
             "We are looping though these fields: %s",
             [field.name for field in request.DESCRIPTOR.fields],
         )
+        # get all oneof fields excluding from optional oneofs
+        one_of_fields = []
+        for oneof_name, oneof in request.DESCRIPTOR.oneofs_by_name.items():
```
gabe-l-hart, line 243: I'm probably missing something here, but I would expect that `request` here is a `DataObjectBase` and not a `proto`? The code says I'm wrong, but if I weren't, this logic to find non-optional oneofs would already be handled there and wouldn't need to be duplicated.

gabe-l-hart (commented):
Thanks for adding this @prashantgupta24! If I'm reading this correctly, this will _always_ translate `Union[DataModelType, equivalent_primitive]` into a `oneof`, right? Presumably, it would also always fail if it were `Union[supported_type, unsupported_type]`?

I'm wondering if we somehow want more nuanced semantics around these cases. Specifically, if the `Union` is just python sugar for saying "you can use the native python type for this" (e.g. `RawDocument`/`str`), do we want to have the interface require a `oneof`, or just pick one?

In the case of `Union[supported_type, unsupported_type]`, I think we'd want to flatten down that to the single `supported_type` and not have a `oneof`.

caikit/runtime/utils/servicer_util.py
prashantgupta24: From what I understand `request` is a proto since it's the argument that's directly passed from the `def Train(self, request, *_, **__)` within `caikit/runtime/servicers/global_train_servicer.py`.

gabe-l-hart: Ah, that makes sense

gabe-l-hart:
>Not sure about RawDocument / str -> so we want to figure out if RawDocument internally contains a str and hence only pick one? What if it was RawDocument / int? In that case, we would want a oneof right?

Yeah, this one is much less straightforward. I think the most generic you could get would be some way of configuring sets of types that all map to the same data model object. This is probably less important than the other case, though.

tests/runtime/test_grpc_server.py
```diff
@@ -280,7 +283,7 @@ def test_train_fake_module_does_not_change_another_instance_model_of_block(
 
     # make sure the trained model can run inference, and the batch size 100 was used
     predict_request = sample_inference_service.messages.OtherTaskRequest(
-        sample_input=HAPPY_PATH_INPUT
+        sample_inputsampleinputtype=HAPPY_PATH_INPUT
```
joerunde, line 286: 🤢🤢🤢 oof this name...

I think I'd expect a `_` between the oneof name and the type name, should we change that?

caikit/runtime/utils/servicer_util.py
```diff
@@ -254,45 +284,45 @@ def build_caikit_library_request_dict(
...
+                        field_name,
                         e,
                     )
                     # iterables
                     if isinstance(field_value, Iterable):
                         if len(field_value) != 0:
                             # cast only if it's actually a list
                             if not isinstance(field_value, (str, bytes)):
-                                if "training_data" in field.name:
+                                if "training_data" in field_name:
```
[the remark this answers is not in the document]
joerunde: For real though, I imagine we should be handling all data stream types in the same way?
And that should probably involve resolving any `DataStreamSourceX` types into lazily evaluated `DataStream`s?

(For future work though, for sure)

tests/runtime/service_generation/test_primitives.py
```diff
@@ -38,9 +37,41 @@ def test_to_primitive_signature_union_dm():
...
     ) == {"name": SampleInputType}
 
 
+def test_to_primitive_signature_multiple_no_dm_primitives_in_union():
+    """We have 1 dm arg that's not supported, and 2 primitive args. We
+    return the first primitive arg"""
+    assert to_primitive_signature(
+        signature={"name": Union[SampleInputType, str, int]},
+        primitive_data_model_types=[],
+    ) == {"name": str}
```
joerunde, line 72: 🤔 These last two test cases are kinda making me question this whole design with "primitiveness" of data models.

Given the move to `@task`s to specify the minimum set of required args, I bet this whole "is a DM a primitive of the domain" idea goes away entirely, as we'll probably flip the logic of
- `include an api if all inputs are domain primitive` to 
- `include an api if it contains the required parameters for this task`

And I also bet it's more important to consider type unions with non-protobuffable things, like 
```
def run(foo: MyCoolDataModel | pandas.DataFrame | spark.DataFrame)
```

Given that, I might lean more towards the behavior of `handle_primitives_in_union` to be a simple removal of any types in the union that cannot be serialized to protobuf. 
Thoughts?

tests/runtime/test_grpc_server.py
prashantgupta24: Yeah I would honestly like that too, but that's set in `py-to-proto` right?

gabe-l-hart: Feel free to submit a PR! https://github.com/IBM/py-to-proto/blob/main/py_to_proto/dataclass_to_proto.py#L257

tests/runtime/service_generation/test_primitives.py
joerunde: This might be directly in opposition to @gabe-l-hart's thoughts on `Union[supported_type, unsupported_type]`, though I think I'm redefining `unsupported_type` simply as anything that is not serializable

tests/runtime/test_grpc_server.py
prashantgupta24: Ookay, so this is a breaking change and will be fixed by a major bump for py-to-proto then!  Let's keep it for a later release ?

joerunde: Sure!

joerunde (approved):
looks 🌶️

prashantgupta24:
Thanks for the review! Will merge this in, I think a lot of discussion points will be handled by this issue: https://github.com/caikit/caikit/issues/136

PR #124, opened 2023-05-18
:sparkles: add log config to dump services cli

joerunde:
It'd be real nice to be able to set LOG_LEVEL when dumping services


evaline-ju (approved):
Definitely would be useful for debugging!

PR #126, opened 2023-05-18
Numpy type support

gabe-l-hart:
## Description

This PR adds `numpy` as a dependency and allows common numpy `dtypes` to be used as type hints for data objects

## Discussion

While this is a small change, it does bring in a new dependency which should not be taken lightly! The rule in `caikit` has always been that it does not contain any "algorithmic" dependencies and instead downstream libraries should use the algorithmic libraries that best fit their needs. In the case of `numpy`, there are strong arguments to be made for flexing this rule:

* Almost _every_ AI library will depend on `numpy` either directly or indirectly
* `numpy` is only _moderately_ "heavyweight" as compared to libraries like `torch`, `tensorflow`, etc, so it is a much less onerous unnecessary dependency in the case where it's not actually needed


pyproject.toml
```diff
@@ -21,17 +21,18 @@ dependencies = [
     "grpcio-health-checking>=1.35.0,<2.0",
     "grpcio>=1.35.0,<2.0",
     "ijson>=3.1.4,<3.3.0",
+    "import-tracker>=3.1.5,<4",
     "munch>=2.5.0,<4.0",
-    "protobuf>=3.19.0,<5",
```
gabe-l-hart, line 25: The changes here are because I sorted the list which I think will help avoid conflicts in the future

pyproject.toml
```diff
@@ -21,17 +21,18 @@ dependencies = [
...
+    "numpy>=1.23.5,<2",
```
alex-jw-brooks, line 26: Any reason for 1.23.5 as the lower bound, specifically?

alex-jw-brooks (commented):
Nice, I think this looks great! Quick question about the lower bound for the numpy dependency, but otherwise I think I'm happy with this.

alex-jw-brooks (approved):
Looks great! LGTM

pyproject.toml
alex-jw-brooks: Changed to 1.20 after discussing internally

pyproject.toml
gabe-l-hart: Good call. I picked that one somewhat at random. It looks like `1.20` is the lowest I can reasonably install, so I'll pin it there

PR #129, opened 2023-05-22
🐛  Load backend arg

joerunde:
This PR is a little bugfix to supply the concrete load backend as a kwarg to `.load` classmethods on alternative backend module implementations.

This is to support users who need to have access to the load backend at initialization time. Previously they used `caikit.get_backend(type)` but this was removed in favor of explicitly setting the load backend. For shared load backends this is fine, but for backend-specific module implementations this deferred knowledge of the backend until post initialization. 

We have some design work to do to fix this gap, but in the meantime a quick extra kwarg to load will unblock us.

### DIFF
--- a/caikit/core/model_manager.py
+++ b/caikit/core/model_manager.py
@@ -259,6 +259,7 @@ class ModelManager:
                             load_backend.backend_type,
                             module_backend_impl.__name__,
                         )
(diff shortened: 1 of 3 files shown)


evaline-ju (approved):
LGTM

PR #130, opened 2023-05-22
:bug: allow ModuleSaver/Loader import from core

joerunde:
whoopsies

### DIFF
--- a/caikit/core/__init__.py
+++ b/caikit/core/__init__.py
@@ -34,7 +34,7 @@ from .data_model import DataObjectBase, dataobject
 from .model_manager import *
 from .module_backends import *
 from .module_backends.module_backend_config import configure as backend_configure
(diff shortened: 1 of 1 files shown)


gabe-l-hart (approved):
@joerunde Needs a `fmt`, but good otherwise

caikit/core/__init__.py
```diff
@@ -34,7 +34,7 @@
 from .model_manager import *
 from .module_backends import *
 from .module_backends.module_backend_config import configure as backend_configure
-from .modules import ModuleBase, ModuleConfig, module
+from .modules import ModuleBase, ModuleConfig, module, ModuleSaver, ModuleLoader
```
gabe-l-hart, line 37:
```suggestion
from .modules import ModuleBase, ModuleConfig, ModuleLoader, ModuleSaver, module
```

PR #132, opened 2023-05-23
🐛 handle from_proto oneof case

prashantgupta24:
## Issue

The problem is the fact that when we have
```
    @dataobject
    class BazObj(DataObjectBase):
        @dataobject
        class Foo(DataObjectBase):
            data: List[str]

        @dataobject
        class Bar(DataObjectBase):
            data: str

        data_stream: Union[
            Annotated[Foo, FieldNumber(1), OneofField("foo")],
            Annotated[Bar, FieldNumber(2), OneofField("bar")],
        ]
```

During the `from_proto` -> since `foo` and `bar` are part of `cls._fields_message`, we do the `if proto.HasField(field): `check before setting the value of the field. So in this case, we don’t set the value of `bar`.

But, in the case of
```
    @dataobject
    class Foo(DataObjectBase):
        foo: Union[
            Annotated[int, FieldNumber(10), OneofField("foo_int")],
            Annotated[float, FieldNumber(20), OneofField("foo_float")],
        ]
```
since both `foo_int` and `foo_float` are in `cls._fields_primitive`, we don’t check for `hasField` and instead assign the field value directly, and so we set both `foo_int` and `foo_float`.
This crashes the `_make_oneof_init` within `dataobject` which wants only 1 field within a `oneof` to be set at a time.

## Possible solution
I figured the only way to correct this behavior is to tackle it at the `proto` level - because once we are beyond the `proto` level, there’s no way for us to figure out an unset field vs false value of the field.

### DIFF
--- a/caikit/core/data_model/base.py
+++ b/caikit/core/data_model/base.py
@@ -597,8 +597,12 @@ class DataBase(metaclass=_DataBaseMetaClass):
                 )
 
             if field in cls._fields_primitive or field in cls._fields_enum:
-                kwargs[field] = proto_attr
-
+                # special case for oneofs
+                if field in cls._fields_to_oneof:
+                    if proto.HasField(field):
+                        kwargs[field] = proto_attr
+                else:
+                    kwargs[field] = proto_attr
             elif (
                 field in cls._fields_primitive_repeated
                 or field in cls._fields_enum_repeated
--- a/caikit/core/data_model/base.py
+++ b/caikit/core/data_model/base.py
@@ -598,10 +598,7 @@ class DataBase(metaclass=_DataBaseMetaClass):
 
             if field in cls._fields_primitive or field in cls._fields_enum:
                 # special case for oneofs
-                if field in cls._fields_to_oneof:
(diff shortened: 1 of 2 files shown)


caikit/core/data_model/base.py
```diff
@@ -597,8 +597,12 @@ def from_proto(cls, proto):
...
 
             if field in cls._fields_primitive or field in cls._fields_enum:
-                kwargs[field] = proto_attr
-
+                # special case for oneofs
+                if field in cls._fields_to_oneof:
+                    if proto.HasField(field):
+                        kwargs[field] = proto_attr
+                else:
+                    kwargs[field] = proto_attr
```
gabe-l-hart, line 605:
```suggestion
                if field not in cls._fields_to_oneof or proto.HasField(field):
                    kwargs[field] = proto_attr
```

gabe-l-hart (changes requested):
One suggested simplification, but otherwise I think this looks good

gabe-l-hart (approved):
Ship it!

PR #147, opened 2023-05-25
Small update: Tox file change

abhishek-maurya-ibm:
Changed the environment setup line in tox.ini file such that it dynamically detects python version of the user and uses that version as tox python environment. Also updated command in CONTRIBUTING.md

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,7 +1,4 @@
 [tox]
-`; envlist =
-;     3.{11,10,9,8}
-; envlist = py311, lint
 envlist = py, lint, fmt
 
 [testenv]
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -1,6 +1,8 @@
 [tox]
-envlist =
-    3.{11,10,9,8}
+`; envlist =
+;     3.{11,10,9,8}
+; envlist = py311, lint
```
joerunde, line 4: Nice! Can we delete these old commented lines too?

joerunde (approved):
Looks good, thanks for the improvements!

PR #149, opened 2023-05-25
:sparkles: add potential data stream fix

joerunde:
This is an example of how we could change datastreams to be pickle-able.
The reason they are not is purely due to the way the current classmethod initializers create local closures + functions around the arguments passed in. These local closures + functions are not pickleable, pickle needs good concrete function pointers to be able to import again.

Instead, we can use DataStream's already existing functionality to store `*args` and `**kwargs` to pass to an arbitrary genrator-provider-function to reference a well-known generator producing function without requiring a closure to capture the args the user provided.

See the provided example for `.from_jsonl()`, we'll need to do something like this for all of the `.from_*` functions.

### DIFF
--- a/caikit/core/data_model/streams/data_stream.py
+++ b/caikit/core/data_model/streams/data_stream.py
@@ -168,28 +168,28 @@ class DataStream(Generic[T]):
         """
         error.file_check("<COR32600575E>", filename)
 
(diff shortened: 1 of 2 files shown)


evaline-ju:
Closable with #192 ?

PR #154, opened 2023-05-29
Changes for hosting API docs on readthedocs

### LINKED ISSUE #69: Caikit API Docs
## Description

Caikit python references on readthedocs.org which describe Caikit API.

## Discussion

We've used sphinx in the past with some success. We had been heavily relying on the `autodoc` features to walk the import tree and document everything, but the resulting docs were a bit scattered and sometimes missing.

It might be worth explicitly annotating which apis should be documented, and laying that out in a docs tree for sphinx to generate instead.

## Acceptance Criteria

- [x] Add mechanism to ganerate API docs (#125)
- [ ] caikit refs on [readthedocs.org](https://docs.readthedocs.io) (#154)

hickeyma:
[readthedocs](https://docs.readthedocs.io/en/stable/index.html) requires the docs to be generated from source code prior to readthedocs build as it runs
[`sphinx-build` only and not `sphinx-apidoc` as well](https://github.com/readthedocs/readthedocs.org/issues/1139).

This PR adds [autoapi extension](https://github.com/readthedocs/sphinx-autoapi) to the sphinx config to run the automatic generation of docs from the source files during `sphinx-build` . It also tidies the sphinx files into the `source` directory and removes the redundant call to `sphinx-apidoc`.

Close #69 

## Notes for reviewers:

- For now, will go with latest code version
- Once this PR is merged, the full docs will be hosted at https://caikit.readthedocs.io/en/latest/


docs-requirements.txt
```diff
@@ -1,3 +1,4 @@
 # For API doc generation
 sphinx>=4.0.2,<8.0
+sphinx-autoapi
```
gabe-l-hart, line 3: Worth putting a version here?

docs/source/conf.py
```diff
@@ -14,7 +14,7 @@
 import os
 import sys
 
-sys.path.insert(0, os.path.abspath("../"))
+sys.path.insert(0, os.path.abspath("../../"))
```
gabe-l-hart, line 17: NIT: This is not platform independent as written. Probably not going to be running this on Windows, but `os.path.join("..", "..")` would be cleaner

docs/source/conf.py
```diff
@@ -30,7 +30,9 @@
 # extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
 # ones.
 extensions = [
-    "sphinx.ext.autodoc",
+    # Generate complete API docs by parsing source code
+    "autoapi.extension",
+    # Add links to source from generated docs
```
gabe-l-hart, line 35: Thanks for adding these comments! It will make it much easier to reuse this template elsewhere

docs/source/conf.py
```diff
@@ -42,6 +44,14 @@
 # This pattern also affects html_static_path and html_extra_path.
 exclude_patterns = ["_build", "Thumbs.db", ".DS_Store", "caikit/config"]
 
+# -- autoapi configuration ---------------------------------------------------
+
+# Language of source code to parse
+autoapi_type = "python"
+
+# Source code to parse to generate API docs relative to 'docs' directory
+autoapi_dirs = ["../../caikit"]
```
gabe-l-hart, line 53: Same NIT about `os.path.join` (not important, just pointing it out!)

gabe-l-hart (approved):
Really excited to have this! A couple of little NITs, but it looks ready to go any time

docs-requirements.txt
hickeyma: Yes, added

docs/source/conf.py
hickeyma: Probably Linux only but a good point all the same. Updated to be platform agnostic.

Issue #146, opened 2023-05-25
Surface coverage report to GitHub, gate on coverage level


afrittoli:
## Is your feature request related to a problem? Please describe.

This is related to issue #15 - to caikit CI infrastructure rather than caikit itself.

## Describe the solution you'd like

to keep the coverage level above the desired threshold it would be helpful to have the coverage data published as a comment on PRs and possibly enforce a minimum level of coverage required.

This helps PR reviewers to catch PRs that would reduce coverage before they are merged.

## Describe alternatives you've considered

There are github apps in the marketplace that could provide this feature, like for instance [python-coverage](https://github.com/marketplace/actions/python-coverage). I'm not recommending this one specifically, it would need to be verified.

afrittoli:
I would be happy to help with this.

PR #166, opened 2023-05-31
Add coverage report and threshold

afrittoli:
Add coverage report thanks to the python-coverage GitHub action https://github.com/marketplace/actions/python-coverage

The action reports the coverage as a GitHub comment. It also allows defining thresholds for the level of coverage. For now, setting an 80% thresholds for new files and zero for existing content. Once the report is published to PRs we can start working on bringing the coverage up to the desired level and enable the threshold for existing code.

We run unit tests for four different versions of python, but I don't think we need coverage for each.
Nonetheless, I changed the tox.ini to make sure that the coverage reports from the four runs do not override each other, and produce both html and xml reports.
Html is convenient when running locally, xml is needed by the action.

The action uses the coverage report for python 3.11

Fixes: #146

### DIFF
--- a/.github/workflows/build-library.yml
+++ b/.github/workflows/build-library.yml
@@ -23,6 +23,10 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      id-token: write
+      pull-requests: write
     strategy:
       matrix:
         python-version:
@@ -50,4 +54,14 @@ jobs:
           python -m pip install -r setup_requirements.txt
       - name: Build and test with tox
         run: tox -e ${{ matrix.python-version.tox }}
+      - name: Publish coverage 
+        if: ${{ matrix.python-version.setup == '3.11' }}
+        uses: orgoro/coverage@v3.1
+        with:
+            coverageFile: coverage-py311.xml
+            token: ${{ secrets.GITHUB_TOKEN }}
+            thresholdAll: 0.0
+            thresholdNew: 0.8
+            thresholdModified: 0.0
+
 
--- a/tox.ini
+++ b/tox.ini
@@ -15,7 +15,7 @@ passenv =
     LOG_FORMATTER
     LOG_THREAD_ID
     LOG_CHANNEL_WIDTH
-commands = pytest --cov=caikit --cov-report=html {posargs:tests}
+commands = pytest --cov=caikit --cov-report=html:coverage-{env_name} --cov-report=xml:coverage-{env_name}.xml {posargs:tests}
 
 ; Unclear: We probably want to test wheel packaging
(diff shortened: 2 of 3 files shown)


.github/workflows/build-library.yml
```diff
@@ -47,4 +47,13 @@ jobs:
           python -m pip install -r setup_requirements.txt
       - name: Build and test with tox
         run: tox -e ${{ matrix.python-version.tox }}
+      - name: Publish coverage 
+        uses: orgoro/coverage@v3.1
+        with:
+            coverageFile: coverage-py3.11.xml
```
joerunde, line 53: I ran this locally with `tox -e py311` which I think this matrix will do, and it looks like this file is named `coverage-py311.xml` (without the `.` in `3.11`)

tox.ini
```diff
@@ -15,7 +15,7 @@ passenv =
     LOG_FORMATTER
     LOG_THREAD_ID
     LOG_CHANNEL_WIDTH
-commands = pytest --cov=caikit --cov-report=html {posargs:tests}
+commands = pytest --cov=caikit --cov-report=html:coverage-{env_name} --cov-report=xml:coverage-{env_name}.xml {posargs:tests}
```
joerunde, line 18: Nice!

Can we get a .gitignore entry for the coverage xml files?

.github/workflows/build-library.yml
joerunde: Ah, yeah the action fails with this error too:
```
Error: "could not read file coverage-py3.11.xml"
```

tox.ini
afrittoli: Done

.github/workflows/build-library.yml
afrittoli: Thanks, nice catch - fixed that now. 
I was invoking locally `tox -e py3.11` which also works but it's not what's in the github action.

afrittoli:
@joerunde I tested this set of permissions on my fork and it worked fine:

```yaml
    permissions:
      contents: read
      pull-requests: write
```

I think the only bit missing might be to remove the `GITHUB_TOKEN` from the secrets, so that we get the automatic one with the scope specified in the workflow definition.

joerunde:
@afrittoli something's still off then, there is no extra `GITHUB_TOKEN` in the secrets, github won't let you add any secret that starts with `GITHUB_` :(

joerunde:
@afrittoli I found that I needed to go to the org-wide settings to check the "Read and write permissions" button for actions, but this is still failing after I re-ran the job :(

hickeyma (approved):
@afrittoli Looks like the changes to the permissions that we discussed now work. Thanks for asking the coverage @afrittoli

PR #169, opened 2023-06-01
Proto three fix 168

### LINKED ISSUE #168: Fix support for protobuf 3.x
## Describe the bug

```py
>>> import caikit
  File "<stdin>", line 1, in <module>
  File "/Users/ghart/Projects/github/caikit/caikit/caikit/__init__.py", line 7, in <module>
    from . import core, interfaces
  File "/Users/ghart/Projects/github/caikit/caikit/caikit/core/__init__.py", line 33, in <module>
    from .data_model import DataObjectBase, dataobject
  File "/Users/ghart/Projects/github/caikit/caikit/caikit/core/data_model/__init__.py", line 20, in <module>
    from . import base, data_backends, enums, producer, protobufs
  File "/Users/ghart/Projects/github/caikit/caikit/caikit/core/data_model/base.py", line 39, in <module>
    from . import enums, json_dict
  File "/Users/ghart/Projects/github/caikit/caikit/caikit/core/data_model/json_dict.py", line 22, in <module>
    from google.protobuf.message_factory import GetMessageClass
ImportError: cannot import name 'GetMessageClass' from 'google.protobuf.message_factory' (/opt/miniconda3/envs/caikit/lib/python3.8/site-packages/google/protobuf/message_factory.py)

```

```
pip freeze | grep protobuf
protobuf==3.20.3
```

## Context

This was introduced with new usages of the `MessageFacotry` for `JsonDict` (https://github.com/caikit/caikit/pull/131)

gabe-l-hart:
## Description

Fixes #168

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,5 @@
 [tox]
-envlist = py, lint, fmt
+envlist = py, lint, fmt, proto3
 
 [testenv]
 description = run tests with pytest with coverage
@@ -24,7 +24,7 @@ package=wheel
 
 [testenv:docs]
 recreate = True
-deps = 
+deps =
     -r {toxinidir}/docs-requirements.txt
     alchemy-config>=1.1.1,<2.0.0
     alchemy-logging>=1.0.4,<2.0.0
@@ -80,3 +80,17 @@ setenv =
     FLIT_USERNAME = __token__
 commands = flit publish
 skip_install = True
+
+# Ensure compatibility is maintained with protobuf 3.X
+[testenv:proto3]
+description = run tests with pytest with coverage
+deps =
+    pytest>=6.2.5,<7.0
+    pytest-cov>=2.10.1,<3.0
+    pytest-html>=3.1.1,<4.0
+    tls_test_tools>=0.1.1
+    wheel>=0.38.4
+commands =
+    pip uninstall grpcio-health-checking -y
+    pip install protobuf==3.19.0 grpcio-health-checking --upgrade
+    pytest --cov=caikit --cov-report=html {posargs:tests}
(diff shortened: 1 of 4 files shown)


tox.ini
```diff
@@ -80,3 +80,17 @@ setenv =
...
+description = run tests with pytest with coverage
+deps =
+    pytest>=6.2.5,<7.0
+    pytest-cov>=2.10.1,<3.0
+    pytest-html>=3.1.1,<4.0
+    tls_test_tools>=0.1.1
+    wheel>=0.38.4
+commands =
+    pip uninstall grpcio-health-checking -y
+    pip install protobuf==3.19.0 grpcio-health-checking --upgrade
```
joerunde, line 95: 🤔 does this fail with the simple: 
```
deps =
    protobuf==3.19.0
```?
And if so does that mean users can't easily install with protobuf 3.19 anyway?

gabe-l-hart: Ug, this is just me punting on figuring out how to do `tox` right. The issue here is that the installation of the project requirements happens _after_ the installation of things in the `deps` list, so what happens is:

* `deps` install
* Generic install against `pyproject.proto` happen
    * This installs the _latest_ of `protobuf` and `grpcio-health-checking`
    * The `latest` `grpcio-health-checking` requires protobuf 4.X, but earlier versions that are in our range allow 3.X

Because of this ordering, the two other things I tried don't work:

* Put `protobuf==3.19.0` in `deps`: This fails because it gets overwritten when the install of `pyrpoject.proto` happens
* Just `pip install protobuf==3.19.0 --upgrade` in the `commands`: This fails because `grpcio-health-checking` is already installed at a version that is incompatible with `3.X`

PR #170, opened 2023-06-01
do not merge: test for gh actions permissions

joerunde:
just checking if @afrittoli's changes work when the PR author has write access


joerunde:
Nope, it in fact does not work that way :(

PR #164, opened 2023-05-31
Fix most unittest warnings

afrittoli:
Fix unit test warnings related to:
- deprecations
- unhandled thread exceptions

The only warnings left are related to a caikit deprecation: caikit/core/modules/config.py:66: DeprecationWarning: No module_id found in config.

This cleans-up the log output and makes it easier to detect new warnings.

PR #171, opened 2023-06-01
Remove the module_id warnings from unit tests

afrittoli:
Some of the test models - those defined in zip files, where not updated to the new config format with module_id, which caused a lot of warnings being logged when running unit tests.

Update the zipped models to use the module_id instead. Add a new test that specifically covers the no-module-id deprecation warning, verifying that the deprecation is raised via pytest.deprecated_call()


afrittoli:
@joerunde this is the second part after https://github.com/caikit/caikit/pull/164 🙏

tests/core/test_model_manager.py
```diff
@@ -185,6 +194,7 @@ def test_import_module_registry(self):
 TEST_DATA_PATH = os.path.join("tests", "fixtures")
 DUMMY_LOCAL_MODEL_NAME = "dummy_module_foo"
 DUMMY_BACKEND_MODEL_NAME = "dummy_module_backend"
+DUMMY_NO_ID_MODEL_NAME = "dummy_module_no_id"
```
evaline-ju, line 197: nit: this doesn't seem used - was this intended for the added test?

afrittoli: Nice catch, thank you, I defined it and forgot to use it in the test 🤦 
Fixed now

evaline-ju (approved):
LGTM - thanks for cleaning up the warnings!

PR #172, opened 2023-06-01
NoDanglingTestArtifacts: Use temp dirs instead of "test" everywhere

gabe-l-hart:
## Description

This PR cleans up the unit tests so that they don't depend on running from the base of the repo and they don't create a `test` directory with a bunch of temporary files in them that don't get cleaned up on completion.

In general, we want to _always_ use `tempfile` for any interactions with disk in tests unless there is a concrete reason not to.


alex-jw-brooks (approved):
Looks good to me - thanks for fixing this!

PR #167, opened 2023-06-01
💥 Generate inference rpcs with module task

abhishek-maurya-ibm:
- Update inference RPC creation with use of a module's task if present
- Removal of `get_module_info` function

Issue #90, opened 2023-05-03
Stop inferring tasks in service generation


joerunde:
## Is your feature request related to a problem? Please describe.

Modules will now have a `TASK_CLASS` attribute, pointing to a `@task` with required input parameters and an output type.

Currently the `service_generation` code tries to infer task by the model's module path which is obscure and flaky.
Additionally we rely on the `runtime.service_generation.primitive_data_model_types` to define what is "allowed" to be in the inputs for our tasks.

## Describe the solution you'd like

We can use the module.TASK_CLASS to group modules by task

We can turn the "which data models are allowed to be here" logic around and just check that the modules have the required input parameters to be included in the task rpc. Any other proto-able input parameters in the run signatures should be fine.

## Additional context

To help support users who have some modules in a task that accept extra data model parameters, which are currently disallowed

PR #173, opened 2023-06-01
💥 Remove primitive configs

joerunde:
This PR deletes all the gorpy `is_primitive` logic and the ever confusing `runtime.service_generation.primitive_data_model_types` configurations.

As a follow-on to #167, this should actually finish up #90 and help with #30

This is all made possible by the explicit `tasks`, which now validate up-front that a module has the required input parameters and output type. The remainder of the work at service-generation time is to strip any non-protobufable args out of the request messages, which has much simpler implementation logic than what we were previously doing

### DIFF
--- a/caikit/runtime/service_generation/protoable.py
+++ b/caikit/runtime/service_generation/protoable.py
@@ -129,7 +129,7 @@ def extract_data_model_type_from_union(arg_type: Type) -> Type:
 
 def is_protoable_type(arg_type: Type) -> bool:
     """
-    Returns True is arg_type is in PROTO_TYPE_MAP(float, int, bool, str, bytes)
+    Returns True if arg_type is in PROTO_TYPE_MAP(float, int, bool, str, bytes)
     Or if it's an imported Caikit data model class.
     Or if it's a Union of at least one of those.
     Or if it's a List of one of those.
--- a/caikit/runtime/service_generation/protoable.py
+++ b/caikit/runtime/service_generation/protoable.py
@@ -134,6 +134,9 @@ def is_protoable_type(arg_type: Type) -> bool:
 
     if typing.get_origin(arg_type) == list:
         log.debug2("Arg is List")
+        if len(typing.get_args(arg_type)) == 0:
+            log.debug2("List annotation has no type")
+            return False
         return typing.get_args(arg_type)[0] in proto_primitive_set
 
     if typing.get_origin(arg_type) == Union:
--- a/caikit/runtime/service_generation/protoable.py
+++ b/caikit/runtime/service_generation/protoable.py
@@ -21,10 +21,10 @@ import inspect
 import typing
 
 # First Party
+from py_to_proto.dataclass_to_proto import PY_TO_PROTO_TYPES
 import alog
 
 # Local
-from .type_helpers import PROTO_TYPE_MAP
 from caikit.core.data_model.base import DataBase
 
 log = alog.use_channel("PROTOABLES")
@@ -134,7 +134,7 @@ def is_protoable_type(arg_type: Type) -> bool:
     Or if it's a Union of at least one of those.
     Or if it's a List of one of those.
     False otherwise"""
-    proto_primitive_set = list(PROTO_TYPE_MAP.keys())
+    proto_primitive_set = list(PY_TO_PROTO_TYPES.keys())
 
     if arg_type in proto_primitive_set:
         return True
--- a/caikit/runtime/service_generation/protoable.py
+++ b/caikit/runtime/service_generation/protoable.py
@@ -21,11 +21,11 @@ import inspect
 import typing
 
 # First Party
-from py_to_proto.dataclass_to_proto import PY_TO_PROTO_TYPES
 import alog
 
 # Local
 from caikit.core.data_model.base import DataBase
+from caikit.core.data_model.dataobject import DATAOBJECT_PY_TO_PROTO_TYPES
 
 log = alog.use_channel("PROTOABLES")
 
@@ -134,7 +134,7 @@ def is_protoable_type(arg_type: Type) -> bool:
     Or if it's a Union of at least one of those.
     Or if it's a List of one of those.
     False otherwise"""
-    proto_primitive_set = list(PY_TO_PROTO_TYPES.keys())
+    proto_primitive_set = list(DATAOBJECT_PY_TO_PROTO_TYPES.keys())
 
     if arg_type in proto_primitive_set:
         return True
--- a/caikit/runtime/service_generation/protoable.py
+++ b/caikit/runtime/service_generation/protoable.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """
-This file contains our logic about what constitutes a "primitive" for RPC generation purposes
+This file contains our logic about what constitutes a proto-able for RPC generation purposes
 """
 
 # Standard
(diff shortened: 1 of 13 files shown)


tests/core/test_task.py
```diff
@@ -130,6 +131,43 @@ def run(self, foo: int):
                 pass
 
 
+def test_task_validation_throws_on_wrong_return_type():
```
joerunde, line 134: These were missing test cases in the task validator

caikit/runtime/service_generation/protoable.py
```diff
@@ -0,0 +1,158 @@
...
+        return True
+    if isinstance(arg_type, type) and issubclass(arg_type, DataBase):
+        return True
+
+    if typing.get_origin(arg_type) == list:
+        log.debug2("Arg is List")
+        # check that list is not nested
+        if len(typing.get_args(arg_type)) == 1:
+            return typing.get_args(arg_type)[0] in proto_primitive_set
+        # TODO: that check is always true
```
evaline-ju, line 149: which check is always true? This is a bit unclear - could this be rephrased to what needs to be done?

caikit/runtime/service_generation/protoable.py
```diff
@@ -0,0 +1,158 @@
...
+            if len(union_protoables) == 1:
+                return union_protoables[0]
+            # otherwise, try to get the data model objects in the Union
+            dm_types = [
+                arg
+                for arg in union_protoables
+                if inspect.isclass(arg) and issubclass(arg, DataBase)
+            ]
+            # if there are multiple, pick the first one
+            if len(dm_types) > 0:
```
gabe-l-hart, line 84: Ah, this logic is only for dm types. Is the idea here that `Union[RawDocument, str]` should result in just `RawDocument`?

caikit/runtime/service_generation/protoable.py
```diff
@@ -0,0 +1,158 @@
...
+def extract_data_model_type_from_union(arg_type: Type) -> Type:
+    """Helper function that determines the right data model type to use from a Union"""
+
+    # Decompose this type using typing to determine if it's a useful typing hint
+    typing_origin = get_origin(arg_type)
+    typing_args = get_args(arg_type)
+
+    # If this is a data model type, no need to do anything
+    if isinstance(arg_type, type) and issubclass(arg_type, DataBase):
+        return arg_type
```
gabe-l-hart, line 109: NIT: above you used `reflection.is_class` for this. I kind of like that since it's a little more explicit, but it would probably be good to be consistent

gabe-l-hart: Also, it looks like this same check is repeated in a lot of places, so it might be worth consolidating it into a helper function

caikit/runtime/service_generation/protoable.py
```diff
@@ -0,0 +1,158 @@
...
+        ]
+        if dm_types:
+            log.debug2(
+                "Found data model types in Union: [%s], taking first one", dm_types
+            )
+            return extract_data_model_type_from_union(dm_types[0])
+
+    # if it's anything else we just return as is
+    # we don't actually want to throw errors from service generation
+    log.warning("Return type [%s] not a DM type, returning as is", arg_type)
```
gabe-l-hart, line 126: It's an interesting question whether we want to actually support simple primitive return types. Not for this pr, though

caikit/runtime/service_generation/protoable.py
```diff
@@ -0,0 +1,158 @@
...
+    return arg_type
+
+
+def is_protoable_type(arg_type: Type) -> bool:
+    """
+    Returns True is arg_type is in PROTO_TYPE_MAP(float, int, bool, str, bytes)
```
gabe-l-hart, line 132:
```suggestion
    Returns True if arg_type is in PROTO_TYPE_MAP(float, int, bool, str, bytes)
```

caikit/runtime/service_generation/protoable.py
gabe-l-hart: I think the `List[]` type hint always takes exactly one arg. It can do no args, though, I think

gabe-l-hart (changes requested):
A couple of nits and a request to consolidate proto type maps that I probably should have caught a while ago

caikit/runtime/service_generation/protoable.py
```diff
@@ -0,0 +1,158 @@
...
+
+    If arg_type is a union, then this will return the union back if all types in it are proto-able,
+    or the first proto-able arg type if the union has non-protoable arg types.
+
+    If arg_type is not a union nor protoable at all, this returns None.
+
+    Examples:
+    Union[protoable_type, non_protoable_type] -> protoable_type
+    Union[protoable_type_1, protoable_type_2] -> Union[protoable_type_1, protoable_type_2]
+    Union[protoable_type_1, protoable_type_2, non_protoable_type] -> protoable_type_1
```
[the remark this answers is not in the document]
joerunde: I agree, I just didn't want to change behavior if @prashantgupta24 had it like this for a good reason for his use case.

I can go ahead and simplify, and check back in with that team about it

caikit/runtime/service_generation/protoable.py
joerunde: Looks like it can only do a single arg, zero and >1 args both fail

```
>>> foo: List[]
  File "<stdin>", line 1
    foo: List[]
              ^
SyntaxError: invalid syntax
>>> foo: List[int, float]
  File "<stdin>", line 1, in <module>
  File "/opt/homebrew/Cellar/python@3.11/3.11.3/Frameworks/Python.framework/Versions/3.11/lib/python3.11/typing.py", line 351, in inner
    return func(*args, **kwds)
           ^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.11/3.11.3/Frameworks/Python.framework/Versions/3.11/lib/python3.11/typing.py", line 1556, in __getitem__
    _check_generic(self, params, self._nparams)
  File "/opt/homebrew/Cellar/python@3.11/3.11.3/Frameworks/Python.framework/Versions/3.11/lib/python3.11/typing.py", line 281, in _check_generic
    raise TypeError(f"Too {'many' if alen > elen else 'few'} arguments for {cls};"
TypeError: Too many arguments for typing.List; actual 2, expected 1
```

joerunde: and for the simple `list` case:
```
>>> typing.get_origin(list) == None
True
```

caikit/runtime/service_generation/protoable.py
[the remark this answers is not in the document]
gabe-l-hart: Ah, I didn't realize this was a compatibility thing. Fine to merge as is and add a ticket to fix later once Prashant can comment (or maybe @tharapalanivel would know?)

caikit/runtime/service_generation/protoable.py
gabe-l-hart:
```py
>>> from typing import List, get_origin
>>> get_origin(List)
<class 'list'>
```

gabe-l-hart: So I think you need to keep the check rather than moving to an `assert` (which is what I was going to suggest)

gabe-l-hart (approved):
Still might be worth consolidating logic for checking if an arbitrary thing is a subclass of `DataBase`, but certainly not required. Good to go otherwise

caikit/runtime/service_generation/type_helpers.py
```diff
@@ -14,14 +14,14 @@
 """
 Some type conversion helpers for going between python and protocol buffer interfaces
 """
-
 # Standard
 from inspect import isclass
 from typing import Optional, Type, Union, get_args, get_origin
+import inspect
```
gabe-l-hart:
```suggestion
from inspect import isclass
from typing import Optional, Type, Union, get_args, get_origin
```

caikit/runtime/service_generation/type_helpers.py
```diff
@@ -63,5 +63,9 @@ def is_model_type(arg_type: Type) -> bool:
     return False
 
 
+def is_data_model_type(arg_type: Type) -> bool:
+    return inspect.isclass(arg_type) and issubclass(arg_type, DataBase)
```
gabe-l-hart, line 67:
```suggestion
    return isclass(arg_type) and issubclass(arg_type, DataBase)
```

gabe-l-hart (changes requested):
Tiny cleanup to avoid duplicate import of `isclass`

gabe-l-hart (approved):
🚀

Issue #177, opened 2023-06-02
Excessive and unclear warnings about "Could not deduct output" on startup.


markstur:
## Describe the bug

Excessive and unclear warnings about "Could not deduct output" on startup.

## Platform

Please provide details about the environment you are using, including the following:

- Interpreter version: 3.8
- Library version: 0.6.0

## Sample Code

caikit-huggingface-demo

## Expected behavior

These are unclear (no function name) an in this case are not useful because they are all complaining about train() which is not implemented (not needed) in this app.  In this case, these should not be warnings but probably debug.

## Observed behavior

Could not deduct output type from function (*args, **kwargs) for module class Embeddings.
Could not deduct output type from function (*args, **kwargs) for module class Conversational.
Could not deduct output type from function (*args, **kwargs) for module class ImageClassification.
Could not deduct output type from function (*args, **kwargs) for module class ImageSegmentation.
Could not deduct output type from function (*args, **kwargs) for module class ObjectDetection.
Could not deduct output type from function (*args, **kwargs) for module class Sentiment.
Could not deduct output type from function (*args, **kwargs) for module class Summarization.
Could not deduct output type from function (*args, **kwargs) for module class TextGeneration.

## Additional context

ModuleBase train() throws NotImplementedError anyway.  So if this was an actual problem it'll be obvious at runtime.  I think debug logging, if any, would be adequate in that case.

markstur:
I'd like to contribute a fix for this, but I cannot assign.

PR #178, opened 2023-06-02
Improve warning message and suppress it for base functions

markstur:
Improving this warning: "Could not deduct output type from function %s for module class %s."

* Log fn.__name__ (e.g. train) which is more useful than a signature like () or (*args, **kwargs)
* Don't log warning for functions in the base module
  - Because it is too spammy for optional overrides like train()
  - It'll be obvious enough if these functions are accidentally called at runtime
  - Keep and enhanced version of the message using log.debug for the base method case.

Fixes: #177

### DIFF
--- a/caikit/core/signature_parsing/parsers.py
+++ b/caikit/core/signature_parsing/parsers.py
@@ -65,11 +65,21 @@ def get_output_type_name(
         return type_from_docstring
 
     # If we get here, it means no annotation or docstring for type was provided
-    log.warning(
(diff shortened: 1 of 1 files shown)


markstur:
I didn't add tests because the current tests at least cover the code and I didn't think mocking a log() or extracting a method just to test is_base() was particularly helpful, but up to reviewers.

gabe-l-hart (approved):

gabe-l-hart:
Thanks for the cleanup Mark!

Issue #136, opened 2023-05-23
Delete `build_caikit_library_request_dict`


joerunde:
## Description

As a caikit maintainer, I want to burn `build_caikit_library_request_dict` to the ground because it replicates all the logic from `DataModel`s

## Discussion

Previously in history, runtime request messages themselves were not data model objects, so we required custom deserialization logic to `.from_proto` each field individually.

This is no longer the case, the request message itself is now a data model object, so it has the ability to `.from_proto` itself.
However, there is some extra runtime-specific logic that we would want to keep around, like:
- Fields that were not set at all should not be passed along
- For task inference RPCs, if fields were set that the requested model does not support, they should also not be passed along

There may be some more custom logic hiding in that function as well that we'll need to dig into a lil' bit

## Acceptance Criteria

- [ ] Unit tests cover new/changed code
- [ ] Examples build against new/changed code
- [ ] READMEs are updated
- [ ] Type of [semantic version](https://semver.org/) change is identified

PR #182, opened 2023-06-06
:recycle: Use data model deserialization to build request kwargs dict

evaline-ju:
This PR refactors `build_caikit_library_request_dict` to use existing data model logic and module signatures - specifically now that request messages are all data model objects, `.from_proto` can be called on each data model object instead of handling all field values individually.
- Mainly a few cases are focused on - any "unset" fields, or fields not present in the request, are removed, any fields absent from the module signature are removed, and model pointers are handled separately (since the model instance has to be retrieved). 
- `to_kwargs` was added so that any field names backed by oneofs could be preserved. Otherwise only the internal field names would persist. `to_dict` cannot be used since we do not want recursive conversion.

This changes a few behaviors present before the refactor:
- Previously, for datastreams, the datastream source was fetched and `to_data_stream` was called. This meant error handling that would have been done in `to_data_stream` no longer happens at request creation time but will happen when the request is used (i.e. the deserialization logic is on the datastream class and will not be treated specially compared to other classes).
- The `training_data` field is no longer specially handled. There was "default" behavior to initialize an empty stream if training_data was not passed in by the user but on the module signature without default or marked optional. The changes here are made with the assumption that this behavior was incorrect - if a module signature has field without a default and not marked Optional, this field is required. An empty stream should not be initialized in this case - a user should be informed through error that this required field is missing

Key test changes
- `sleep` was removed for those tests waiting for training to finish and then calling `sleep`
- Tests not properly waiting for training to finish had polling added and sleep removed
- `primitive_party` proto and compiled file are removed. We no longer support compiled service generation, and data models could not be deduced from just the proto/compiled file. There is still testing for primitives done through the `SamplePrimitiveModule`.

Closes: #136 

Co-authored-by: Joe Runde <Joseph.Runde@ibm.com>


tests/runtime/utils/test_servicer_util.py
```diff
@@ -520,145 +439,5 @@ def test_build_caikit_library_request_dict_works_when_data_stream_directory_incl
         # no error because at least 1 json file exists within the provided dir
         caikit.core_request = build_caikit_library_request_dict(
             train_request,
-            sample_lib.modules.sample_task.SampleModule().train,
-        )
-
-
-# ---------------- Error tests for build_caikit_library_request_dict --------------------
```
evaline-ju, line 527: Removing these error tests that depend on `to_data_stream`. These errors will occur not at request building time but at request usage time

tests/runtime/servicers/test_global_train_servicer_impl.py
```diff
@@ -421,6 +388,9 @@ def test_global_train_returns_exit_code_with_oom(
 #####################################################################
 
 
+@pytest.mark.skip(
+    reason="This test fails intermittently. Functionality has to be debugged for race condition"
+)
```
evaline-ju, line 393: While not parametrized here, we discovered that this test will fail when run ~10 times

caikit/runtime/utils/servicer_util.py
```diff
@@ -211,209 +212,70 @@ def validate_data_model(
...
+                kwargs_dict.pop(unset_field_name)
+
+        # 2. Remove any fields not in the module signature
+        absent_field_names = [
+            field
+            for field in kwargs_dict.keys()
+            if field not in module_signature.parameters.keys()
+        ]
+        for absent_field_name in absent_field_names:
+            if absent_field_name in kwargs_dict:
```
prashantgupta24, line 265: is this a redundant check? The `absent_field_names` list is populated by iterating through all `fields` in `kwargs_dict.keys()` right?

prashantgupta24 (approved):
Looks good! 2 nits:
- Redundant check (see above)
- since we now get lists for lists in function arguments and not datastreams, it’s worth adding a test case for that?

evaline-ju:
> since we now get lists for lists in function arguments and not datastreams, it’s worth adding a test case for that?

Yes @prashantgupta24 this should actually be tested already -`test_global_predict_build_caikit_library_request_dict_works_for_set_primitives`  https://github.com/caikit/caikit/pull/182/files#diff-023b0d0b974e4243f7feb408bfc6486b6998258d6aceb39b59d2064d0ea8467bR266 (may need to expand `test_servicer_util.py` to see this)

joerunde (approved):
LGTM!

The one check is failing with the coverage report publishing that is giving us trouble, but the tests and linting all look 💯

PR #183, opened 2023-06-06
:bug: Remove manual training_id initialization

gkumbhat:
### Changes
- Fix long running cancellation test


joerunde (approved):
lgtm? But I don't think it fixes anything :(

gabe-l-hart (approved):
Ok, thanks for pointing that out @joerunde. I agree that this is good, but doesn't look like it will actually fix anything

PR #189, opened 2023-06-07
Dockefiles for the grpc and rest gateway

alanbraz:
@markstur take a look at this idea of Dockerfile for caikit grpc and adding the rest gateway too. This will make the example deployable and accessible using a rest API.

Still pending the deployment steps in a cluster, but I don't know if this is the right place.

For your information @gabe-l-hart @joerunde


markstur (commented):
I'm hoping to try this soon but not sure if I'll really get that done.  @alanbraz I've used an older version of this so I had a question about where we get golang.  See inline.

markstur:
Tested the gateway hitting my local server like this and it works:

docker run -it -e PROXY_ENDPOINT=host.docker.internal:8085 -p 8080:8080 2f804936fa13

markstur:
Just to try it out I now tested with both Dockerfile and Dockerfile.gateway and got them to work together like this:

```
# create docker network
docker network create test-net
```

```
# run grpc server  (Dockerfile image) on the network with a host named "grpc"
docker run -it --network test-net --name grpc --rm bbff1a071f36
```
In another terminal:
```
# run the Dockerfile.gateway image (rest server) on the same network and tell it where the grpc server is:
docker run -it --rm -e PROXY_ENDPOINT=grpc:8085 -p 8080:8080 --network test-net 2f804936fa13
```

So that's cool.  It works.  Not sure about all the documentation.  I prefer deploy to OpenShift it's hard to figure the right way to document things.  I'm just sharing the docker tips up there to save/share my testing notes.

gabe-l-hart:
@alanbraz @markstur is this still under development? If not, let's choose for now and we can always reopen later

alanbraz:
If the gateway won't be needed anymore, I think we can drop this one!

Issue #180, opened 2023-06-05
Migrate from GabeStyle to Google Style docstrings


joerunde:
## Is your feature request related to a problem? Please describe.

Our docstrings are mostly in @gabe-l-hart style, which is close to but not quite the Google docstring format. We would like to migrate to the Google docstring format so that tools like Sphinx can understand the docstrings and create good user-facing resources. We would also like this style to be enforced going forward, so that nothing ends up documented in a different style.

## Describe the solution you'd like

All docstrings are in Google format

## Describe alternatives you've considered

Other formats? But once we're in Google format there are public tools available to migrate between them.

## Additional context

This will probably require a custom conversion script, nobody should be doing this by hand.

As a bonus, we could consider requiring docstrings on all "public" functions (things that don't start with `_`)

joerunde:
@WillMJ https://www.sphinx-doc.org/en/master/usage/extensions/example_google.html

afrittoli:
We could use something like the `docstrings` plugin for `flake8`, it supports the Google format of docstring, e.g.

```
➜ flake8 --docstring-convention google | grep ': D'
./__init__.py:16:1: D200 One-line docstring should fit on one line with quotes
./base.py:16:1: D200 One-line docstring should fit on one line with quotes
./base.py:70:1: D202 No blank lines allowed after function docstring
./base.py:70:1: D205 1 blank line required between summary line and description
./base.py:70:1: D415 First line should end with a period, question mark, or exclamation point
./base.py:156:1: D202 No blank lines allowed after function docstring
./base.py:156:1: D205 1 blank line required between summary line and description
./base.py:156:1: D415 First line should end with a period, question mark, or exclamation point
./base.py:289:1: D415 First line should end with a period, question mark, or exclamation point
./base.py:341:1: D202 No blank lines allowed after function docstring
(...)
```

Once the check is in place, we can start addressing the different style rules one at the time and enforce them in the `flake8` config.

afrittoli:
See #186 too.

sboagibm:
@joerunde said:
> As a bonus, we could consider requiring docstrings on all "public" functions (things that don't start with _)

I think this is too prescriptive, and not practical.  For some very public interfaces, like `run` on blocks, an enforcement mechanism makes sense.

PR #191, opened 2023-06-07
Converts docstrings from Gabe Style to Google Style

WillMJ:
#180 
Created a script in scripts called dsconverter.py which takes a file or directory name and converts that file or all .py files in that directory from Gabe Style Docstrings to Google Style Docstrings. I also created a testenv in tox.ini which allows the script to be easily run. However, there is a problem I am unsure how to solve in this script which gets rid of a lot of the formatting in the python files (explaining the 9,126 deletions) when it writes the new docstrings. As far as I'm aware, it doesn't break the code - just breaks the formatting. Any idea as to what could be causing the squashed code or what the fix could be?

### DIFF
--- a/caikit/runtime/service_factory.py
+++ b/caikit/runtime/service_factory.py
@@ -37,7 +37,6 @@ from caikit.interfaces.runtime.data_model import (
 )
 from caikit.runtime import service_generation
 from caikit.runtime.service_generation.rpcs import snake_to_upper_camel
-from caikit.runtime.types.caikit_runtime_exception import CaikitRuntimeException
 from caikit.runtime.utils import import_util
 import caikit.core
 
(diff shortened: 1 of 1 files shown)


caikit/__init__.py
```diff
@@ -3,11 +3,6 @@
 model APIs while enabling AI developers build algorithms and models in a
 modular/composable framework.
 """
-# Local
```
gabe-l-hart, line 6: We definitely need to be careful to not delete these comments

caikit/config/__init__.py
```diff
@@ -1,21 +1,3 @@
-# Copyright The Caikit Authors
```
gabe-l-hart, line 1: We need to make sure we retain these copyright headers

gabe-l-hart (changes requested):
A quick look through the PR looks like your script is making a lot of changes that it's not intended to make. Specifically, it looks like it's removing all of the comments from top-level files and it's removing whitespace (empty lines) in a lot of places. We'll need to make sure that isn't happening before moving forward

caikit/runtime/interceptors/caikit_runtime_server_wrapper.py
```diff
@@ -52,12 +33,18 @@ def __init__(self, server, global_predict, intercepted_svc_descriptor):
                 gRPC request message and a grpc.ServicerContext, and return
                 a suitable gRPC response message
         """
+<<<<<<< Updated upstream
 
+=======
+        if not excluded_methods:
+            excluded_methods = []
+>>>>>>> Stashed changes
```
joerunde, line 41: @WillMJ you also have unresolved stash conflicts in here!

caikit/config/config.py
```diff
@@ -62,8 +62,9 @@ def configure(
             with overrides for your usage.
         config_dict (Optional[Dict]): Config overrides in dictionary form
 
-    Returns: None
-        This only sets the config object that is returned by `caikit.get_config()`
+    Returns:
+			None: This only sets the config object that is returned by `caikit.get_config()`
```
joerunde, line 66: indents look pretty excessive here :D

caikit/runtime/servicers/global_predict_servicer.py
```diff
@@ -131,8 +131,8 @@ def Predict(self, request, context):
             context(grpc.ServicerContext): Context object (contains request metadata, etc)
 
         Returns:
-            response (object):
-                A Caikit Library data model response object
+			response (object):: A Caikit Library data model response object
```
joerunde, line 134: double `::`!

caikit/config/config.py
WillMJ: They are! Have this as a set value which I shouldn't.

caikit/runtime/service_generation/primitives.py
```diff
@@ -0,0 +1,117 @@
+"""
```
joerunde, line 1: I think this file was deleted (or renamed).... looks like you should check out the latest `caikit/caikit` off main before re-running the script again

caikit/runtime/work_management/destroyable_thread.py
```diff
@@ -137,8 +136,8 @@ def get_or_throw(self):
         """
             After the thread has completed it's work, call this to get the output.
         Returns:
-            The resulting value of runnable_func(*runnable_args, **runnable_kwargs)
-        Raises:
+			The resulting value of runnable_func(*runnable_args, **runnable_kwargs): Raises:
```
joerunde, line 139: No newline any more before `Raises`?

caikit/config/config.py
```diff
@@ -133,13 +134,12 @@ def merge_configs(
...
             The override config
-        merge_strategy: str
-            The merging strategy, either `merge` or `override`
+        merge_strategy (str): The merging strategy, either `merge` or `override`
             `override` will replace values in base with those from overrides
             `merge` will deep-merge dictionaries and prepend-merge lists
     Returns:
-        merged: dict
-            The merged results of overrides merged onto base
+			merged: dict: The merged results of overrides merged onto base
```
joerunde, line 141: These are tricky: the name of the thing to be returned (`merged` here) should not be in the docstring and we'd probably want to go ahead and delete it. If you can find a good enough heuristic to catch most of these cases we can probably clean up what's left

caikit/core/data_model/base.py
```diff
@@ -589,8 +588,8 @@ def from_proto(cls, proto):
                 A protocol buffer to serialize from.
 
         Returns:
-            protobufs
-                A DataBase object.
+			protobufs: A DataBase object.
```
joerunde, line 591: hah, this is another funky case of the named return problem. `protobufs` is not a type but `DataBase` is...

Dunno how you'd catch this case, we'll probably just need to come clean these up to read more like
```
Returns:
    DataBase: An instance of the data model class
```

scripts/dsconverter.py
```diff
@@ -0,0 +1,154 @@
...
+
+        Args:
+            custom_docstring (str): The original docstring.
+
+        Returns:
+            str: The converted docstring in Google style.
+        """
+
+        # Extract the name, data type, and description of arguments using regular expressions
+        arg_pattern = r"(?!Args)(?!rgs)(?!gs)(?!s)(\*\*\w+|\*\w+|\w+):\s*((?:\w+\s*\|\s*)*(?:\w+\s*\|)?\s*(?:\w+(?:\.\w+)*\s+\([^)]+\)|\w+(?:\.\w+)*|\w+(?:\.\w+)*\s*\|\s*\w+(?:\.\w+)*))\s*\n\s*((?:(?!Returns:).)*?)($|\n)"
```
joerunde, line 63: 🤯

joerunde (changes requested):
Needs some test cases!

If you wanna build those out of the current little problems from running this on `caikit`, that'd be 🌶️🌶️🌶️

PR #194, opened 2023-06-08
Add coverage report and threshold

hickeyma:
Add coverage report thanks to the python-coverage GitHub action https://github.com/marketplace/actions/python-coverage

The action reports the coverage as a GitHub comment. It also allows defining thresholds for the level of coverage. For now, setting an 80% thresholds for new files and zero for existing content. Once the report is published to PRs we can start working on bringing the coverage up to the desired level and enable the threshold for existing code.

We run unit tests for four different versions of python, but I don't think we need coverage for each.
Nonetheless, I changed the tox.ini to make sure that the coverage reports from the four runs do not override each other, and produce both html and xml reports.
Html is convenient when running locally, xml is needed by the action.

The action uses the coverage report for python 3.11

Fixes: #146

### DIFF
--- a/.github/workflows/build-library.yml
+++ b/.github/workflows/build-library.yml
@@ -23,6 +23,10 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      id-token: write
+      pull-requests: write
     strategy:
       matrix:
         python-version:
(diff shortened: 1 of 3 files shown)


afrittoli (approved):
Thanks @hickeyma

hickeyma:
This PR was a duplicate of #166 and what pushed to test if coverage issue was with a PR pushed from a fork.

Duplicate of #166

Issue #186, opened 2023-06-06
Improve code linting


afrittoli:
## Is your feature request related to a problem? Please describe.

Today we don't catch things like unused imports.
Statistics from `flake8` with default config:

```
2     E203 whitespace before ':'
21    E266 too many leading '#' for block comment
2     E302 expected 2 blank lines, found 1
7     E402 module level import not at top of file
1474  E501 line too long (87 > 79 characters)
3     E712 comparison to False should be 'if cond is False:' or 'if not cond:'
1     E713 test for membership should be 'not in'
1     E722 do not use bare 'except'
80    F401 '.core' imported but unused
10    F403 'from .model_manager import *' used; unable to detect undefined names
1     F405 'ModelManager' may be undefined, or defined from star imports: .model_manager, .module_backends, .toolkit
1     F523 '...'.format(...) has unused arguments at position(s): 1, 2
6     F821 undefined name 'DataModelBackendBase'
1     W291 trailing whitespace
3     W293 blank line contains whitespace
```

## Describe the solution you'd like

We could complement `pylint` with `flake8`, or configure `pylint`, if possible, to discover more issues.
Rule `E501` can be tricky to achieve in some cases, and not everyone likes it, so that one we could skip, but most other rules are probably good to apply.

`flake8` has a nice plugins, like `flake8-docstring` to check docstring formatting.

## Describe alternatives you've considered

- Enhance `pylint` config - I'm not sure if pylint could address all of `flake8` findings
- Use both `flake8` and `pylint`. It may feel like an overhead, but it should be possible to configure both linter so that they do not conflict
- Replace `pylint` with `flake8`. Need to check which features in `pylint` are not avaialble in `flake8`

afrittoli:
The `pydeps` check used today as part of `tox -e imports` in the lint CI job does not check for unused imports.

joerunde:
@afrittoli I think we do check for unused imports, `tox -e lint` does yell about those though it only checks our source code and not our test code at the moment. `tox -e imports` is a bit different, it runs some custom import checking logic that makes sure that we keep our internal structure clean, where `caikit.core` is not allowed to import from `caikit.interfaces` nor `caikit.runtime`

We also do have our line-length configured with pylint at 100 characters right now, and I agree that 79 characters is a bit restrictive. I do agree though that some of these extra flake8 checks would be good to have, these seem to highlight some cruft that we know exists but haven't yet bothered to clean up. I'd be happy to have those enforced!

PR #196, opened 2023-06-08
Add lint for code under examples

markstur:
The existing tox lint was only processing the caikit module.

This adds:
* examples/*.py: example files outside of text-sentiment
* examples/text-sentiment/text_sentiment: the text_sentiment module
* examples/text-sentiment/*.py: text_sentiment files outside of the module

Note: Without the *.py use outside of modules you can get into recursion errors processing an arbitrarily named "venv" dir.

Part of: #186

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -62,7 +62,7 @@ skip_install = True # Skip package install since fmt doesn't need to execute cod
 [testenv:lint]
 description = lint with pylint
 deps = pylint>=2.16.2,<3.0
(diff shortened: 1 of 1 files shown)


markstur:
Marking as a draft because this will cause tox errors until examples are cleaned up.

markstur:
This is ready for review but should merge #204  first so that this doesn't break CI.

markstur:
combining with #204 and closing this one to address review

PR #199, opened 2023-06-08
Run coverage on pull_request only

afrittoli:
The coverage action does not support push events.
Only run the coverage step on pull_request.


joerunde (approved):
We think this might be failing just because this PR is editing the workflow, gonna merge and see if a subsequent PR fails as well

Issue #151, opened 2023-05-26
Testing the examples in CI


afrittoli:
## Is your feature request related to a problem? Please describe.

We should test the examples as part of CI to ensure that they are kept up to date with the latest changes to caikit so that the project always provides a great experience for newcomers trying out caikit. 

## Describe the solution you'd like

A CI job that runs all the examples under https://github.com/caikit/caikit/tree/main/examples

## Describe alternatives you've considered

The alternative today is to ask PR authors and reviewers to check the examples manually.
It could be added to the PR template as a checkbox to remind contributors about it.

## Additional context

See the issue solved in #145

markstur:
Also see #184

PR #200, opened 2023-06-08
Add E2E coverage to example scenarios

afrittoli:
Add e2e test for examples

Add an e2e test for the text sentiment example.
The test is written as a standard python test so it can be
discovered by pytest and all the nice python tooling for
tests works as usual.

The test creates a new virtual environment and installs
caikit from the local disk, to ensure it tests the version of
caikit from the pull request (or local code changes).

The test then follows the same process described in the README:
- installs requirements.txt
- runs the start_runtime.py script
- runs the client.py

This test ensures that the example stays in a working state.
The example is likely to be one of the first points of contact
of people looking at caikit and it should always be working.

This test also helps in detecting unintended API changes.
Eventually we should have a API driven test suite that covers
the whole API.

The new test is not executed by default via tox, to keep
tests execution on laptops as quick as possible. The new
test is enabled in the GitHub action workflow.

Fixes: #151

Signed-off-by: Andrea Frittoli <andrea.frittoli@uk.ibm.com>

### DIFF
--- a/.github/workflows/build-library.yml
+++ b/.github/workflows/build-library.yml
@@ -49,4 +49,4 @@ jobs:
           python -m pip install --upgrade pip
           python -m pip install -r setup_requirements.txt
       - name: Build and test with tox
-        run: tox -e ${{ matrix.python-version.tox }}
+        run: tox -e ${{ matrix.python-version.tox }} -- tests
--- a/examples/text-sentiment/client.py
+++ b/examples/text-sentiment/client.py
@@ -12,6 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# Standard
+import os
+
 # Third Party
 import grpc
 
@@ -24,12 +27,12 @@ inference_service = ServicePackageFactory().get_service_package(
 )
 
 port = 8085
-channel = grpc.insecure_channel(f"localhost:{port}")
 
+# Setup the client
+channel = grpc.insecure_channel(f"localhost:{port}")
 client_stub = inference_service.stub_class(channel)
 
-# print(dir(client_stub))
-
+# Run inference for two sample prompts
 for text in ["I am not feeling well today!", "Today is a nice sunny day"]:
     input_text_proto = TextInput(text=text).to_proto()
     request = inference_service.messages.HuggingFaceSentimentTaskRequest(
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,3 +38,8 @@ dependencies = [
 
 [project.urls]
 Source = "https://github.com/caikit/caikit"
+
+[tool.pytest.ini_options]
+markers = [
+    "examples: marks tests as e2e examples (deselect with '-m \"not examples\"')",
+]
--- a/tests/examples/shared.py
+++ b/tests/examples/shared.py
@@ -1,0 +1,75 @@
+# Copyright The Caikit Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# Standard
+
+# Standard
+from contextlib import contextmanager
+from os import path
+from typing import Tuple
+import asyncio
+import subprocess
+import tempfile
+import time
+import venv
+
+# Third Party
+import pytest
+
+caikit_dir = path.abspath(path.join(path.dirname(__file__), "..", ".."))
+
+
+async def waitForPort(port, timeout: int) -> bool:
+    tmax = time.time() + timeout
+    while time.time() < tmax:
+        try:
+            _, writer = await asyncio.wait_for(
+                asyncio.open_connection("127.0.0.1", port), timeout=5
+            )
+            writer.close()
+            await writer.wait_closed()
+            return True
+        except:
+            await asyncio.sleep(1)
+    return False
+
+
+@contextmanager
+def requirements(example: str) -> Tuple[str, str]:
+    """Installs caikit and requirements.txt from relative path, returns python from venv
+    and absolute path to the example.
+    Raises subprocess.CalledProcessError if installation fails
+    """
+    with tempfile.TemporaryDirectory() as venv_dir:
+        python = path.abspath(path.join(venv_dir, "bin", "python3"))
+        pip = path.abspath(path.join(venv_dir, "bin", "pip3"))
+        example_dir = path.abspath(path.join(caikit_dir, "examples", example))
+
+        # Create a venv, install local version of caikit and local version of requirements
+        venv.create(env_dir=venv_dir, system_site_packages=False, with_pip=True)
+        try:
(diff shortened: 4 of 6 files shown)


markstur (changes requested):
use the config override ability instead of adding code to do it

.github/workflows/build-library.yml
```diff
@@ -28,15 +28,20 @@ jobs:
         python-version:
           - setup: '3.8'
             tox: 'py38'
+            grpc-port: 10085
```
afrittoli, line 31: Different ports make sure the test matrix can run without conflict.

pyproject.toml
```diff
@@ -38,3 +38,8 @@ dependencies = [
 
 [project.urls]
 Source = "https://github.com/caikit/caikit"
+
+[tool.pytest.ini_options]
+markers = [
+    "examples: marks tests as e2e examples (deselect with '-m \"not examples\"')",
+]
```
afrittoli, line 45: A custom test marker is used to make examples e2e tests. 
Example tests are slightly longer running, so this gives an easy way to skip them for local runs.

tests/examples/test_examples.py
```diff
@@ -0,0 +1,128 @@
...
+                    )
+                )
+
+            # Server is running, start the client
+            # Use a timeout of 10s for inference. Capture outputs to report
+            # them in case of failure.
+            try:
+                subprocess.run(
+                    [self.python, path.join(self.example_dir, "client.py")],
+                    check=True,
```
afrittoli, line 118: Simply check that the client returns a 0 exit code. 
More detailed checks on the grcp server API are already covered by the `runtime/*` tests.

hickeyma (changes requested):
Thanks for adding this @afrittoli, it was on my todo list. One small platform agnostic change. Also, when I run the tests locally 30 secs is timing out for me. When I bump it to 60 secs it works. Maybe consider changing the timeout value?

.github/workflows/build-library.yml
afrittoli: No, you're right, each one gets its own worker - I thought it would be useful to have dedicated ports, but I could remove that part, it's not likely to run them in parallel locally.

examples/text-sentiment/client.py
```diff
@@ -23,13 +26,14 @@
     ServicePackageFactory.ServiceType.INFERENCE,
 )
 
-port = 8085
-channel = grpc.insecure_channel(f"localhost:{port}")
+# Check if an alternate port is defined in RUNTIME_PORT
+port = os.getenv("RUNTIME_PORT", "8085")
```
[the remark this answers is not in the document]
afrittoli: I think I'll keep it simple drop the extra port settings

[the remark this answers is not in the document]
afrittoli: Or I can remove that too now, and when we add a second tests we make sure we don't run it in parallel.

tests/examples/shared.py
```diff
@@ -0,0 +1,75 @@
...
+    with tempfile.TemporaryDirectory() as venv_dir:
+        python = path.abspath(path.join(venv_dir, "bin", "python3"))
+        pip = path.abspath(path.join(venv_dir, "bin", "pip3"))
+        example_dir = path.abspath(path.join(caikit_dir, "examples", example))
+
+        # Create a venv, install local version of caikit and local version of requirements
+        venv.create(env_dir=venv_dir, system_site_packages=False, with_pip=True)
+        try:
+            subprocess.run([pip, "install", "-e", caikit_dir], check=True)
+            subprocess.run(
```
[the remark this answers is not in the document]
afrittoli: I agree about the API compatibility; however I wanted to test installing the requirements with the different versions of python as that may behave differently.

I was trying to reproduce as closely as possible the experience of an end user running the example

pyproject.toml
afrittoli: Do you usually run the test on laptops through `tox`?
I can make the tox job to not run the test by default and enable it in the github action only

tests/examples/test_examples.py
```diff
@@ -0,0 +1,80 @@
...
+
+    with requirements("text-sentiment") as (python_venv, example_dir):
+        # Client and server environment variables
+        env_vars = dict(
+            RUNTIME_PORT=str(grpc_port),
+            RUNTIME_METRICS_PORT=str(grpc_port + 1),
+        )
+        # Start the server
+        with subprocess.Popen(
+            [python_venv, "start_runtime.py"],
```
joerunde, line 46: If we refactor these two scripts (start_runtime.py and client.py) to have `main()` functions, we could invoke those mains here directly instead of using subprocesses. (The server can be run in a non-blocking mode so the caller doesn't need to create an extra thread or process for it to use)

This is probably not a big deal but I know I am always slightly annoyed when I have to debug something that went wrong in a subprocess vs. being able to spin up a debugger and walk down the full stack

.github/workflows/build-library.yml
afrittoli: Removed now

pyproject.toml
afrittoli: I made the test excluded by default via tox, it's only enabled in the GitHub workflow

joerunde: > Do you usually run the test on laptops through tox?

Yeah, either a simple `tox` for tests + formatting, or a `tox -e py` for just the tests

tests/examples/shared.py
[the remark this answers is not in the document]
joerunde: > installing the requirements with the different versions of python as that may behave differently

Ah, that is true :/

We can leave this in then, though I would like to talk more later about drawing a line somewhere. At some point this becomes more burden than it is useful: The examples are great to show how to use the caikit API but the actual content repos (the incoming caikit-nlp and caikit-code, names TBD) would cover testing that our dependency stacks are aligned and working for real-world models on all of our supported python versions.

tests/examples/test_examples.py
hickeyma: We need to test all code in the example, so need to use the client and server python scripts to be tested.

afrittoli:
> @afrittoli While testing the PR, I decided to break the API in the code to see how the test would handle this. The reason being is that is how the test will pickup API changes hopefully.
> 
> So I disabled the `task` in the `module` annotation i.e. disabled https://github.com/caikit/caikit/blob/main/examples/text-sentiment/text_sentiment/runtime_model/hf_module.py#L42
> 
> When run from CLI, the server will start with warning that no modules loaded and the client fails as follows:
> 
> ```shell
> $ python3 client.py
> 
> <function register_backend_type at 0x7fad78570ee0> is still in the BETA phase and subject to change!
> Traceback (most recent call last):
>   File "/Users/mhickey/Documents/Dev/github.com/caikit/caikit/examples/text-sentiment/client.py", line 39, in <module>
>     request = inference_service.messages.HuggingFaceSentimentTaskRequest(
> AttributeError: module 'ClientMessages' has no attribute 'HuggingFaceSentimentTaskRequest'
> ```
> 
> When run as a test, the test just hangs. It doesn't seem to propagate the error from the client.

Thanks @hickeyma - nice catch! 
The problem was calling `pytest.fail()` before terminating the server - it works now.

hickeyma (approved):
LGTM, thanks @afrittoli for adding the test and infra around it. This will be very useful for API changes.

hickeyma:
@joerunde This PR is looking good. Do you have any other concerns or are you ok with me merging it?

joerunde (approved):
took a whopping 72 seconds with my internet connection, but works on my machine!

Thanks for the detailed feedback @hickeyma and thanks for the test coverage @afrittoli

PR #202, opened 2023-06-08
Remove explict permissions from the build workflow

afrittoli:


joerunde (approved):
a last attempt but probably won't work

PR #203, opened 2023-06-08
Drop the coverage action

afrittoli:
The coverage action does not get the correct permissions in the GITHUB_TOKEN on PRs from non-maintainers, so dropping it for now.


joerunde (approved):
😭

joerunde:
oof, the 3.8 build hit a deadlock or failed to clean up a thread or something :/

Shouldn't be related to this though

PR #204, opened 2023-06-08
Format and lint examples

markstur:
This fixes the errors that will show up with tox after #196 is merged

The arguments differ problem is due to the way ModuleBase is defined.  I think disabling on each line is best fix so we still have that check elsewhere.  Is there a better fix?

Also I'm disabling the import error on transformers because I haven't found a convenient way to ensure we have the example requirements (and don't want to install transformers for now).

### DIFF
--- a/examples/text-sentiment/start_runtime.py
+++ b/examples/text-sentiment/start_runtime.py
@@ -20,7 +20,9 @@ import sys
 import alog
 
 # Local
+from caikit.runtime import grpc_server
 import caikit
+import text_sentiment  # pylint: disable=unused-import
 
 models_directory = path.abspath(path.join(path.dirname(__file__), "models"))
 caikit.config.configure(
@@ -34,12 +36,6 @@ sys.path.append(
     path.abspath(path.join(path.dirname(__file__), "../"))
 )  # Here we assume that `start_runtime` file is at the same level of the `text_sentiment` package
 
-# Local
-import text_sentiment
-
 alog.configure(default_level="debug")
-
-# Local
-from caikit.runtime import grpc_server
 
 grpc_server.main()
--- a/examples/text-sentiment/text_sentiment/runtime_model/hf_module.py
+++ b/examples/text-sentiment/text_sentiment/runtime_model/hf_module.py
@@ -16,7 +16,7 @@
 import os
 
 # Third Party
-from transformers import pipeline
+from transformers import pipeline  # pylint: disable=import-error
 
 # Local
 from caikit.core import ModuleBase, ModuleLoader, ModuleSaver, TaskBase, module, task
@@ -51,7 +51,9 @@ class HuggingFaceSentimentModule(ModuleBase):
         model = pipeline(model=config.hf_artifact_path, task="sentiment-analysis")
         self.sentiment_pipeline = model
 
-    def run(self, text_input: TextInput) -> ClassificationPrediction:
+    def run(  # pylint: disable=arguments-differ
+        self, text_input: TextInput
+    ) -> ClassificationPrediction:
         """Run HF sentiment analysis
         Args:
             text_input: TextInput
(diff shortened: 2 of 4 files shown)


hickeyma (dismissed):
@markstur thanks for pushing the changes. Would it be better to merge #196 into this PR and close #196?

examples/text-sentiment/start_runtime.py
```diff
@@ -19,16 +19,13 @@
...
 
+# Local
+from caikit.runtime import grpc_server
+
 sys.path.append(
     path.abspath(path.join(path.dirname(__file__), "../"))
 )  # Here we assume that `start_runtime` file is at the same level of the `text_sentiment` package
 
-# Local
-import text_sentiment
```
evaline-ju, line 27: We configure the library at import time, so while this looks unused, I would expect the removal here to break the example. We may want to add a `disable=unused-import` line here

markstur:
> @markstur thanks for pushing the changes. Would it be better to merge #196 into this PR and close #196?

It would be convenient, but I think it is best to have the CI things added in a separate commit without the clutter of a bunch of reformatting in code.  So it's chicken-vs-egg problem, but here I have done linting (trust me) and updated CI.  So I'd merge the fixes first and then do the merge that enables CI.

If you really want them in one PR you have to promise that they won't get squashed.  Just lemme know.

examples/text-sentiment/start_runtime.py
markstur: ouch thanks for catching that.

hickeyma:
> If you really want them in one PR you have to promise that they won't get squashed. Just lemme know.

Add them, to 1 PR. Its better self contained.

gabe-l-hart:
@markstur, is this worth rebasing and merging?

markstur:
Updating and rebasing for the request to combine the PRs.

examples/text-sentiment/text_sentiment/runtime_model/hf_module.py
```diff
@@ -16,7 +16,7 @@
 import os
 
 # Third Party
-from transformers import pipeline
+from transformers import pipeline  # pylint: disable=import-error
```
gabe-l-hart, line 19: I'm curious why we need to import this. Is there an import-time side effect that we need? If so, it would be good to have a comment to that effect

gabe-l-hart (approved):
Looks good to me! One question, but it will be nice to get these in.

Also, one other thing that would be good to add in here is a special `tox` run set that _does_ run the example tests. Currently [they're disabled](https://github.com/caikit/caikit/blob/main/tox.ini#L15C126-L15C143) in all of the test targets. That can be another PR, though.

gabe-l-hart:
@hickeyma Let us know if you have further change requirements, otherwise I'll dismiss your review and merge

examples/text-sentiment/text_sentiment/runtime_model/hf_module.py
markstur: It is imported because pipeline is used directly in the example init() [L51](https://github.com/caikit/caikit/pull/204/files#diff-3cc4fc99fd1cd93a8820c146bdcd7997619a6346e3a8df00df3d12f0b90b0a8cR51),  but since examples its own requirements.txt and we lint at the project level this is disabled.

markstur:
> @markstur thanks for pushing the changes. Would it be better to merge #196 into this PR and close #196?

done

markstur:
I think this is ready.  Does the one pylint disable need a comment?  I'm a bit reluctant to comment a comment.

examples/text-sentiment/text_sentiment/runtime_model/hf_module.py
gabe-l-hart: aaaah, sorry, I mixed this up with an unused import. Carry on!

gabe-l-hart:
Ha! Entirely fair. I'll go ahead with the merge

PR #205, opened 2023-06-08
Display OpenSSF Badge

Nhan-Hoang:
caikit has passed the OpenSSF basic level as `passing`. We want to display the badge in the README as standard practice.
Result from this [issue](https://github.com/caikit/caikit/issues/152)


gabe-l-hart (approved):
Let's go!

PR #207, opened 2023-06-08
:fire: delete unused compiled proto helpers

joerunde:
cleanup we missed when we deleted the compiled proto service support


gabe-l-hart (approved):
👋

PR #208, opened 2023-06-08
:sparkles: Add streaming task support

### LINKED ISSUE #101: Streaming output support

Text generation will require supporting modules that return `DataStream[T]` to stream output.
This will need to be hooked up to a unary-stream RPC.

Tasks can specify `DataStream[T]` as output type

Service generation code adds output streaming flag in py-to-proto

Servicer implementation returns something like...
```
stream = model.run
return iter(stream.map(stream_class.to_proto()))
```

## Additional context

joerunde:
Closes #101 

This was remarkably simple after all the 🌶️  refactoring that we did in the service generation / implementation packages


caikit/core/task.py
```diff
@@ -194,3 +238,8 @@ def decorator(cls: Type[TaskBase]) -> Type[TaskBase]:
         return cls
 
     return decorator
+
+
+def _is_iterable_type(type_: Type) -> bool:
+    """Checks if a type is iterable"""
+    return isinstance(type_, collections.abc.Iterable)
```
joerunde, line 245: there are a lot of checks in `isa.py` but I can't import that here or it'll cycle

caikit/runtime/interceptors/caikit_runtime_server_wrapper.py
```diff
@@ -243,18 +248,22 @@ def _make_new_handler(
         replace_with_global_predict: bool = True,
     ):
         if original_rpc_handler.unary_unary:
+            if replace_with_global_predict:
+                behavior = self.safe_rpc_wrapper(self._global_predict)
+            else:
+                behavior = self.safe_rpc_wrapper(original_rpc_handler.unary_unary)
```
joerunde, line 254: I un-inlined this for both easier debugging and better code coverage checks

tests/conftest.py
```diff
@@ -138,17 +138,21 @@ def runtime_grpc_server(
         training_service=sample_train_service,
     ) as server:
 
-        grpc_thread = threading.Thread(
-            target=server.start,
-        )
-        grpc_thread.daemon = False
-        grpc_thread.start()
-        _check_server_readiness(server)
```
joerunde, line 146: This was a fun bug, the context manager already boots the server so this was just spinning up a thread that would immediately crash by trying to boot the server again

tests/conftest.py
```diff
@@ -208,6 +203,22 @@ def loaded_model_id(good_model_path) -> str:
     model_manager.unload_model(model_id)
 
 
+@pytest.fixture
+def loaded_streaming_model_id(streaming_model_path) -> str:
```
evaline-ju, line 207: like #213 change can we also rename to `sample_task_streaming....`

tests/fixtures/sample_lib/modules/sample_task/streaming_module.py
```diff
@@ -0,0 +1,50 @@
+"""
+A sample module for sample things!
```
evaline-ju, line 2: maybe to distinguish from the other module? could probably be more nuanced
```suggestion
A streaming module for streaming things!
```

tests/fixtures/sample_lib/modules/sample_task/streaming_module.py
```diff
@@ -0,0 +1,50 @@
...
+        config = loader.config
+        return cls(config["stream_size"])
+
+    def run(self, sample_input: SampleInputType) -> DataStream[SampleOutputType]:
+        """
+        Args:
+            sample_input (sample_lib.data_model.SampleInputType): the input
+
+        Returns:
+            sample_lib.data_model.SampleOutputType: The output
```
evaline-ju, line 34: nit: Datastream

tests/runtime/test_grpc_server.py
```diff
@@ -195,6 +195,25 @@ def test_predict_fake_module_ok_response(
     assert actual_response == HAPPY_PATH_RESPONSE
 
 
+def test_predict_streaming_module(
+    loaded_streaming_model_id, runtime_grpc_server, sample_inference_service
+):
+    """Test RPC CaikitRuntime.WidgetPredict successful response"""
```
evaline-ju, line 202: WidgetPredict does not seem appropriate here?

tests/runtime/test_grpc_server.py
```diff
@@ -195,6 +195,25 @@ def test_predict_fake_module_ok_response(
...
+    )
+    stream = stub.StreamingTaskPredict(
+        predict_request, metadata=[("mm-model-id", loaded_streaming_model_id)]
+    )
+
+    count = 0
+    for response in stream:
+        assert response == HAPPY_PATH_RESPONSE
+        count += 1
+    assert count == 10
```
evaline-ju, line 215: This is a bit non-obvious when hidden in the module config, could be part of the module itself as default?

tests/core/test_task.py
```diff
@@ -51,6 +51,43 @@ class SampleTask(TaskBase):
             pass
 
 
+def test_task_decorator_can_have_iterable_output():
```
evaline-ju, line 54: maybe two test cases because this looks like a "happy path" test but also tests with the type error?

caikit/runtime/interceptors/caikit_runtime_server_wrapper.py
```diff
@@ -192,8 +192,13 @@ def __init__(self, method):
                     original_rpc_handler = handler.service(DummyHandlerCallDetails(fqm))
 
                     # Make sure this is a unary-unary RPC
```
evaline-ju, line 194: nit: comment needs update?

caikit/core/task.py
```diff
@@ -112,6 +104,47 @@ def get_output_type(cls) -> Type[DataBase]:
...
+                    # Something in the union has an acceptable type
+                    return
+
+        # Iterable output checks:
+        # `output_type` here can be any iterable, so long as it is type annotated with the type
+        # that is iterated over. So we first only check that it is iterable at all, and then look
+        # at the annotated type's args.
+        if isiterable(output_type) and isiterable(cls.get_output_type()):
+            # in this case we validated up-front that cls.output_type is Iterable[T]
+            streaming_type = typing.get_args(cls.get_output_type())[0]
```
gabe-l-hart, line 125: This is error prone if a user uses `List` (with no args) as the type hint

```suggestion
        # at the annotated type's args.
        task_output_type = cls.get_output_type()
        task_output_type_args = get_args(task_output_type)
        if isiterable(output_type) and isiterable(task_output_type) and task_output_type_args:
            # in this case we validated up-front that cls.output_type is Iterable[T]
            streaming_type = task_output_type_args[0]
```

caikit/runtime/interceptors/caikit_runtime_server_wrapper.py
```diff
@@ -191,9 +192,14 @@ def __init__(self, method):
                     # Get the original grpc.RpcMethodHandler for this RPC method
                     original_rpc_handler = handler.service(DummyHandlerCallDetails(fqm))
 
-                    # Make sure this is a unary-unary RPC
-                    if not original_rpc_handler.unary_unary:
-                        raise NotImplementedError("Unary-unary RPCs only!")
+                    # Make sure this is a supported RPC flavor
```
gabe-l-hart, line 195: This line is pretty confusing to me as I'd expect the logic to be exactly the opposite of what you have written, but I've seen it working in action, so I'll trust I'm just misreading it!

caikit/runtime/service_generation/rpcs.py
```diff
@@ -95,12 +96,22 @@ def create_request_data_model(self, package_name: str) -> Type[DataBase]:
 
     def create_rpc_json(self, package_name: str) -> Dict:
         """Return json snippet for the service definition of this RPC"""
-        output_type_name = self.return_type.get_proto_class().DESCRIPTOR.full_name
+        if isiterable(self.return_type):
```
gabe-l-hart, line 99: Looks like you're using this again for type analysis rather than instance analysis. I don't think we actually used this at all in `caikit` proper, but some of our users might have been using to check instances, so we should make sure it wouldn't be better to do something like `isiterable_type` as a separate check.

caikit/runtime/servicers/global_predict_servicer.py
```diff
@@ -137,10 +137,10 @@ def Predict(self, request, context):
             response (object):
                 A Caikit Library data model response object
         """
-        desc_name = request.DESCRIPTOR.name
-        outer_scope_name = "GlobalPredictServicer.Predict:%s" % desc_name
+        request_name = request.DESCRIPTOR.name
+        outer_scope_name = "GlobalPredictServicer.Predict:%s" % request_name
```
gabe-l-hart, line 141: Funny that for these scopes we're using `%` encoding! Nothing to change (and I have plans to move these lines around a bit, so no need to mess with them), but some day we should actually get explicit about when to use `f"{}"`, `"{}".format(...)`, and `"%s" % ...` style string interpolation (and of course lazy log interpolation with `"%s", ...`!)

caikit/runtime/utils/servicer_util.py
```diff
@@ -26,12 +26,14 @@
...
 
 # Local
+from caikit.core.data_model import DataStream
 from caikit.core.data_model.base import DataBase
 from caikit.core.signature_parsing import CaikitMethodSignature
 from caikit.interfaces.runtime.data_model.training_management import ModelPointer
 from caikit.runtime.model_management.model_manager import ModelManager
 from caikit.runtime.types.caikit_runtime_exception import CaikitRuntimeException
 from caikit.runtime.utils.import_util import get_data_model
+import caikit.core
```
gabe-l-hart, line 36: Is this pulled in just for `isiterable`? Seems a bit overkill when you could pull it in directly from `caikit.core.toolkit.isa`, but 🤷

caikit/runtime/utils/servicer_util.py
```diff
@@ -65,8 +67,19 @@ def validate_caikit_library_class_method_exists(caikit_library_class, method_nam
...
 
 
-def build_proto_response(caikit_library_response):
+def build_proto_response(
+    caikit_library_response: Union[DataBase, Iterable[DataBase]]
+) -> Union[ProtoMessageType, Iterator[ProtoMessageType]]:
+    """Either serializes a data model instance into a protobuf message, or returns a lazy iterator
+    that serializes each data model from an iterable of data models"""
     try:
+        if caikit.core.isiterable(caikit_library_response):
```
gabe-l-hart, line 76: Here we're using it to check an instance!

caikit/runtime/utils/servicer_util.py
```diff
@@ -65,8 +67,19 @@ def validate_caikit_library_class_method_exists(caikit_library_class, method_nam
...
+            # probably a streaming rpc!
```
gabe-l-hart, line 77: Can we be more certain than `probably` here? Seems like this information is known explicitly when the handler gets bound

caikit/core/toolkit/isa.py
```diff
@@ -61,7 +58,11 @@ def isprimitive(obj):
 
 def isiterable(obj):
     """Returns True if obj can be iterated over, otherwise False."""
-    return isinstance(obj, collections.abc.Iterable)
+    try:
+        iter(obj)
```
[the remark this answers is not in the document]
joerunde: There are better checks that are available in 3.11, but not in 3.9 and this one seems simple enough. It's a bit janky but for the contexts we use it in, `list` wouldn't be something that we'd want to consider iterable

caikit/core/task.py
```diff
@@ -112,6 +104,47 @@ def get_output_type(cls) -> Type[DataBase]:
...
+            f"Found {output_type} but expected {cls.get_output_type()}"
+        )
+
+    @staticmethod
+    def _subclass_check(this_type, that_type):
+        """Wrapper around issubclass that first checks if both args are classes.
+        Returns True if the types are the same, or they are both classes and this_type
+        is a subclass of that_type
+        """
+        if this_type == that_type:
```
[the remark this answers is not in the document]
joerunde: unclear!

You're right that both this isn't a correct general-purpose implementation, and it should only be called on types from this package. I'm inclined to leave as is because we don't need a generally correct implementation?

caikit/runtime/interceptors/caikit_runtime_server_wrapper.py
joerunde: 🤔 
It raises if neither `unary_unary` nor `unary_stream` are set, and those are the only two flavors we support. This should be equivalent to
```
if original_rpc_handler.stream_unary or original_rpc_handler.stream_stream:
```

caikit/runtime/service_generation/rpcs.py
joerunde: Yeah, I could be convinced to split that into a separate function

caikit/runtime/utils/servicer_util.py
joerunde: yeah, I'll write something better

caikit/core/task.py
gabe-l-hart: Ah! I misunderstood that comment. Looks reasonable as is then

caikit/runtime/interceptors/caikit_runtime_server_wrapper.py
gabe-l-hart: 🤦 I'm... confused. Carry on!

joerunde:
@gabe-l-hart I've gone and deleted `isa.py` and stored the streaming-ness of the task on the class itself so we can query it instead of guessing later based on the shape of the output

caikit/core/toolkit/isa.py
gabe-l-hart, line 1: Bye!!!

gabe-l-hart (approved):
Ship it!

PR #211, opened 2023-06-09
Create docstring converter

### LINKED ISSUE #180: Migrate from GabeStyle to Google Style docstrings

willmj:
Resolves #180 

dsconverter.py converts custom "Gabe" style docstrings into Google style docstrings. This will allow us to have more functionality with tools like Sphinx which autofills caikit's readthedocs. Script takes a file or directory name and changes all custom style docstrings of either that file or all .py files within the file to Google format. I also added a testenv to tox.ini which can be run with 'tox -e dsconverter'. This command will change all the docstrings in the caikit directory to Google style.

### DIFF
--- a/caikit/config/config.py
+++ b/caikit/config/config.py
@@ -62,8 +62,8 @@ def configure(
             with overrides for your usage.
         config_dict (Optional[Dict]): Config overrides in dictionary form
 
-    Returns: None
-        This only sets the config object that is returned by `caikit.get_config()`
+    Returns: None: This only sets the config object that is returned by
+        `caikit.get_config()`
     """
     if not config_yml_path and not config_dict:
         log.error("<CFG43273054E>", "No config_file or config_dict provided")
@@ -129,12 +129,10 @@ def merge_configs(
     and the type of the key for both is a dict, recursively merge, otherwise
     set the base value to the override value.
     Args:
-        base: Optional[dict]
-            The base config that will be updated with the overrides
-        overrides: Optional[dict]
-            The override config
-        merge_strategy: str
-            The merging strategy, either `merge` or `override`
+        base (Optional[dict]): The base config that will be updated with the
+            overrides
+        overrides (Optional[dict]): The override config
+        merge_strategy (str): The merging strategy, either `merge` or `override`
             `override` will replace values in base with those from overrides
             `merge` will deep-merge dictionaries and prepend-merge lists
     Returns:
--- a/caikit/core/data_model/base.py
+++ b/caikit/core/data_model/base.py
@@ -490,8 +490,8 @@ class DataBase(metaclass=_DataBaseMetaClass):
         message or a repeated message
 
         Args:
-            field_name:  str
-                Field name to check (AttributeError raised if name is invalid)
+            field_name (str): Field name to check (AttributeError raised if name
+                is invalid)
 
         Returns:
             data_model_type:  Type[DataBase]
@@ -610,7 +610,6 @@ class DataBase(metaclass=_DataBaseMetaClass):
 
         Args:
             buf: The binary buffer containing a serialized protobufs message
-
         Returns:
             A data model object instantiated from the protobufs message deserialized out of `buf`
         """
@@ -624,12 +623,9 @@ class DataBase(metaclass=_DataBaseMetaClass):
         """Build a DataBase from protobufs.
 
         Args:
-            proto:
-                A protocol buffer to serialize from.
-
+            proto: A protocol buffer to serialize from.
         Returns:
-            protobufs
-                A DataBase object.
+            protobufs: A DataBase object.
         """
         error.type_check("<COR45207671E>", ProtoMessageType, proto=proto)
         if cls._proto_class.DESCRIPTOR.name != proto.DESCRIPTOR.name:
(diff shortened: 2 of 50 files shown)


WillMJ:
Added the script and updated docstrings in two seperate (most recent) pushes. Had to manually change the following docstrings:
 - required_parameters on line 206 of caikit/core/task.py
 - *args, **kwargs and block_level on line 461 of caikit/core/modules/base.py
 - Example Usages section in caikit/core/toolkit/wip_decorator
 - def resolve_and_load in caikit/core/model_manager.py
 These all did not fit the formatting correctly and caused lint to fail. However, the script works in most cases.

caikit/core/data_model/streams/data_stream.py
```diff
@@ -257,20 +256,17 @@ def from_csv(cls, filename: str, *args, skip=0, **kwargs) -> "DataStream[List]":
...
-                skipping a header line.
-            args, kwargs:
-                Additional arguments passed to the `csv.reader` function.  These can be used to
-                specify the delimiter or other csv settings.
+            filename (str): A path to a csv file that has rows corresponding to
+                data items and columns corresponding to the elements of each
+                data item.
+            skip (int): Number of lines to skip at the beginning of the csv
+                file.  This is often useful for skipping a header line.
+            args, kwargs (Additional arguments passed to the `csv.reader` function.  These can be
```
anhuong, line 264: Edge case - no data type provided with colon

caikit/core/data_model/streams/data_stream.py
```diff
@@ -257,20 +256,17 @@ def from_csv(cls, filename: str, *args, skip=0, **kwargs) -> "DataStream[List]":
...
+                used to): specify the delimiter or other csv settings.
         Returns:
-            DataStream
-                A data stream that produces a data item for each line of the csv file and where each
-                element of the data item corresponds to a column in the csv file.
-        Examples:
+            DataStream: A data stream that produces a data item for each line of
+                the csv file and where each element of the data item corresponds
+                to a column in the csv file.Examples:
```
anhuong, line 269: Edge case: Examples added to description

caikit/config/config.py
```diff
@@ -47,23 +47,24 @@ def configure(
...
+           5. Environment variables, in ALL_CAPS_SNAKE_FORMAT
+
+       Args:
+           config_yml_path (Optional[str]): The path to the base configuration yaml
+               with overrides for your usage.
+           config_dict (Optional[Dict]): Config overrides in dictionary form
+
+       Returns:
+    None: This only sets the config object that is returned by
+           `caikit.get_config()`
```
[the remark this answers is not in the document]
WillMJ: Fixed (partially). Now docstring converter will turn this into.
```
Returns: None: This only sets the config object that is returned by
           `caikit.get_config()`
```

caikit/core/data_model/base.py
```diff
@@ -570,8 +570,7 @@ def from_binary_buffer(cls, buf):
         """Builds the data model object out of the binary string
 
         Args:
-            buf: The binary buffer containing a serialized protobufs message
-
+            buf (The binary buffer containing a serialized protobufs message)
```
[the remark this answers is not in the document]
WillMJ: Fixed. Docstring converted will now leave this unmodified.

caikit/core/data_model/streams/data_stream.py
WillMJ: Fixed. Docstring converted will now pull the description up but not put parentheses around it. However, if the first line of description is too long you may have to manually format it to be within 100 characters.

joerunde:
@WillMJ I am excited to see some better sphinx links after this, thanks!

WillMJ:
Got the py38 test to pass by committing again.

scripts/test_dsconverter.py
```diff
@@ -0,0 +1,432 @@
...
+# Third Party
+# Import the script code
+from dsconverter import CustomDocstringConverter
+import pytest
+
+
+# Test case for convert_to_google_style function
+def test_get_docstrings():
+    converter = CustomDocstringConverter()
+    test_path = "/Users/mwjohnson/Code/caikit/caikit/core/modules/config.py"
```
joerunde, line 24: nit: This is an absolute path that will only work on your machine

scripts/test_dsconverter.py
```diff
@@ -0,0 +1,432 @@
...
+        Returns:
+            list
+                List of labels in the format of the module_type that is being called.
+        """
+    '''
+
+    converted = converter.convert_to_google_style(test1)
+    print("Pre-conversion:\n", test1, "\nPost-conversion:\n", converted)
+
+    assert 1 == 2
```
joerunde, line 56: So I like that there are "tests" here, though it looks like they all just fail and print out the converted strings so that a human can look through and approve them. I don't think that's quite at the level of polish that we would want to include in caikit (though maybe it's not the end of the world since this isn't in the main unit test suite).

This style of testing could be easily handled with data-driven testing, if you wanted to dig into that a bit. Basically you don't have to write tests for each of these cases, you can simply store some gabe-style docstrings and their expected conversions, and write one test that makes sure the conversions are all correct.

joerunde (changes requested):
Nice!
The only thing I think we'd want to change here is to have functioning tests if we're gonna commit those. I think it should be a relatively painless exercise to make that change, if you're up for it @WillMJ

scripts/test_dsconverter.py
WillMJ: Good point. I have added functionality to the test cases so they actually work as tests now.

joerunde (approved):

PR #213, opened 2023-06-09
:sparkles: Add rpc validation on inference

### LINKED ISSUE #215: Validate correct RPCs are invoked on models

I am very frustrated when I call the wrong RPC for a model, and get a serialization error back

When I call the wrong task rpc for a model, I should be told which RPC I need to invoke instead 

## Additional context

We had previously implemented this by trying to check the model output's type name against the rpc name and doing some fuzzy matching to see if they were the same. This had the downside of not being very robust, and making you wait for the model to run before erroring.

Now that we have an internal representation of our RPCs and the modules they support, we can check the model against the rpc up-front, and return an error that points to the correct RPC if the wrong one is given

joerunde:
This PR adds some more robust checks that the module loaded for inference is supported by the inference rpc invoked.

This should actually be helpful for users, so that they don't just get serialization errors back when they invoke the wrong rpc for a model. (Which I just did and floundered around for a while wondering what was wrong).

Closes #215

### DIFF
--- a/caikit/runtime/service_factory.py
+++ b/caikit/runtime/service_factory.py
@@ -36,7 +36,7 @@ from caikit.interfaces.runtime.data_model import (
     TrainingInfoResponse,
 )
 from caikit.runtime import service_generation
-from caikit.runtime.service_generation.rpcs import snake_to_upper_camel
+from caikit.runtime.service_generation.rpcs import CaikitRPCBase, snake_to_upper_camel
 from caikit.runtime.utils import import_util
 import caikit.core
 
@@ -73,6 +73,7 @@ class ServicePackage:
     ]
     stub_class: Type
     messages: ModuleType
+    caikit_rpcs: Set[CaikitRPCBase]
 
 
 class ServicePackageFactory:
@@ -110,6 +111,7 @@ class ServicePackageFactory:
                 registration_function=grpc_service.registration_function,
                 stub_class=grpc_service.client_stub_class,
                 messages=None,  # we don't need messages here
+                caikit_rpcs=set(),  # No caikit RPCs
             )
 
         # First make sure we import the data model for the correct library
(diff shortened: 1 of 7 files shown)


caikit/runtime/service_factory.py
```diff
@@ -160,6 +162,7 @@ def get_service_package(cls, service_type: ServiceType) -> ServicePackage:
             registration_function=grpc_service.registration_function,
             stub_class=grpc_service.client_stub_class,
             messages=client_module,
+            caikit_rpcs=set(task_rpc_list),
```
evaline-ju, line 165: nit: `task_rpc..` here seems confusing if valid for both inference and training

joerunde: oops, I had made that change in my other in-flight PR, you good with merging this and I'll resolve the conflict on that one?

evaline-ju: that is acceptable

tests/runtime/test_grpc_server.py
```diff
@@ -212,6 +213,55 @@ def test_predict_fake_module_error_response(
...
+    sample_task_model_id, runtime_grpc_server, sample_inference_service
+):
+    """Check that the server catches models sent to the wrong task RPCs"""
+    stub = sample_inference_service.stub_class(runtime_grpc_server.make_local_channel())
+    predict_request = sample_inference_service.messages.OtherTaskRequest(
+        sample_inputsampleinputtype=HAPPY_PATH_INPUT
+    )
+    with pytest.raises(grpc.RpcError) as context:
+        stub.OtherTaskPredict(
+            predict_request, metadata=[("mm-model-id", sample_task_model_id)]
```
evaline-ju, line 226: offline discussed changing the fixture to make it more obvious why the RPC here is the wrong RPC

evaline-ju (approved):
LGTM!

PR #214, opened 2023-06-09
:bug: Fix incorrect cancellation of training request for thread executor

gkumbhat:
### Changes
- Currently the local thread executor is incorrectly calling `cancel` function when getting to `__del__` function at the end. This will cause false alarms where training gets successfully completed but at the end we would see `Terminated` error message, which can cause confusion.
- Add fix to overcome python3.9-3.11 bug where if you create a subprocess inside a ThreadPoolExecutor then it will always exit with `exitcode` 1 thus signaling error.


gkumbhat:
Potentially more changes coming so moving this to draft for now

gabe-l-hart (approved):
Wow, nice find!

PR #221, opened 2023-06-12
💥  API changes for py-to-proto 0.3.x

prashantgupta24:
⬆️ bump `py-to-proto` to 0.3.x

This adds an underscore between the field name and the type for generated oneof field names


joerunde (approved):
@prashantgupta24 I went ahead and changed the PR title to reflect that there are API changes, but this looks good 👍

PR #231, opened 2023-06-13
updated config file to fix hf model loading

simona-rc:


simona-rc:
I now see that if I change directory to caikit/examples/text-sentiment, then we don't need this fix. 
So no need to merge this change and we can close it

gabe-l-hart:
@simona-rc For the record (and for anyone else that encounters this issue) you can also override this config value (and any other) from the environment when you run any program that uses `caikit` (including launching the `caikit.runtime.grpc_server`). The pattern for environmental overrides is to join the nested keys with `_` and make the string uppercase. For example, you want to override `training.local_models_dir`, so you would `export TRAINING_LOCAL_MODELS_DIR=/path/to/my/models`.

Issue #198, opened 2023-06-08
Union of lists doesn't work


prashantgupta24:
## Describe the bug

A function argument (for either `run` or `train`) like `Union[List[str], List[int]]` doesn't work with the `_dataobject_to_proto` at the moment.

According to https://protobuf.dev/programming-guides/proto3/#using-oneof,
```
You then add your oneof fields to the oneof definition. You can add fields of any type, except map fields and repeated fields.
```
we can't add repeated fields in `oneof`s.

## Platform

Please provide details about the environment you are using, including the following:

- Interpreter version: 3.9
- Library version: 0.7

## Sample Code

This errors out:

```
    @dataobject
    class Foo(DataObjectBase):
        foo: Union[List[str], List[int]]
```

## Expected behavior

The above code should run and create a `oneof`, something like this?
```
message ThingWithRepeatedOneof {
  message IntSequence {
    repeated int32 values = 1;
  }
  message StrSequence {
    repeated string values = 1;
  }

  oneof value_sequence {
    IntSequence int_sequence = 1;
    StrSequence str_sequence = 2;
  }
}
```

## Observed behavior

```
    def test_dataobject_union_repeated():
        """Make sure that a JsonDict type is handled correctly in a dataobject"""
    
        @dataobject
>       class Foo(DataObjectBase):

tests/core/data_model/test_dataobject.py:820: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
caikit/core/data_model/dataobject.py:197: in dataobject
    return decorator(args[0])
caikit/core/data_model/dataobject.py:163: in decorator
    descriptor = _dataobject_to_proto(dataclass_=cls, **kwargs)
caikit/core/data_model/dataobject.py:226: in _dataobject_to_proto
    return _DataobjectConverter(*args, **kwargs).descriptor
.tox/3.9/lib/python3.9/site-packages/py_to_proto/dataclass_to_proto.py:118: in __init__
    super().__init__(
[...]

PR #232, opened 2023-06-13
♻️ Union list support

prashantgupta24:
Add example to support simple lists and lists in Union!

Heavy lifting done by this PR: https://github.com/IBM/py-to-proto/pull/54

fix https://github.com/caikit/caikit/issues/198

(also please squash commits, unable to git squash when there's a `merge main` commit)


caikit/runtime/utils/servicer_util.py
```diff
@@ -47,8 +47,15 @@ def validate_caikit_library_class_exists(cdm, class_name):
         return getattr(cdm, class_name)
 
     except AttributeError as e:
-        log.error("<RUN24024010E>", "No Caikit Library CDM class for %s", class_name)
-        raise e
+        log.warning("<RUN24024050W>", "No Caikit Library CDM class for %s", class_name)
```
prashantgupta24, line 50: Based on internal discussions, I'm also okay in removing the whole `validate_data_model` function altogether!

joerunde: Here's a story: https://github.com/caikit/caikit/issues/247

tests/core/data_model/test_dataobject.py
```diff
@@ -878,6 +878,51 @@ class Derived(Base):
...
+
+    #     @dataobject
+    #     class StrSequence(DataObjectBase):
+    #         values: List[str]
+
+    #     foo: Union[IntSequence, StrSequence]
+
+    # proto round trip
+    foo_int = Foo.IntSequence(values=[1, 2])
+    foo1 = Foo(foo=foo_int)
```
joerunde, line 905: Is this the only way these can be constructed?

It's good, but I think a native python @dataclass user would expect to be able to write
```
foo_with_ints = Foo([1, 2, 3])
foo_with_strs = Foo(["1", "2", "3"])
```

tests/runtime/utils/test_servicer_util.py
```diff
@@ -68,7 +68,7 @@ def test_servicer_util_validate_caikit_library_class_exists_returns_caikit_class
 
 def test_servicer_util_validate_caikit_library_class_exists_raises_for_garbage_input():
     """Test that validate_caikit_library_class_exists raises for garbage 'model'"""
-    with pytest.raises(AttributeError):
+    with pytest.raises(ValueError):
```
joerunde, line 71: I don't understand how your changes made this now be a value error? 🤷 

But I'm okay with it if we're discussing removing these checks anyway

tests/fixtures/sample_lib/modules/other_task/other_implementation.py
```diff
@@ -51,11 +51,13 @@ def train(
         cls,
         training_data: DataStream[int],
         sample_input: Union[SampleInputType, str],
+        list_value: Union[List[str], List[int]],
```
joerunde, line 54: I like the completeness of testing, but I do worry that we don't have much organization of these test fixtures.
There is a "primitive party implementation" test module, should we put all of these python primitive cases there instead?

joerunde (changes requested):
I probably missed the design discussion around how to instantiate these things in python, did we decide to not support simple lists or is that just not in a test? (Or did I just miss it?)

joerunde:
Yeah I don't know about the git history problem, but I'm fine if we just use the squash merge button when merging if you want

tests/fixtures/sample_lib/modules/other_task/other_implementation.py
prashantgupta24: Can we open a separate issue on organization of test fixtures and tackle it there?

joerunde: > Actually let me change that in this PR itself

Yup, I'd rather do it this way. Thanks!

Should this PR wait to merge until the dict PR merges?

prashantgupta24:
> did we decide to not support simple lists or is that just not in a test? (Or did I just miss it?)

I can add a test for that, I'm assuming the only thing missing is a test for it

tests/core/data_model/test_dataobject.py
prashantgupta24: Hmm, yeah that would be ideal, but can that be a separate feature request? This PR is mostly to handle the lists within function arguments, I believe in order to support the above, there _might_ have to be metaclass changes? Not sure

caikit/runtime/utils/servicer_util.py
prashantgupta24: Will handle in subsequent PR

prashantgupta24:
Hey @joerunde, update on this PR:

- added the test for a simple list
- moved test to `primitive_party`
- separate issue for removing validation of data model
- separate issue for python style initialization of lists within unions

joerunde (approved):
lpgtm!

PR #234, opened 2023-06-13
Allow task argument subclasses

gabe-l-hart:
## Description

This builds on #233 to allow module implementations to use subclassed data model types to implement tasks. Whether or not this is a good idea is definitely up for discussion!


gabe-l-hart:
We decided we don't want this functionality for now

PR #238, opened 2023-06-14
Split global predict servicer

### LINKED ISSUE #223: Refactor: Extract protocol logic from global predict servicer

For REST support, we're building a native rest server that would also like to use the `GlobalPredictServicer` to run model inference, instead of rewriting another inference handler.

But, the global predict handler currently assumes the request object is a `ProtobufMessage` and assumes that it should also serialize responses as `ProtobufMessage(s)`.

The protocol-specific (de)serialization logic should be extracted from the global predict servicer so that it can instead accept something more along the lines of a model id and a dictionary of kwargs to pass the model, and output the response of the model directly. This will allow us to plug in a REST translator alongside the gRPC translator so that we can run both servers using the same inference handler.

Copy-pasting code into a REST specific handler, but we're on team re-use over here

gabe-l-hart:
## Description

Closes #223 

This PR splits out the core inferencing logic from the `GlobalPredictServicer`'s `Predict` function so that it can be called directly with the data model objects. This is needed to allow a REST server to reuse the same inferencing logic as the grpc server.

### DIFF
--- a/caikit/runtime/servicers/global_predict_servicer.py
+++ b/caikit/runtime/servicers/global_predict_servicer.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # Standard
+from contextlib import contextmanager
 from importlib.metadata import version
 from typing import Set
 import traceback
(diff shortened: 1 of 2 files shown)


caikit/runtime/servicers/global_predict_servicer.py
```diff
@@ -188,20 +170,54 @@ def Predict(self, request, context):
                         response_proto = build_proto_stream(response)
                     else:
                         response_proto = build_proto_response(response)
+                return response_proto
+
+    def predict_model(self, request_name: str, model_id: str, context=None, **kwargs):
+        """Run a prediction against the given model. This does not require that"""
```
joerunde, line 176: This does not require that...?

tests/runtime/servicers/test_global_predict_servicer_impl.py
```diff
@@ -89,6 +86,20 @@ def test_global_predict_works_on_good_inputs(
...
+def test_global_predict_predict_model_direct(
+    sample_inference_service, sample_predict_servicer, sample_task_model_id
+):
+    """Test that the direct predict_model function can be called to bypass the
+    gRPC handler
+    """
+    response = sample_predict_servicer.predict_model(
+        request_name="SampleTaskRequest",
+        model_id=sample_task_model_id,
+        sample_input=HAPPY_PATH_INPUT,
```
joerunde, line 98: something looks funky here, this `HAPPY_PATH_INPUT` is a proto message, not a data model object?

caikit/runtime/servicers/global_predict_servicer.py
```diff
@@ -188,20 +170,54 @@ def Predict(self, request, context):
...
+    @contextmanager
+    def _handle_predict_exceptions(self, model_id: str, request_name: str):
+        try:
+            yield
 
         except CaikitRuntimeException as e:
             log_dict = {
                 "log_code": "<RUN50530380W>",
                 "message": e.message,
-                "model_id": get_metadata(context, self.MODEL_MESH_MODEL_ID_KEY),
```
joerunde, line 204: that's nice

tests/runtime/servicers/test_global_predict_servicer_impl.py
joerunde: Oh, we just never verify in the module that the input is a SampleInputType type, whoops!

joerunde (approved):

PR #240, opened 2023-06-14
Support Dicts in Service Generation

### LINKED ISSUE #133: Supporting Dicts with data model serialization
## Description

As a caikit library developer, I want to be able to support Dicts in function signatures (.train, .run) so that I can serve my library that may have Dict method parameters.

## Acceptance Criteria

- [ ] Support Dicts in service generation
- [ ] Unit tests

tharapalanivel:
## Description

Closes #133 

Supporting input params in function signatures of types simple dict and `JsonDict` for service generation.

### DIFF
--- a/tests/runtime/service_generation/test_protoable.py
+++ b/tests/runtime/service_generation/test_protoable.py
@@ -132,3 +132,12 @@ def test_to_protoable_signature_unsupported_dict():
         )
         == {}
     )
+
+
+def test_to_protoable_signature_dict_incomplete_type_hint():
+    assert (
+        to_protoable_signature(
+            signature={"name": dict},
+        )
+        == {}
+    )
--- a/tests/core/data_model/test_dataobject.py
+++ b/tests/core/data_model/test_dataobject.py
@@ -17,7 +17,7 @@
 # Standard
 from dataclasses import dataclass, field, is_dataclass
 from enum import Enum
-from typing import List, Optional, Union
+from typing import Dict, List, Optional, Union
 import copy
 import json
 import os
@@ -231,6 +231,28 @@ def test_dataobject_arrays():
 
     assert check_field_type(Foo.get_proto_class(), "bar", "TYPE_STRING")
     assert check_field_label(Foo.get_proto_class(), "bar", "LABEL_REPEATED")
+
+
+def test_dataobject_simple_dict():
+    """Make sure simple dicts work as expected"""
+
+    @dataobject
+    class Foo(DataObjectBase):
+        bar: Dict[str, int]
+
+    assert check_field_type(Foo.get_proto_class(), "bar", "TYPE_MESSAGE")
+    assert (
+        Foo.get_proto_class()
+        .DESCRIPTOR.fields_by_name["bar"]
+        .message_type.GetOptions()
+        .map_entry
+    )
+
+    dict_input = {"foo": 1, "bar": 2}
+    foo = Foo(bar=dict_input)
+    assert foo.bar == dict_input
+
+    assert Foo.from_proto(foo.to_proto()).bar == dict_input
 
 
 def test_dataobject_obj_refs_no_opt_types():


tests/runtime/test_grpc_server.py
```diff
@@ -365,6 +365,10 @@ def test_train_fake_module_does_not_change_another_instance_model_of_block(
 
     # Train an OtherModule with batch size 100
     stream_type = caikit.interfaces.common.data_model.DataStreamSourceInt
+    training_parameters_json_dict = caikit.core.data_model.json_dict.dict_to_struct(
```
prashantgupta24, line 368: can this `caikit.core.data_model.json_dict.dict_to_struct` conversion not happen in the background if the user passes in a dict? 🤔

caikit/runtime/utils/servicer_util.py
```diff
@@ -205,6 +208,17 @@ def validate_data_model(
             # and check to make that it is either a primitive protobufs type or that
             # we have a data model class that we can deserialize the protobufs with
             if not is_protobuf_primitive_field(field):
+
+                if field.message_type and field.message_type.GetOptions().map_entry:
```
prashantgupta24, line 212: I guess the same question here as well - whether we want to remove `validate_data_model` altogether?

caikit/runtime/service_generation/protoable.py
```diff
@@ -124,25 +124,33 @@ def is_protoable_type(arg_type: Type) -> bool:
     Or if it's an imported Caikit data model class.
     Or if it's a Union of at least one of those.
     Or if it's a List of one of those.
+    Or if it's a Dict of one of those.
```
joerunde, line 127: Should this be tested in `test_protoable.py`? I'd expect some tests of `to_protoable_signature` for valid and invalid dict cases

caikit/runtime/service_generation/protoable.py
```diff
@@ -124,25 +124,33 @@ def is_protoable_type(arg_type: Type) -> bool:
...
+        )
+    elif typing.get_origin(arg_type) == Union:
         log.debug2("Arg is Union")
         # pylint: disable=use-a-generator
-        return any([is_protoable_type(arg) for arg in typing.get_args(arg_type)])
+        protoable = any([is_protoable_type(arg) for arg in typing.get_args(arg_type)])
 
     log.debug2("Arg is not protoable, arg_type: %s", arg_type)
-    return False
+    return protoable
```
joerunde, line 156: nit: the above log message is invalid if `protoable` can now be true here

caikit/runtime/service_generation/protoable.py
```diff
@@ -124,25 +124,33 @@ def is_protoable_type(arg_type: Type) -> bool:
...
             return False
-        return typing.get_args(arg_type)[0] in proto_primitive_set
-
-    if typing.get_origin(arg_type) == Union:
+        protoable = typing.get_args(arg_type)[0] in proto_primitive_set
+    elif typing.get_origin(arg_type) == dict:
+        log.debug2("Arg is Dict")
+        if len(typing.get_args(arg_type)) == 0:
+            log.debug2("Dict annotation has no type")
+            return False
```
joerunde, line 145: nit: this is mixing the "return early" pattern with the "save and return later" pattern. I don't really mind picking one over the other as long as we're consistent

tests/runtime/test_grpc_server.py
joerunde: 🤔  In all of these tests we're building the grpc request directly, but we don't necessarily have to. We could use the datamodel classes directly instead.

Definitely don't have to swap for this PR but that could make our tests a bit simpler, and then all the server tests would cover the `.to_proto`ing of request messages as well

tests/fixtures/sample_lib/modules/other_task/other_implementation.py
```diff
@@ -52,10 +62,14 @@ def train(
         training_data: DataStream[int],
         sample_input: Union[SampleInputType, str],
         batch_size: int = 64,
+        training_parameters_json_dict: JsonDict = None,
+        training_parameters: Dict[str, int] = field(default_factory=dict),
```
joerunde, line 66: Similar to Prashant's PR: https://github.com/caikit/caikit/pull/232#discussion_r1232801605

Should we put all these python primitive test cases into the "primitive" test module?

tests/runtime/test_grpc_server.py
```diff
@@ -374,6 +378,8 @@ def test_train_fake_module_does_not_change_another_instance_model_of_block(
         sample_input_sampleinputtype=SampleInputType(name="Gabe").to_proto(),
         batch_size=100,
         training_data=training_data,
+        training_parameters_json_dict=training_parameters_json_dict,
+        training_parameters=training_parameters,
```
joerunde, line 382: Kinda related to the other comment about moving these to the primitive test module, this particular test is already pretty dang beefy and according to the name is testing some model management logic. I think it would be better not to add extra potential failure cases to this test

caikit/runtime/service_generation/protoable.py
```diff
@@ -124,25 +124,33 @@ def is_protoable_type(arg_type: Type) -> bool:
...
+        protoable = (
+            typing.get_args(arg_type)[0] in proto_primitive_set
```
joerunde, line 147: I thought that mappings had to have strings as keys? But I could be way wrong

Would be worth a test to show that other protoables can be map keys if it is possible

joerunde (changes requested):
Thanks Thara!

Implementation looks 🌶️ , mostly just some nits around more test coverage and test organization

caikit/runtime/utils/servicer_util.py
prashantgupta24: Will be handled in a subsequent PR, not needed now

tests/runtime/test_grpc_server.py
```diff
@@ -405,6 +405,53 @@ def test_train_fake_module_does_not_change_another_instance_model_of_block(
     assert original_inference_response == expected_original_inference_response
 
 
+def test_train_primitive_model(
```
joerunde, line 408: nice!

prashantgupta24:
Is it worth having a `simple dict` test in `test_dataobject`?

tests/runtime/service_generation/test_protoable.py
```diff
@@ -111,3 +111,24 @@ def test_to_output_dm_type_with_union_optional_dm():
...
+    ) == {"name": Dict[str, int]}
+
+
+def test_to_protoable_signature_dict_int_keys():
+    assert to_protoable_signature(
+        signature={"name": Dict[int, float]},
+    ) == {"name": Dict[int, float]}
+
+
+def test_to_protoable_signature_unsupported_dict():
```
joerunde, line 128: little nit: can we test the `dict` case as well?
```
assert (
    to_protoable_signature(
        signature={"name": dict},
    )
    == {}
)
```

joerunde (approved):
lgtm, love the test improvements!

tests/runtime/test_grpc_server.py
prashantgupta24: Opened an issue for this as well! https://github.com/caikit/caikit/issues/251

prashantgupta24 (approved):

PR #242, opened 2023-06-15
Cv interfaces

alex-jw-brooks:
Adds preliminary interfaces for vision tasks, starting with:
- image classification
- object detection


alex-jw-brooks:
On second thought - closing this guy for now so we can try out our data model promotion process! 😄

PR #241, opened 2023-06-14
Optional deps 228

gabe-l-hart:
## Description

This PR sets up the framework for optional dependencies in `caikit` and refactors the existing dependency management into the following sets:

Closes: #228 

### User Deps
* `caikit`: Dependencies for `caikit.core`
* `caikit[runtime-grpc]`: Dependencies for the `grpc` runtime
* `caikit[all]`: All dependencies for user functionality in `caikit` (currently core + `runtime-grpc`)

### Developer Deps
* `dev-test`: Dependencies for running unit tests
* `dev-docs`: Dependencies for building documentation
* `dev-fmt`: Dependencies for formatting, linting, and other stylistic checks
* `dev-build`: Dependencies for building the wheel
* `all-dev`: All developer dependencies (including all user dependencies)

**NOTE**: The choice of `-` as a delimiter seems to be required when using `flit` as our build tool. I had originally used `.` which seemed a little cleaner to me, but the `flit` build commands used to build the temporary wheel that `tox` then used to create the venvs was not being built correctly.

PR #245, opened 2023-06-16
Sync to async iter wrapper 224

### LINKED ISSUE #224: Wrap sync in async for fastapi handler

The fastapi handlers are async, and our module inference apis are sync.

We'll need to do something about that, a generic little sync-over-async wrapper should be fine. I bet @gabe-l-hart already has a POC

We could see if there's a sync handler api, but word on the street is that SSE support (server-side events, how we want to do "json streaming") generally needs asnyc

We could also change our generic module inference code to be async, but that would probably just punt the problem down the stack because the `caikit.module` api is synchronous. We probably don't want to ask everybody to write async modules to run their ML models.

gabe-l-hart:
## Description

Closes #224 

Depends on #241 

This PR adds a utility for wrapping synchronous iterators as `async` generators. It is needed for support of SSE streaming in the HTTP server.

### DIFF
--- a/tests/core/toolkit/test_sync_to_async.py
+++ b/tests/core/toolkit/test_sync_to_async.py
@@ -1,0 +1,96 @@
+# Copyright The Caikit Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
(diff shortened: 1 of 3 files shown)


tests/core/toolkit/test_sync_to_async.py
```diff
@@ -0,0 +1,131 @@
...
+    start_event.wait()
+
+    # Use the same pool to start the async iteration
+    agen = async_wrap_iter(lst, pool)
+
+    # Make sure the async iteration is blocked
+    assert pool._work_queue.qsize() == 1
+
+    # Wait for the slow task to finish
+    slow_future.result()
```
joerunde, line 126: instead of relying on timing, can we use a `finish_event`?

```
def slow_task():
    start_event.set()
    finish_event.wait()
...

assert pool._work_queue.qsize() == 1
finish_event.set()
slow_future.result()
```

gabe-l-hart: Yep, much cleaner. Thanks!

joerunde (approved):
lgtm!

PR #248, opened 2023-06-16
:mute: remove WIP warning on local backend

joerunde:
For @evaline-ju 

The context is that this warning could be misleading for any `caikit` module users. While the local backend is registered by default. the "WIP" (work in progress/beta) message shows up in any library using caikit, and individual caikit library module users would not be dealing with the backend directly.


gabe-l-hart (approved):
I'm assuming the reason we're not just removing WIP from backend stuff in general is that it _is_ actually WIP, right?

joerunde:
Yeah, it is _suuuuper_ wip

PR #253, opened 2023-06-19
Make dataobject 252

### LINKED ISSUE #252: Utility for dynamically creating dataobject classes

As a user of caikit, I want the ability to make `@dataobject` classes on-the-fly without mucking around with the `metaclass`'s `__new__` function (as seen [here](https://github.com/caikit/caikit/blob/main/caikit/runtime/service_generation/data_stream_source.py#L206))

Add a `make_dataobject` function that encapsulates the logic for easy reuse

Continue to make it hard so that people don't do it!

## Additional context

On the one hand, this is a usecase that we want to reserve as a power-user-only feature. On the other hand, we do this enough internally around service deduction that the syntax for doing this is too error prone to keep using everywhere.

gabe-l-hart:
## Description

Closes #252

### DIFF
--- a/caikit/core/data_model/dataobject.py
+++ b/caikit/core/data_model/dataobject.py
@@ -19,7 +19,18 @@ model objects inline without manually defining the protobufs representation
 
 # Standard
 from enum import Enum
-from typing import Any, Callable, List, Type, Union, get_args, get_origin
+from typing import (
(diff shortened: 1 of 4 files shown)


evaline-ju (approved):
LGTM!

Issue #256, opened 2023-06-20
data model oneof fields are not sorted


wai25:
## Description

The nlp interface unit tests fail for rules data model where the value type is incorrectly set.

## Discussion

Example:

```
dict_val = {"k1": 1, "k2": 2}
msg = dm.ViewProperty(dict_val)
msg.aql_property["k1"].which_oneof("value")
'float_val'
```

## Acceptance Criteria

- [ ] Unit tests cover new/changed code
- [x] Examples build against new/changed code
- [ ] READMEs are updated
- [ ] Type of [semantic version](https://semver.org/) change is identified

PR #257, opened 2023-06-20
Handle type disambiguation for which_oneof

wai25:
Closes https://github.com/caikit/caikit/issues/256
Make sure that when inferring the which field from the python type, the  value of bool < int < float is respected

### DIFF
--- a/caikit/core/data_model/base.py
+++ b/caikit/core/data_model/base.py
@@ -22,13 +22,13 @@
 # Standard
 from dataclasses import dataclass
 from enum import Enum
-from typing import Any, Dict, Optional, Type, Union
+from typing import Any, Dict, List, Optional, Type, Union
 import base64
 import json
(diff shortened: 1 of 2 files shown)


gabe-l-hart (changes requested):
I think we need two changes on this PR:

1. Per Martin's guidelines, let's update the title and body of this PR a bit (title: `Handle type disambiguation for which_oneof` or something like that)
2. We need to include [the unit test](https://github.com/alanbraz/caikit/blob/nlp_interfaces/tests/core/data_model/test_dataobject.py#L510) too

wai25:
@gabe-l-hart thanks for the pointer. Originally I only looked in test_base.py. Both should be addressed.

gabe-l-hart (approved):
Thanks for bringing this in Majid!

PR #268, opened 2023-06-22
Add REST APIs

### LINKED ISSUE #226: Implement a REST server class (for inference)

We have a `RuntimeGRPCServer` class that implements a grpc server that servers all of our runtime apis. We need the REST version of this.

The rest server (`FastAPIServer`?) class should be able to accept the same set of service packages as the grpc server, and expose REST endpoints for them (actual endpoints.... totally TBD)

This should hook up to the `GlobalPredictServicer` to run model inference, and models with streaming output should be handled with server-side events.

grpc-gateway-wrapper, but it's a clunky build and doesn't give us the flexibility we need to customize our rest endpoints and apis right from python.

## Additional context

According to @gabe-l-hart , the ServiceFactory already outputs everything you need to be able to whip up a fastapi spec and wire up handlers for it.

prashantgupta24:
Add REST APIs in-addition to the gRPC APIs. 

Missing:
- [ ] ADR
- [ ] PR description
- [ ] More tests?

Partial fix #226


caikit/runtime/http_server.py
```diff
@@ -0,0 +1,426 @@
...
+
+# First Party
+from py_to_proto.dataclass_to_proto import Annotated  # Imported here for 3.8 compat
+import aconfig
+import alog
+
+# Local
+from caikit.config import get_config
+from caikit.core.data_model import DataBase
+from caikit.core.data_model.dataobject import make_dataobject, render_dataobject_protos
```
prashantgupta24, line 46: will remove `render_dataobject_protos`

aluu317:
@gabe-l-hart @prashantgupta24 The HTTP server now has a new function called `run_in_thread` that the tests can use to start, and should be killed when the test is done. TLS tests are added.

2 things I wanted to note:
- I didn't refactor the `server_base.py` because both grpc server and http server still use the base `init()` method. And the `find_port()` method is also called by `init()` method.
- The TLS tests now add some overhead time to spin up a server thread with certs, (without TLS tests, total time was around 15s, and with the TLS tests, 75s). Maybe I'm doing something redundant here?

caikit/runtime/http_server.py
```diff
@@ -0,0 +1,460 @@
...
+    def start(self):
+        """Start the server (blocking)"""
+        self.server.run()
+
+    @contextmanager
+    def run_in_thread(self):
+        self.uvicorn_server_thread.start()
+        try:
+            while not self.server.started:
+                time.sleep(1e-3)
```
prashantgupta24, line 180: whoa what does this mean

caikit/config/config.yml
```diff
@@ -80,18 +80,35 @@ runtime:
...
+        # Listen on a unix socket (model mesh feature)
+        unix_socket_path: /tmp/mmesh/grpc.sock
+
+    # Configuration for the http server
+    http:
+        enabled: false
+        port: 8080
+        # String prefix for API routes
+        route_prefix: api/v1
+        timeout: 5
```
gabe-l-hart, line 100: I think this should be a more descriptive name like `graceful_shutdown_timeout`

caikit/runtime/http_server.py
```diff
@@ -0,0 +1,460 @@
...
+    """An implementation of a FastAPI server that serves caikit runtimes"""
+
+    ###############
+    ## Interface ##
+    ###############
+
+    def __init__(self, tls_config_override: Optional[aconfig.Config] = None):
+        super().__init__(get_config().runtime.http.port, tls_config_override)
+        # if not set in config, timeout is None and unvicorn accepts None or number of seconds
+        unvicorn_timeout_graceful_shutdown = get_config().runtime.http.timeout
```
gabe-l-hart, line 108: NIT: This should live down next to the place where it's used. Having it up here makes it difficult to trace how it's used

caikit/runtime/http_server.py
```diff
@@ -0,0 +1,460 @@
...
+        )
+        self.global_predict_servicer = GlobalPredictServicer(inference_service)
+
+        # Set up the central train servicer
+        # TODO: uncomment later on
+        # train_service = ServicePackageFactory().get_service_package(
+        #     ServicePackageFactory.ServiceType.TRAINING,
+        # )
+        # self.global_train_servicer = GlobalTrainServicer(train_service)
+        self.package_name = inference_service.descriptor.full_name.rsplit(".", 1)[0]
```
gabe-l-hart, line 124: NIT: Can we throw a newline and comment between the commented train stuff and the package name? A bit confusing as written since it makes it look like the package name is there instead of the train service (which it's not)

caikit/runtime/http_server.py
```diff
@@ -0,0 +1,460 @@
...
+    # TODO: what should actually happen here?
+    def __exit__(self, type_, value, traceback):
+        pass
+
+    def start(self):
+        """Start the server (blocking)"""
+        self.server.run()
+
+    @contextmanager
+    def run_in_thread(self):
```
gabe-l-hart, line 176: Nice! I think this should actually be merged into `start` and `stop` as follows:

* Add `blocking: bool = True` as an argument to `start` to match how the `grpc` server does it
* If `blocking` is `False`, launch with a thread like you're doing here (also move the creation of the thread inside the condition so that we don't create an unnecessary thread when running with blocking)
* In `stop`, do the `self.server.should_exit = True` and if `self.uvicorn_server_thread is not None`, also `join` it

With this, you can remove the implementations of `__enter__` and `__exit__` here and rely on the ones in the base class!

caikit/runtime/http_server.py
```diff
@@ -0,0 +1,460 @@
...
+            self.app,
+            host="0.0.0.0",
+            port=self.port,
+            log_level=None,
+            log_config=None,
+            timeout_graceful_shutdown=unvicorn_timeout_graceful_shutdown,
+            **tls_kwargs,
+        )
+        self.server = uvicorn.Server(config=config)
+        self.uvicorn_server_thread = threading.Thread(target=self.server.run)
```
gabe-l-hart, line 156: This should move into `start` and only be instantiated if running with `blocking=False` so that we don't create an unnecessary thread

caikit/runtime/http_server.py
```diff
@@ -0,0 +1,460 @@
...
+
+        finally:
+            log.info("Shutting down HTTP Server")
+            self.server.should_exit = True
+            self.uvicorn_server_thread.join()
+
+    ##########
+    ## Impl ##
+    ##########
+    def _stop_metering(self):
```
gabe-l-hart, line 192: Very nice! I wonder if this should be a function on `global_predict_servicer` itself and if it should be called in the grpc server too

caikit/runtime/http_server.py
```diff
@@ -0,0 +1,460 @@
...
+        self.global_predict_servicer.rpc_meter.flush_metrics()
+        self.global_predict_servicer.rpc_meter.end_writer_thread()
+
+    def _bind_routes(self, service: ServicePackage):
+        """Bind all rpcs as routes to the given app"""
+        for rpc in service.caikit_rpcs:
+            rpc_info = rpc.create_rpc_json("")
```
gabe-l-hart, line 199: This will need to be extended for new streaming types! (but not now 😉)

caikit/runtime/server_base.py
```diff
@@ -0,0 +1,83 @@
...
+import aconfig
+import alog
+
+# Local
+from caikit.config import get_config
+
+log = alog.use_channel("SERVR-BASE")
+
+
+class RuntimeServerBase:
```
gabe-l-hart, line 30: Since we can update the HTTP server to have `blocking` as an arg to `start`, let's make this an abstract class (`class RuntimeServerBase(abc.ABC)`) and add `@abc.abstractmethod` signature definitions for `start(self, blocking: bool = True)` and `stop(self)`

tests/runtime/test_http_server.py
```diff
@@ -0,0 +1,255 @@
...
+
+
+def test_insecure_server():
+    with generate_tls_configs():
+        insecure_http_server = http_server.RuntimeHTTPServer()
+        with insecure_http_server.run_in_thread():
+            resp = requests.get(f"http://localhost:{insecure_http_server.port}/docs")
+            resp.raise_for_status()
+
+        insecure_http_server._stop_metering()
```
gabe-l-hart, line 131: We should get this into `def stop` so that it can be in the context manager automatically

tests/runtime/test_http_server.py
```diff
@@ -0,0 +1,255 @@
...
+                    config_overrides["use_in_test"]["client_key"],
+                ),
+            )
+            resp.raise_for_status()
+        http_server_with_mtls._stop_metering()
+
+
+# Third Party
+## Tests #######################################################################
+from fastapi.testclient import TestClient
```
gabe-l-hart, line 172: Let's move this import up to the top and update the comment bar to differentiate from the tests further up

gabe-l-hart (changes requested):
Thanks for taking this so much further @prashantgupta24 and @aluu317! I've got a few cleanup suggestions for how to get the http and grpc servers more tightly aligned, but I think this is really close

caikit/config/config.yml
```diff
@@ -77,17 +77,41 @@ runtime:
...
             cert: ""
 
+    # Configuration for the grpc server
+    grpc:
+        enabled: true
+        port: 8085
+        # Number of workers with which we will run the gRPC server
+        server_thread_pool_size: 5
+        # gRPC Server shutdown grace period
+        # the server shuts down immediately when stopped,
```
aluu317, line 87: added comment for `server_shutdown_grace_period_seconds`

caikit/config/config.yml
```diff
@@ -77,17 +77,41 @@ runtime:
...
     port: 8085
+    unix_socket_path: /tmp/mmesh/grpc.sock
+
+    # Configuration for the http server
+    http:
+        enabled: false
+        port: 8080
+        # String prefix for API routes
+        route_prefix: api/v1
+        # Maximum number of seconds to wait for graceful shutdown
```
aluu317, line 102: added comment for `server_shutdown_grace_period_seconds`

caikit/runtime/grpc_server.py
```diff
@@ -250,15 +246,14 @@ def stop(self, grace_period_seconds: Union[float, int] = None):
             grace_period_seconds (Union[float, int]): Grace period for service shutdown.
                 Defaults to application config
         """
-        if not grace_period_seconds:
```
gabe-l-hart, line 253: Interesting. I'm assuming this is to allow an explicit `0` to be used?

caikit/runtime/http_server.py
```diff
@@ -0,0 +1,477 @@
...
+from caikit.runtime.servicers.global_predict_servicer import GlobalPredictServicer
+from caikit.runtime.types.caikit_runtime_exception import CaikitRuntimeException
+import caikit.core.toolkit.logging
+
+## Globals #####################################################################
+
+log = alog.use_channel("HTTP")
+
+# Registry of DM -> Pydantic model mapping to avoid errors when reusing messages
+# across endpoings
```
gabe-l-hart, line 62: Oops!

```suggestion
# across endpoints
```

caikit/runtime/http_server.py
```diff
@@ -0,0 +1,477 @@
...
+        config = uvicorn.Config(
+            self.app,
+            host="0.0.0.0",
+            port=self.port,
+            log_level=None,
+            log_config=None,
+            timeout_graceful_shutdown=unvicorn_timeout_graceful_shutdown,
+            **tls_kwargs,
+        )
+        self.server = uvicorn.Server(config=config)
```
gabe-l-hart, line 157: I think we need to pre-declare `self.uvicorn_server_thread` (and maybe add a leading `_` to the name) so that line 184 below doesn't raise an `AttributeError` if running synchronously. I think this case could happen if you inverted the pattern here and rather than running the server in a thread, you launched a "shutdown listener" thread that called `stop` based on some signal it got.

```suggestion
        self.server = uvicorn.Server(config=config)
        
        # Placeholder for thread when running without blocking
        self._uvicorn_server_thread = None
```

caikit/runtime/http_server.py
```diff
@@ -0,0 +1,477 @@
...
+            grace_period_seconds (Union[float, int]): Grace period for service shutdown.
+                Defaults to application config
+        """
+        log.info("Shutting down HTTP Server")
+        self.server.should_exit = True
+        if (
+            self.uvicorn_server_thread is not None
+            and self.uvicorn_server_thread.is_alive()
+        ):
+            self.uvicorn_server_thread.join()
```
gabe-l-hart, line 187:
```suggestion
            self._uvicorn_server_thread is not None
            and self._uvicorn_server_thread.is_alive()
        ):
            self._uvicorn_server_thread.join()
```

caikit/runtime/http_server.py
```diff
@@ -0,0 +1,477 @@
...
+
+        # Ensure we flush out any remaining billing metrics and stop metering
+        if self.config.runtime.metering.enabled:
+            self.global_predict_servicer.stop_metering()
+
+    def run_in_thread(self):
+        self.uvicorn_server_thread = threading.Thread(target=self.server.run)
+        self.uvicorn_server_thread.start()
```
gabe-l-hart, line 195:
```suggestion
        self._uvicorn_server_thread = threading.Thread(target=self.server.run)
        self._uvicorn_server_thread.start()
```

docs/adrs/021-http-server.md
gabe-l-hart, line 1: I think we need to flesh this out to follow [the template](https://github.com/caikit/community/blob/main/adrs/README.md#template)

tests/runtime/test_grpc_server.py
```diff
@@ -67,7 +66,7 @@
 from tests.conftest import random_test_id
 from tests.fixtures import Fixtures
 from tests.runtime.conftest import runtime_grpc_test_server
-import caikit
+import caikit.interfaces.common
```
gabe-l-hart, line 69: I forget exactly why I added this, but I _think_ it was to fix a bug that got [fixed elsewhere](https://github.com/caikit/caikit/pull/272)

tests/runtime/test_http_server.py
```diff
@@ -0,0 +1,241 @@
...
+def test_mutual_tls_server():
+    with generate_tls_configs(
+        tls=True, mtls=True, http_config_overrides={}
+    ) as config_overrides:
+        http_server_with_mtls = http_server.RuntimeHTTPServer(
+            tls_config_override=config_overrides["runtime"]["tls"]
+        )
+        # start a non-blocking http server with mutual tls
+        with http_server_with_mtls:
+            resp = requests.get(
```
gabe-l-hart, line 150: It would be good to verify that if the client does NOT send a valid client cert it gets rejected as well

gabe-l-hart (changes requested):
A few more little NITs, but this is looking really close. Thanks for all the hard work getting this in shape @aluu317 and @prashantgupta24!

caikit/runtime/grpc_server.py
aluu317: yes, I can be swayed either way tho?

gabe-l-hart: Given that this is just used to check the config value, I think this is a little better

gabe-l-hart (approved):
Let's ship it!

Issue #270, opened 2023-06-23
dataobjects don't have static typing


sboagibm:
## Describe the bug

For the `@dataobject` decorator, you loose the tool-time ability to have typing on the dataobject class, because the way the decorator is typed, with `Type[DataBase]` hides the original class type.  

So, in PyCharm:

1. Trying to access instance variables by dot addressing will show as errors, and won't allow code completion.
<img width="578" alt="image" src="https://github.com/caikit/caikit/assets/6641871/cb4b2304-e361-48a3-8b2b-a4dbbfe3a761">

2. The constructor is missing and apparently PyCharm is not smart enough to know the dataobject is a `dataclass`.  With a dataclass, you do get the constructor hints:
<img width="325" alt="image" src="https://github.com/caikit/caikit/assets/6641871/a97a4f53-55b8-4433-8060-01746c926a89">

## Platform

Please provide details about the environment you are using, including the following:

- Interpreter version: Python 3.11.4
- Library version:  main ee0fbf4434ae84eb72d7eb79dfd8a006c28d6b7f
- PyCharm 2023.1.3 (Professional Edition) Build #PY-231.9161.41, built on June 20, 2023
  VM: OpenJDK 64-Bit Server VM by JetBrains s.r.o.
  macOS 13.4
  Memory: 12000M
  Cores: 10

## Sample Code

```Python
from typing import List

from caikit.core import dataobject, DataObjectBase


@dataobject(package="caikit_data_model.wdu")
class TextLine(DataObjectBase):
    bbox: List[float]  # Coordinates
    text_line_id: int


def test_text_line_to_proto_and_back():
    msg = TextLine(bbox=[22.0, 45.0, 324.0, 897.0], text_line_id=934)

    assert msg.bbox == [22.0, 45.0, 324.0, 897.0]
    assert msg.text_line_id == 934
```

Shows up in the ID as:
<img width="642" alt="image" src="https://github.com/caikit/caikit/assets/6641871/c7205bd9-9e40-44e1-8b81-b641072e721e">

## Expected behavior

Trying to access instance variables by dot addressing won't show as errors, and will show code completion options.

## Observed behavior

Per above screen dumps.

## Additional context

Slack discussion: https://ibm-analytics.slack.com/archives/C04K2B8QUG7/p1687233366250279

PR #271, opened 2023-06-23
fix: Fix initial problems with typing of dataobject decorator

sboagibm:
This uses a `TypeVar` for the dataobject decorator, which addresses the first part of [Issue 227](https://github.com/caikit/caikit/issues/270). 

But the `TypeVar` does not fix the second issue, about type completion for the constructor.  I think the only solution for that is to add the `dataclass` decorator.  To that effect, I've adjusted the docstring for `dataobject`.

This PR also fixes some type inspection issues for the modified file, `dataobject.py` per principle of clean code practice (i.e. I would rather avoid submitting files with type check errors in them).

For [ai-foundation/watson-nlp-tracker/issue#](https://github.com/caikit/caikit/issues/270)

### DIFF
--- a/caikit/core/data_model/dataobject.py
+++ b/caikit/core/data_model/dataobject.py
@@ -27,10 +27,10 @@ from typing import (
     List,
     Optional,
     Type,
+    TypeVar,
     Union,
     get_args,
     get_origin,
-    TypeVar,
 )
 import dataclasses
 
@@ -132,7 +132,7 @@ class DataObjectBase(DataBase, metaclass=_DataObjectBaseMetaClass):
 _DataObjectBaseT = TypeVar("_DataObjectBaseT", bound=DataObjectBase)
 
 
-def dataobject(*args, **kwargs) -> Callable[[_DataObjectBaseT], Type[_DataObjectBaseT]]:
+def dataobject(*args, **kwargs) -> Callable[[_DataObjectBaseT], _DataObjectBaseT]:
     """The @dataobject decorator can be used to define a Data Model object's
     schema inline with the definition of the python class rather than needing to
     bind to a pre-compiled protobufs class. For example:
@@ -148,6 +148,20 @@ def dataobject(*args, **kwargs) -> Callable[[_DataObjectBaseT], Type[_DataObject
         directly, the metaclass that links protobufs to the class will be called
         before this decorator can auto-gen the protobufs class.
 
+    The `dataobject` decorator will not provide tools with enough information
+    to perform type completion for constructions in an IDE, or static
+    typechecking.  In order to have that, the `dataclass` decorator
+    may optionally be added, with the slight overhead of wasted effort in
+    creating the "standard" __init__ function which then gets re-done by
+    @dataobject.  The `dataclass` must follow the `dataobject` decorator.  For example:
+
+    @dataobject("foo.bar")
+    @dataclass
+    class MyDataObject(DataObjectBase):
+        '''My Custom Data Object'''
+        foo: str
+        bar: int
+
     Kwargs:
         package:  str
             The package name to use for the generated protobufs class
@@ -157,7 +171,7 @@ def dataobject(*args, **kwargs) -> Callable[[_DataObjectBaseT], Type[_DataObject
             The decorator function that will wrap the given class
     """
 
-    def decorator(cls: _DataObjectBaseT) -> Type[_DataObjectBaseT]:
+    def decorator(cls: _DataObjectBaseT) -> _DataObjectBaseT:
         # Make sure that the wrapped class does NOT inherit from DataBase
         error.value_check(
             "<COR95184230E>",
@@ -277,7 +291,7 @@ def make_dataobject(
             class derived from DataObjectBase with the given name and
             annotations
     """
-    bases = tuple([Type[DataObjectBase]] + list(bases or []))
+    bases = (DataObjectBase,) + tuple(bases or ())
     attrs = {
         "__annotations__": annotations,
         **(attrs or {}),
--- a/caikit/core/data_model/dataobject.py
+++ b/caikit/core/data_model/dataobject.py
@@ -30,6 +30,7 @@ from typing import (
     Union,
     get_args,
     get_origin,
+    TypeVar,
 )
 import dataclasses
 
@@ -128,13 +129,15 @@ class DataObjectBase(DataBase, metaclass=_DataObjectBaseMetaClass):
     """
 
 
-def dataobject(*args, **kwargs) -> Callable[[Type], Type[DataBase]]:
+_DataObjectBaseT = TypeVar("_DataObjectBaseT", bound=DataObjectBase)
+
+
+def dataobject(*args, **kwargs) -> Callable[[_DataObjectBaseT], Type[_DataObjectBaseT]]:
     """The @dataobject decorator can be used to define a Data Model object's
     schema inline with the definition of the python class rather than needing to
     bind to a pre-compiled protobufs class. For example:
 
     @dataobject("foo.bar")
-    @dataclass
     class MyDataObject(DataObjectBase):
         '''My Custom Data Object'''
         foo: str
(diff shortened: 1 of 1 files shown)


caikit/core/data_model/dataobject.py
```diff
@@ -30,6 +30,7 @@
     Union,
     get_args,
     get_origin,
+    TypeVar,
```
gabe-l-hart, line 33: NIT: I think this will fail `fmt`

caikit/core/data_model/dataobject.py
```diff
@@ -128,13 +129,15 @@ class DataObjectBase(DataBase, metaclass=_DataObjectBaseMetaClass):
     """
 
 
-def dataobject(*args, **kwargs) -> Callable[[Type], Type[DataBase]]:
+_DataObjectBaseT = TypeVar("_DataObjectBaseT", bound=DataObjectBase)
+
+
+def dataobject(*args, **kwargs) -> Callable[[_DataObjectBaseT], Type[_DataObjectBaseT]]:
```
gabe-l-hart, line 135: Interesting. I think this original type hint was from before I introduced `DataObjectBase` and the decorator would inject `DataBase` as a base class. now that we require the input class to explicitly inherit from `DataObjectBase`, I think the input argument should also be `Type[_DataObjectBaseT]` and in fact it _might_ be just ok with `Type[DataObjectBase]`?

caikit/core/data_model/dataobject.py
```diff
@@ -145,7 +148,7 @@ class MyDataObject(DataObjectBase):
         directly, the metaclass that links protobufs to the class will be called
         before this decorator can auto-gen the protobufs class.
 
-    Args:
+    Kwargs:
```
gabe-l-hart, line 163: Nice, thanks for fixing this

caikit/core/data_model/dataobject.py
```diff
@@ -199,7 +214,7 @@ def decorator(cls: Type) -> Type[DataBase]:
 
         # Declare the merged class that binds DataBase to the wrapped class with
         # this generated proto class
-        if isinstance(proto_class, type):
+        if not isinstance(proto_class, EnumTypeWrapper):
```
gabe-l-hart, line 217: Took me a bit to catch what was going on here, but I think this is right

caikit/core/data_model/dataobject.py
```diff
@@ -274,7 +289,7 @@ def make_dataobject(
             class derived from DataObjectBase with the given name and
             annotations
     """
-    bases = tuple([DataObjectBase] + list(bases or []))
+    bases = tuple([Type[DataObjectBase]] + list(bases or []))
```
gabe-l-hart, line 292: This is interesting. These `bases` are actually the concrete bases passed to the metaclass' `__new__` function, so I actually _explicitly_ want `DataObjectBase` here. I'm honestly not sure what happens if you pass `Type[BaseClassIWant]` to the `__new__` of a `type`

caikit/core/data_model/dataobject.py
```diff
@@ -310,7 +325,7 @@ def get_concrete_type(self, entry: Any) -> Any:
         if (
             isinstance(unwrapped, type)
             and issubclass(unwrapped, DataBase)
-            and unwrapped._proto_class is not None
+            and entry.get_proto_class() is not None
```
gabe-l-hart, line 328: Yep, that's better. Thanks

caikit/core/data_model/dataobject.py
gabe-l-hart: I also wonder if the funkiness around needing to add `@dataclass` is because this type hint should technically include a `Union[]` to itself for the special case handling when it's called with no arguments ([line 223](https://github.com/caikit/caikit/pull/271/files#diff-96758051650f0b4a3b80930a8851adc1c588bf8d4d8c22bfe7e4751a9f094e71R238))

gabe-l-hart (changes requested):
I think this needs a `fmt` and I think the type hint on the `dataobject` function can be tighter now that you've started.

caikit/core/data_model/dataobject.py
```diff
@@ -145,7 +148,19 @@ class MyDataObject(DataObjectBase):
         directly, the metaclass that links protobufs to the class will be called
         before this decorator can auto-gen the protobufs class.
 
-    Args:
+    The `dataobject` decorator will not provide tools with enough information to perform type completion
```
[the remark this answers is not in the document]
sboagibm: Hmm, sure, will do, but I think the default column width seems to be 100, per `.pylintrc`?  (I had my PyCharm set  at 120)

Also, should we have
```
[tool.black]
line-length = 100
include = '\.pyi?$'
exclude = '''
/(
    \.git
  | \.hg
  | \.mypy_cache
  | \.tox
  | \.venv
)/
'''
```

in `pyproject.toml`, so that black formatting will work consistently?

caikit/core/data_model/dataobject.py
sboagibm: Yeah, I think I misread the code there, due to rather confusing error message.  The error I am trying to fix is:
image

After a little more digging, this seems to be some oddity of the tuple construction, it wants a list that is a single type, some sort of co-constraint thing I guess.
image

Seems to be a type error and not an indication of actual error, though maybe it inhibits an optimization, not sure.

I tried a bunch of things.  This seems to work type-wise and , and is fairly clean and elegant, and doesn't construct intermediary lists:
```python
bases = (DataObjectBase,) + tuple(bases or ())
```

I spent way too much time on this.  :-)

caikit/core/data_model/dataobject.py
sboagibm: Otherwise you get:
image

Which I think the `if isinstance(proto_class, type)` forces it to be a type, and you would need to do a cast down or something to pass it to that function.

caikit/core/data_model/dataobject.py
gabe-l-hart: ha! Love me a rabbit hole. This looks good, thanks for digging so deep!

caikit/core/data_model/dataobject.py
sboagibm: @gabe-l-hart said:
>  think the input argument should also be Type[_DataObjectBaseT]

Hmm, ok, done.  I'm a little unclear at what `Type[C]` does exactly.   Doesn't the TypeVar already do the trick that  `Type[C]` does, i.e. ["while C (when used as an annotation) refers to instances of class C, Type[C] refers to subclasses of C."](https://peps.python.org/pep-0484/#the-type-of-class-objects)?  If I don't wrap `_DataObjectBaseT` at all with `Type`, it seems to make no difference.  

> and in fact it might be just ok with Type[DataObjectBase]

Seems to be, though I'm a little surprised.  I thought the whole point was the input class defined the output class.  If we define input as `Type[DataObjectBase]` , how does it know that `_DataObjectBaseT` is `TextLine`, for example?  I thought there was some sort of co-constraint thing going on with input and output.

Checking it in for now with

```python
Callable[[_DataObjectBaseT], _DataObjectBaseT]
```

I wish I had a bit better way of playing with type inference.

sboagibm: @gabe-l-hart also said:
> I also wonder if the funkiness around needing to add @dataclass is because this type hint should technically include a Union[] to itself for the special case handling when it's called with no arguments ([line 223](https://github.com/caikit/caikit/pull/271/files#diff-96758051650f0b4a3b80930a8851adc1c588bf8d4d8c22bfe7e4751a9f094e71R238))

I've no idea what that would look like, or what you mean by "itself".  It seems to me that doing argument completion for dataclasses without initializers have to be solely a matter of the tooling being able to understand that they are dataclasses?

caikit/core/data_model/dataobject.py
sboagibm: This is fixed, or at least `tox fmt` passes on my machine.

caikit/core/data_model/dataobject.py
gabe-l-hart: Yeah, this is definitely a game of reverse engineering what the IDE is doing, but it could either be using `@dataclass` as a keyword directly or it could be knowledgable about all the different flavors of decorators (no-call, zero-arg, with-arg) and what their type hint signatures are.

gabe-l-hart (approved):
Thanks for tightening this up Scott!

PR #272, opened 2023-06-23
🐛 caikit.interfaces.common bug fix

prashantgupta24:
Import `caikit.interfaces.common` explicitly in `caikit/runtime/service_generation/data_stream_source.py` since this module needs it.


gabe-l-hart (approved):
Thanks Prashant!

Issue #54, opened 2023-04-21
Reshuffle backend semantics


joerunde:
~~We want to make the backends configuration a bit cleaner and to fit in nicely with the top-level `caikit.configure()` method.
We don't really want an extra `backends.configure()` that users need to call separately.~~ JK we do want `caikit.configure()` to be side-effect free

Goals:
- The backend configuration should be smart enough to not barf if called multiple times. Currently it can only be called once or it throws. We should probably:
  - Actually call `backend_instance.reconfigure()` if config for that backend has changed
  - Throw only if that call throws
  - Do not attempt to call reconfigure if the configuration for that backend has not changed
- `start` and `stop` are interesting, they can probably be invoked lazily by the backend itself. Do we need the public `start_all` that is currently unused?

That should alleviate the burden of many model loading tests needing to use the clunky `reset_globals` (needs a better name anyway) config. New backend configurations shouldn't collide with existing ones from other tests.

We should also consider refactoring to extract the loading logic of backends into its own module (or at least method) for better decoupling and testability.

PR #273, opened 2023-06-23
Add a basic base factory to codify factory keys and semantics

gabe-l-hart:
This PR adds a simple `Factory` that can be reused across different config-oriented factory objects. There's not a lot of code, but it does make explicit that the `"type"` and `"config"` strings should be used and how they should be organized, so it's worth standardizing as the number of factories grows.

Supports: https://github.com/caikit/caikit/issues/54


gkumbhat (approved):
LGTM. Thanks Gabe

Issue #258, opened 2023-06-20
Intermittent test failures on grpc server tests


evaline-ju:
## Describe the bug

The grpc server tests will sometimes fail. Failures have been seen on 3.8 and 3.10.

## Platform

Github actions with failing tests: 
- https://github.com/caikit/caikit/actions/runs/5313668900/jobs/9619870475
- https://github.com/caikit/caikit/actions/runs/5326984887/jobs/9649759901?pr=257

## Sample Code

Pointer to tests that have failed: 
- https://github.com/caikit/caikit/blob/b3e25e9a767cb7ad64887311e6893853c14161f9/tests/runtime/test_grpc_server.py#L494
- https://github.com/caikit/caikit/blob/b3e25e9a767cb7ad64887311e6893853c14161f9/tests/runtime/test_grpc_server.py#L323

## Expected behavior

GRPC server tests consistently pass

## Observed behavior

Example
<img width="1309" alt="Screen Shot 2023-06-20 at 3 28 58 PM" src="https://github.com/caikit/caikit/assets/69598118/4aaab40d-8e93-44b5-a69c-e4252209de35">
<img width="1356" alt="Screen Shot 2023-06-20 at 3 28 49 PM" src="https://github.com/caikit/caikit/assets/69598118/7b65b0f3-ba6e-48ed-9e0e-08e0cf27a662">

Test intermittently fails

PR #274, opened 2023-06-23
Don't error out of the training manager when not started

gabe-l-hart:
The test was flakey because it was hitting a race where the future had not yet started or completed and fell through to the Unexpected Error below. That error was there to handle shutdown and cancellation which seems unnecessary.

Closes: https://github.com/caikit/caikit/issues/258


hickeyma (approved):
Logic seems sound and test running locally and in CI. Its hard to be sure if flakiness is indeed fixed but time will tell.

Thanks for the fix @gabe-l-hart

PR #276, opened 2023-06-26
Add a way to disable just the test_exaples

gabe-l-hart:
They're really slow and I'm not touching them right now!

Supports: https://github.com/caikit/caikit/issues/54

### DIFF
--- a/tests/examples/test_examples.py
+++ b/tests/examples/test_examples.py
@@ -25,6 +25,7 @@ from tests.examples.shared import requirements, waitForPort
 
 
 @pytest.mark.examples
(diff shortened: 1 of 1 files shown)


prashantgupta24:
This should already be in place, 

```
pytest tests -m 'not examples'
```

skips all example tests.

Also using `tox` for testing automatically skips the example tests (which uses `-m 'not examples'` in the background)

prashantgupta24 (changes requested):
This should already be in place,
```
pytest tests -m 'not examples'
skips all example tests.
```
Also using tox for testing automatically skips the example tests (which uses `-m 'not examples'` in the background)

gabe-l-hart:
🤦 Thanks Prashant! Read the docs Gabe

PR #279, opened 2023-06-26
Fix None edge case of docstring converter

willmj:
Addressing an edge case I didn't notice during initial PR #211. arg_replacement function would sometimes append the word "None" in front of an argument when it shouldn't have been there. This is because in another edge case where there is a description but no data type, I forgot to include the case that there might not be additional description past the first line. I also went back and manually fixed where "None" was appended in docstrings from the previous PR.

### DIFF
--- a/scripts/test_dsconverter.py
+++ b/scripts/test_dsconverter.py
@@ -805,9 +805,7 @@ def test_multiword_args():
                 The index of the object in the stream
         """
     '''
-    print(test1)
     converted = converter.convert_to_google_style(test1)
-    print(converted)
     assert (
         converted
         == '''
--- a/scripts/test_dsconverter.py
+++ b/scripts/test_dsconverter.py
@@ -792,6 +792,7 @@ def test_multiword_args():
     '''
     )
 
+
 def test_multiword_args():
     converter = CustomDocstringConverter()
     test1 = '''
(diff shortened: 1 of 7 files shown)


scripts/test_dsconverter.py
```diff
@@ -793,6 +793,34 @@ def test_multiword_args():
...
+        Args:
+            data_item: 
+                A data object yielded by the stream
+            data_item_number: 
+                The index of the object in the stream
+        """
+    '''
+    print(test1)
+    converted = converter.convert_to_google_style(test1)
+    print(converted)
```
joerunde, line 810: tiny nit that we try to avoid putting `prints` in test cases, since that'll spit out a lot of strings to stdout when a test fails. (Which may be good for you as the test author but confusing for a different person who is coming in and editing some code). We try to instead use debuggers during testing or using logging to get this visibility in a more user-configurable way

joerunde (approved):
Thanks for the quick fix! Would be nice to nix the `prints` in the tests before context-switching away if you have time

scripts/test_dsconverter.py
willmj: I agree! I am used to debugging with prints which definitely isn't best-practice. In the future, I will attempt to use a debugger. Prints have been nixed though, so the the test should be good to go!

PR #288, opened 2023-06-29
Model manager training

gabe-l-hart:
## Description

This PR extends the `finder`/`initializer` implementations behind `ModelManager.load` to two net-new public functions on `ModelManager`: `ModelManager.train` and `ModelManager.get_model_future`. These functions provide asynchronous traning support that can be abstracted behind the `ModelTrainer` abstractions introduced in #275


caikit/core/model_manager.py
```diff
@@ -67,12 +69,119 @@ class ModelManager:
     def __init__(self):
         """Initialize ModelManager."""
         # Map to store module caches, to be used for singleton model lookups
-        self.singleton_module_cache = {}
-        self._singleton_lock = Lock()
+        self._singleton_module_cache = {}
```
gabe-l-hart, line 72: Here I was just trying to shore up consistency on "protected" naming

caikit/core/model_manager.py
```diff
@@ -67,12 +69,119 @@ class ModelManager:
...
+        self._trainers = {}
         self._finders = {}
         self._initializers = {}
+        self.__singleton_lock = Lock()
```
gabe-l-hart, line 76: This gets a double `__` so that the wrapper `contextmanager` can be `self._singleton_lock`

caikit/core/model_manager.py
```diff
@@ -67,12 +69,119 @@ class ModelManager:
...
+                submission's ModelFuture
+
+        Returns:
+            model_future (ModelTrainerBase.ModelFutureBase): The future handle
+                to the model which holds the status of the in-flight training.
+        """
+        try:
+            trainer = self._get_trainer(ModelTrainerBase.get_trainer_name(training_id))
+
+        # Fall back to the default trainer to try to find this ID
```
gabe-l-hart, line 179: This is done to support external IDs when running locally for Model Train. These external IDs will not contain the trainer name prefix since they'll come from an external system

caikit/core/model_manager.py
```diff
@@ -153,6 +262,110 @@ def load(
                 ),
             )
 
+    def extract(
```
gabe-l-hart, line 265: This just got moved so that public functions are all grouped together

caikit/core/model_manager.py
```diff
@@ -153,6 +262,110 @@ def load(
...
+            )
+            return model_path
+
+        with zipfile.ZipFile(zip_path, "r") as zip_f:
+            zip_f.extractall(model_path)
+
+        # path to model
+        return model_path
+
+    def resolve_and_load(
```
gabe-l-hart, line 295: Moved for public function grouping

caikit/core/model_manager.py
```diff
@@ -153,6 +262,110 @@ def load(
...
+        error(
+            "<COR50207495E>",
+            ValueError(
+                "could not find model with name `{}`".format(
+                    path_or_name_or_model_reference
+                )
+            ),
+        )
+
+    def get_singleton_model_cache_info(self):
```
gabe-l-hart, line 349: Moved for public function grouping

caikit/core/model_manager.py
```diff
@@ -153,6 +262,110 @@ def load(
...
+        """Returns information about the singleton cache in {hash: module type} format
+
+        Returns:
+            Dict[str, type]: A dictionary of model hashes to model types
+        """
+        return {k: type(v) for k, v in self._singleton_module_cache.items()}
+
+    def clear_singleton_cache(self):
```
gabe-l-hart, line 357: Moved for public function grouping

caikit/core/model_manager.py
```diff
@@ -287,113 +500,13 @@ def get_full_path(folder_name):
                     )
         return model
 
-    def extract(self, zip_path, model_path, force_overwrite=False):
```
gabe-l-hart, line 290: Moved for public function grouping

caikit/core/toolkit/factory.py
```diff
@@ -46,7 +46,7 @@ def name(cls) -> str:
         """
 
     @abc.abstractmethod
-    def __init__(self, config: aconfig.Config):
+    def __init__(self, config: aconfig.Config, instance_name: str):
```
gabe-l-hart, line 49: To support constructibles that are aware of their own name for ID creation, all constructibles must be able to take an `instance_name` argument. If the derived class doesn't need it, it's free to ignore it.

caikit/core/__init__.py
```diff
@@ -42,3 +42,5 @@
 extract = MODEL_MANAGER.extract
 load = MODEL_MANAGER.load
 resolve_and_load = MODEL_MANAGER.resolve_and_load
+train = MODEL_MANAGER.train
+get_model_future = MODEL_MANAGER.get_model_future
```
gkumbhat, line 46: do we need to expose this at this level? This seems like more of an internal use only?

gkumbhat (commented):
Thanks Gabe, just left a tiny question, otherwise looks good to me!

caikit/core/__init__.py
gabe-l-hart: I think we do want to expose this here (and in the top of `caikit`) since they would be how, for example, you'd launch a remote training job from a notebook with a non-local trainer

gkumbhat (approved):
Looks good to me!

Issue #292, opened 2023-06-30
Producer priority no longer has producers


evaline-ju:
## Describe the bug

Users of `caikit` noticed that `producers` was no longer available on `ProducerPriority`

```
/usr/local/lib/python3.8/site-packages/caikit/interfaces/common/data_model/producer.py:50: in __init__
    self.producers = producers
self = <[AttributeError("<class 'caikit.interfaces.common.data_model.producer.ProducerPriority'> missing attribute elements and no backend set") raised in repr()] ProducerPriority object at 0x7f7f1fe33700>
name = 'producers', val = [ProducerId(name='A', version='9999'), ProducerId(name='B', version='9999')]
.......
E           AttributeError: 'ProducerPriority' object has no attribute 'producers'
```
## Platform

Please provide details about the environment you are using, including the following:

- Interpreter version: 3.8
- Library version: latest

## Expected behavior

`producers` is expected to be an available attribute on `ProducerPriority`

## Observed behavior

Only `elements` is available

## Additional context

Seems to be a refactoring typo here: https://github.com/caikit/caikit/commit/3b08dfd580fd59e3a282087625a7835ca88b0def

PR #293, opened 2023-06-30
🐛 Add back producers for producer priority

evaline-ju:
Closes: #292 

Removing the init since dataobject decorator should do init. The custom logic is not particularly needed here, also adding tests


gabe-l-hart (approved):
Thanks for the fix!

Issue #289, opened 2023-06-29
Lists in unions with same arg name causes dm class conflict error


prashantgupta24:
## Describe the bug

The approach of dealing with Lists in Unions has a major flaw: if the same argument name is used in multiple places `union_list: Union[List[str], List[int]]`, then it registers multiple data model classes for it, which causes conflict.

```
Conflicting data model classes for [UnionListStrSequence]: ['caikit.runtime.SampleLib.SampleTaskSampleModuleTrainRequest.UnionListStrSequence', 'caikit.runtime.SampleLib.SampleTaskSamplePrimitiveModuleTrainRequest.UnionListStrSequence']
```
## Platform

Please provide details about the environment you are using, including the following:

- Interpreter version: 3.9
- Library version: 0.9.2

## Context

I think it has to be solved in the same way that `DataStreamSource` solves `DataStream[SampleTrainingType]` dm_model class, registering it once centrally and reusing that every time it's needed.

PR #294, opened 2023-06-30
♻️ create dm objects for union lists

prashantgupta24:
This PR handles params for union lists. For supporting `DataObjects` used by the union of lists, I've added all to `caikit.interfaces.common.data_model` instead of creating them dynamically.

fix https://github.com/caikit/caikit/issues/289


caikit/runtime/service_generation/protoable.py
```diff
@@ -91,6 +107,49 @@ def handle_protoables_in_union(arg_type: Type) -> Type:
...
+def make_union_dataobjects(union_protoables: List) -> Type[DataBase]:
+    """Dynamically create a union list source message type"""
+    package = "caikit_data_model.runtime"
+    cls_name = _make_union_list_source_type_name(union_protoables)
+    if cls_name not in _UNION_LIST_SOURCE_TYPES:
+        data_object_map = {}
+        data_object_list = []
+        for arg in union_protoables:
+            arg_type = get_args(arg)[0]
+            arg_name = f"{arg_type.__name__.capitalize()}Sequence"
```
prashantgupta24, line 124: Based on @joerunde 's comments, I will create a subsequent PR in which I stick all of these `IntSequence`, `StrSequence` etc. directly into the `caikit runtime interfaces`, and reference them when they're needed

prashantgupta24: Also `file, filelist, and directory` from `DataStreamSource`

prashantgupta24: Resolved `#1` in this issue, opened https://github.com/caikit/caikit/issues/300 for `#2`.

caikit/interfaces/runtime/data_model/static_interfaces.py
prashantgupta24, line 1: Happy to change the name to something better!

caikit/runtime/service_generation/protoable.py
```diff
@@ -91,6 +107,47 @@ def handle_protoables_in_union(arg_type: Type) -> Type:
...
+            if not hasattr(caikit.interfaces.runtime.data_model, arg_name):
+                raise AttributeError(
```
prashantgupta24, line 126: need to figure out how to test this line

caikit/interfaces/runtime/data_model/static_interfaces.py
gabe-l-hart: I think these should go in `caikit.interfaces.common` and be called something like `primitive_sequences`

tests/fixtures/sample_lib/modules/sample_task/sample_implementation.py
```diff
@@ -69,6 +70,7 @@ def save(self, model_path):
     def train(
         cls,
         training_data: DataStream[SampleTrainingType],
+        union_list: Union[List[str], List[int]] = None,
```
prashantgupta24, line 73: I needed another module that has the same name argument for unions of lists

caikit/interfaces/common/data_model/__init__.py
```diff
@@ -26,4 +26,5 @@
 # Local
 # Import individual packages
 from . import producer
+from .primitive_sequences import BoolSequence, FloatSequence, IntSequence, StrSequence
```
gabe-l-hart, line 29: For consistency, we want to add `primitive_sequences` to the line above here

caikit/interfaces/common/data_model/primitive_sequences.py
```diff
@@ -0,0 +1,27 @@
+# Standard
+from typing import List
+
+# Local
+from caikit.core.data_model import DataObjectBase, dataobject
+from caikit.interfaces.runtime.data_model.training_management import RUNTIME_PACKAGE
+
+
+# interfaces required for unions of lists
```
gabe-l-hart, line 9: NIT: Let's put this as the module docstring

caikit/interfaces/common/data_model/primitive_sequences.py
```diff
@@ -0,0 +1,27 @@
+# Standard
```
gabe-l-hart, line 1: Missing copyright header

caikit/interfaces/common/data_model/primitive_sequences.py
```diff
@@ -0,0 +1,27 @@
...
+
+# Local
+from caikit.core.data_model import DataObjectBase, dataobject
+from caikit.interfaces.runtime.data_model.training_management import RUNTIME_PACKAGE
+
+
+# interfaces required for unions of lists
+@dataobject(RUNTIME_PACKAGE)
+class IntSequence(DataObjectBase):
+    values: List[int]
```
gabe-l-hart, line 12: To ensure explicit compatibility, let's use `Annotated[List[xxxx]], FieldNumber(1)]` in each of these

caikit/runtime/service_generation/protoable.py
```diff
@@ -16,15 +16,26 @@
...
+    DATAOBJECT_PY_TO_PROTO_TYPES,
+    make_dataobject,
+)
 from caikit.runtime.service_generation.type_helpers import is_data_model_type
+import caikit
+
+# This global holds the mapping of union list types to their respective
+# UnionList wrappers so that the same message is not recreated
+# unnecessarily
+_UNION_LIST_SOURCE_TYPES = {}
```
gabe-l-hart, line 38: Does this need to be pre-populated with the static ones?

caikit/runtime/service_generation/protoable.py
```diff
@@ -66,6 +77,11 @@ def handle_protoables_in_union(arg_type: Type) -> Type:
                 for union_val in typing.get_args(arg_type)
                 if is_protoable_type(union_val)
             ]
+            # handle union of lists in a separate way
+            if len(union_protoables) > 1 and all(
```
gabe-l-hart, line 81: Why is the logic here using `all`? What about `Union[List[int], int]`? I would think we'd want to support this too

caikit/runtime/service_generation/protoable.py
```diff
@@ -91,6 +107,47 @@ def handle_protoables_in_union(arg_type: Type) -> Type:
     log.debug("Skipping non-protoable argument type [%s]", arg_type)
 
 
+def _make_union_list_source_type_name(args: List) -> str:
+    """Make the name for  source class that wraps the given types"""
+    return f"UnionList{''.join([get_args(arg)[0].__name__.capitalize() for arg in args])}Source"
+
+
+def make_union_dataobjects(union_protoables: List) -> Type[DataBase]:
```
gabe-l-hart, line 115: I think this is introducing one unnecessary level of indirection. If I'm reading this correctly, this is creating a net-new message with a single field `union_list` whose type hint is a union of the associated sequence wrapper classes. I think that intermediate class is not needed since the message types in the `Union` already provide the wrapping to allow them to exist in a `Union`. Basically, I think we want to take the type hint you're setting on line 137 and use that as the type hint for the actual field that's being auto-deduced, rather than adding the indirection of the `union_list` field.

tests/fixtures/sample_lib/modules/sample_task/primitive_party_implementation.py
```diff
@@ -79,13 +80,16 @@ def train(
         """Sample training method that produces a trained model"""
         assert type(sample_input) == SampleInputType
         assert isinstance(simple_list, List)
-        assert isinstance(union_list.values, List)
+        assert isinstance(union_list1.union_list.values, List)
```
gabe-l-hart, line 83: If we do my suggestion, the `union_list` level of nesting goes away here

tests/fixtures/sample_lib/modules/sample_task/sample_implementation.py
gabe-l-hart: I'm not really following this exactly, but would the deduced request here be `union_list.union_list.values` to get at the actual values?

Also, if this can be `None`, it should be marked `Optional`

gabe-l-hart (changes requested):
There are a couple of small fixes that are needed (copyright header, docstring, etc) but I also think we need to look more closely at how the auto-deduction logic creates the intermediate message and if that's actually needed at all.

caikit/runtime/service_generation/protoable.py
prashantgupta24: Since we're not creating the static ones anymore, this doesn't populate the static ones. This only has the custom ones that get created through unions

caikit/runtime/service_generation/protoable.py
prashantgupta24: Oh or you want to return the `Union[tuple(data_object_list)]` as is which becomes the type of the `Union List`? 🤔 Hmm interesting

caikit/runtime/service_generation/protoable.py
prashantgupta24: Added a test case to check for that condition as well (`Union[List], int`)

tests/runtime/service_generation/test_protoable.py
```diff
@@ -87,29 +98,6 @@ def test_to_protoable_signature_no_protoable_types():
     )
```
prashantgupta24, line 100: The tests below have been moved to their own section with no changes

tests/runtime/test_grpc_server.py
```diff
@@ -466,18 +468,24 @@ def test_train_primitive_model(
     train_request_class = DataBase.get_class_for_name(
         "SampleTaskSamplePrimitiveModuleTrainRequest"
     )
-    union_list_str_int_dm = caikit.interfaces.common.data_model.UnionListStrIntSource
-    union_list_str_bool_dm = caikit.interfaces.common.data_model.UnionListStrBoolSource
+    union_list_str_dm = (
```
aluu317, line 471: what is the use of parentheses for these statements?

prashantgupta24: None, but for some reason `fmt` adds them for me :D

prashantgupta24: Actually there is something I can do!

caikit/runtime/service_generation/protoable.py
```diff
@@ -91,6 +99,30 @@ def handle_protoables_in_union(arg_type: Type) -> Type:
     log.debug("Skipping non-protoable argument type [%s]", arg_type)
 
 
+def get_union_list_type(field_name: str, union_protoables: List) -> Type[DataBase]:
+    """Create a union from list type objects"""
+    common_dm_package = caikit.interfaces.common.data_model
+    param_list = []
+    for arg in union_protoables:
+        if get_origin(arg) == list:
```
gabe-l-hart, line 107: NIT

```suggestion
        if get_origin(arg) is list:
```

caikit/runtime/service_generation/protoable.py
```diff
@@ -91,6 +99,30 @@ def handle_protoables_in_union(arg_type: Type) -> Type:
...
+            arg_type = get_args(arg)[0]
```
gabe-l-hart, line 108: I think for pedanticism, we need to validate that `get_args(arg)` is not empty (e.g. `foo: Union[list, int]`) unless we somehow know that this will be filtered out by `is_protoable_type`. If we do know that that code path is impossible, it's worth a `NOTE` comment here to avoid peoples' eyes catching on a potentially dangerous index operation.

caikit/runtime/service_generation/protoable.py
```diff
@@ -91,6 +99,30 @@ def handle_protoables_in_union(arg_type: Type) -> Type:
...
+            arg_name = f"{arg_type.__name__.capitalize()}Sequence"
+            if not hasattr(common_dm_package, arg_name):
+                raise AttributeError(
+                    f"Unable to find {arg_name} in {common_dm_package}"
+                )
+            do = getattr(common_dm_package, arg_name)
```
gabe-l-hart, line 114: Is `do` here `data object`? Total NIT, but it would be better to have a more descriptive name here, or alternately just smash this into an inline call

```py
data_obj = getattr(common_dm_package, arg_name, None)
if dat_obj is None:
    raise AttributeError(...)
param_list.append(
    Annotated[data_obj, ...]
)
```

caikit/runtime/service_generation/protoable.py
```diff
@@ -91,6 +99,30 @@ def handle_protoables_in_union(arg_type: Type) -> Type:
...
+            param_list.append(
+                Annotated[
+                    do,
+                    OneofField(field_name + "_" + arg_type.__name__ + "_" + "sequence"),
+                ]
+            )
+        else:
+            param_list.append(arg)
+    return Union[tuple(param_list)]
```
gabe-l-hart, line 123: Is this portable to all of our supported python versions? I had the same thing [over in http server](https://github.com/caikit/caikit/blob/main/caikit/runtime/http_server.py#L466) and my vscode built-in linting went crazy about it

gabe-l-hart (changes requested):
A couple of little NITs and one place to fix the `package` for the primitive DM objects to not introduce a dependency on `caikit.interfaces.runtime` in `caikit.interfaces.common`. Other than that, this looks great!

gabe-l-hart (approved):
Looks good!

PR #295, opened 2023-06-30
:bug: Fix multi-processing exception handling and train_location_handling fixture

gkumbhat:
### Changes
- Fix `set_train_location` fixture, which was not setting the `sample_train_servicer.use_subprocess` value correctly because:
    1. The format of config had changed
    2. The scope of `sample_train_servicer` was `session`, so the value of use_subprocess was not getting changed by test. Additionally, the GlobalTrainServicer interacts with service generation etc, changing the scope to `function` for this fixture would create errors.
- The `error` in `_ErrorCaptureProcess` was actually never getting set in case of exception. The reason for that is after forking another process, the instance with parent process of `_ErrorCaptureProcess` will not be same as with the child process. So even if child process sets the value of `self.error` it will not get propagated to the parent and thus the `SubProcessTrainExecutor` will never get any value in `self._worker.error`. In this fix, I am using python multiprocessing Pipe mechanism to communicate between child and parent.

### DIFF
--- a/caikit/runtime/work_management/train_executors.py
+++ b/caikit/runtime/work_management/train_executors.py
@@ -199,22 +199,31 @@ class SubProcessTrainSaveExecutor(TrainSaveExecutorBase):
 
         def __init__(self, *args, event=None, **kwargs):
             super().__init__(*args, **kwargs)
-            self.error = None
+            self._error = None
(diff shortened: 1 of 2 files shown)


gkumbhat:
closed in favor of #297

PR #297, opened 2023-07-03
🐛 Fix multi-processing exception handling and train_location_handling fixture

gkumbhat:
### Changes
- Fix `set_train_location` fixture, which was not setting the `sample_train_servicer.use_subprocess` value correctly because:
    1. The format of config had changed
    2. The scope of `sample_train_servicer` was `session`, so the value of use_subprocess was not getting changed by test. Additionally, the GlobalTrainServicer interacts with service generation etc, changing the scope to `function` for this fixture would create errors.
- The `error` in `_ErrorCaptureProcess` was actually never getting set in case of exception. The reason for that is after forking another process, the instance with parent process of `_ErrorCaptureProcess` will not be same as with the child process. So even if child process sets the value of `self.error` it will not get propagated to the parent and thus the `SubProcessTrainExecutor` will never get any value in `self._worker.error`. In this fix, I am using python multiprocessing Pipe mechanism to communicate between child and parent.

### DIFF
--- a/tests/runtime/servicers/test_global_train_servicer_impl.py
+++ b/tests/runtime/servicers/test_global_train_servicer_impl.py
@@ -40,12 +40,11 @@ import caikit.core
 
 
 @pytest.fixture(autouse=True, params=[True, False])
-def set_train_location(request):
+def set_train_location(request, sample_train_servicer):
     """This fixture ensures that all tests in this file will be run with both
     subprocess and local training styles
     """
-    with temp_config({"training": {"use_subprocess": request.param}}):
-        yield
+    sample_train_servicer.use_subprocess = request.param
 
 
 # Train tests for the GlobalTrainServicer class ############################################################
@@ -78,6 +77,7 @@ def test_global_train_sample_task(
     training_response = sample_train_servicer.Train(
         train_request, Fixtures.build_context("foo")
     )
+
     assert training_response.model_name == model_name
 
     assert training_response.training_id is not None
@@ -340,6 +340,8 @@ def test_global_train_Edge_Case_Widget_should_raise_when_error_surfaces_from_mod
     training_data = stream_type(
         jsondata=stream_type.JsonData(data=[SampleTrainingType(1)])
     ).to_proto()
+
+    sample_train_servicer.auto_load_trained_model = False
     train_request = sample_train_service.messages.SampleTaskSampleModuleTrainRequest(
         model_name=random_test_id(),
         batch_size=999,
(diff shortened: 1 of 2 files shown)


tests/runtime/servicers/test_global_train_servicer_impl.py
```diff
@@ -40,12 +40,11 @@
...
 
 @pytest.fixture(autouse=True, params=[True, False])
-def set_train_location(request):
+def set_train_location(request, sample_train_servicer):
     """This fixture ensures that all tests in this file will be run with both
     subprocess and local training styles
     """
-    with temp_config({"training": {"use_subprocess": request.param}}):
-        yield
+    sample_train_servicer.use_subprocess = request.param
```
gabe-l-hart, line 47: I think this is still mutating global state. I think the right way to do this is:

```py
prev_value = sample_train_servicer.use_subprocess
sample_train_servicer.use_subprocess = request.param
yield
sample_train_servicer.use_subprocess = prev_value
```

gabe-l-hart (changes requested):
One fix request on the fixture, but otherwise this looks great. Thanks for finding this Gaurav!

tests/runtime/servicers/test_global_train_servicer_impl.py
```diff
@@ -340,6 +344,8 @@ def test_global_train_Edge_Case_Widget_should_raise_when_error_surfaces_from_mod
     training_data = stream_type(
         jsondata=stream_type.JsonData(data=[SampleTrainingType(1)])
     ).to_proto()
+
+    sample_train_servicer.auto_load_trained_model = False
```
gabe-l-hart, line 348: Shoot, should have caught this one too. I think anything that mutates attributes of `sample_train_servicer` needs to be handled carefully

gabe-l-hart (changes requested):
One more global mutation

tests/runtime/servicers/test_global_train_servicer_impl.py
gkumbhat: ops, this came from my testing, doesn't need here. I'll remove it.

tests/runtime/servicers/test_global_train_servicer_impl.py
```diff
@@ -351,11 +356,11 @@ def test_global_train_Edge_Case_Widget_should_raise_when_error_surfaces_from_mod
             train_request, Fixtures.build_context("foo")
         )
 
-        training_result = sample_train_servicer.training_map.get(
+        _ = sample_train_servicer.training_map.get(
```
gabe-l-hart, line 359: NIT: You can actually just omit the LHS here. Fine as is, though!

gabe-l-hart (approved):
Thanks for the fix Gaurav!

Issue #284, opened 2023-06-28
Update service generation code to look for the additional `@taskmethod`s


joerunde:
The current inference service generation code assumes that a module has a single inference method, and RPCs are built by collecting those inference methods.

Instead, we'll need to take into account all the `@taskmethod`s on each module. RPCs should be generated for each (task, supported streaming flavor) combo

joerunde:
From a very abbreviated discussion on naming conventions, we're going with:

RPCS:
TextGenerationPredict
ClientStreamingTextGenerationPredict
ServerStreamingTextGenerationPredict
BidiStreamingTextGenerationPredict

Messages:
TextGenerationRequest
StreamingTextGenerationRequest

PR #302, opened 2023-07-06
✨  Inject RPC data into service handlers

### LINKED ISSUE #286: Update server wrapper to allow global-predict wrapping of input-streaming flavored RPCs
The server wrapper currently only allows unary inputs to be wrapped by global predict- we'll need to ensure that it can allow streaming-input flavored rpcs the be redirected to global predict, and that all the correct grpc bits are created accordingly.

joerunde:
Closes #286 

This PR modifies the server wrapper to inject a third kwarg into rpc handlers that service the caikit runtime.

This third kwarg is the caikit rpc metadata, i.e. some flavor of `CaikitRPCBase`, that holds valuable information about how the request to and response should be handled. 
This is required for input streaming support for inference, as the inference handler will need to know which fields from the request message stream to direct into a `DataStream` and pass along to the model. This is also now required to know a priori whether or not the response from the model will be a stream, and will need to be handled with the appropriate stream deserializer.

This also allows all flavors of streaming rpcs to be wrapped by the global runtime servicers, but changes for #284 are required before testing that is easy to do.


caikit/runtime/interceptors/caikit_runtime_server_wrapper.py
```diff
@@ -139,6 +143,9 @@ def safe_rpc_call(request, context):
             with alog.ContextLog(log.debug, "[Safe RPC]: %s", rpc.__name__):
                 try:
                     IN_PROGRESS_GAUGE.labels(rpc_name=rpc.__name__).inc()
+                    if caikit_rpc:
+                        # Pass through the CaikitRPCBase rpc description to the global handlers
+                        return rpc(request, context, caikit_rpc)
```
joerunde, line 148: should this be by kwarg only?

prashantgupta24 (approved):

Issue #217, opened 2023-06-09
Required parameters for inference tasks should not be altered during service generation


joerunde:
## Describe the bug

When I have a task that describes a required parameter like so:

```python
@task(
    required_parameters={"sample_input": SampleInputType},
    output_type=OtherOutputType,
)
class OtherTask(TaskBase):
    """Another sample `task` for our test models"""
```

And a module implements that task, where the required parameter is annotated as a type union containing the required type:

```python
def run(
    self, sample_input: Union[SampleInputType, str]
) -> Union[OtherOutputType, str]:
```

Then the service should generate a request message for the task that still only includes the type for the parameter given by the task definition. In this case it should not generate a `oneof` in the proto api for the `sample_input` field.

## Platform

Please provide details about the environment you are using, including the following:

- Interpreter version:
- Library version:

## Sample Code

Please include a minimal sample of the code that will (if possible) reproduce the bug in isolation

## Expected behavior

Parameters specified in task definitions should not have their types altered

## Observed behavior

Task parameters are altered if a type union is given by a module

## Additional context

Our goal is for the task apis to be stable and work across all modules, so it seems a bit more reasonable to only expose the type that we know all modules should accept

PR #303, opened 2023-07-06
♻️ refactor RPCs for streaming

prashantgupta24:
## Motivation
The current inference service generation code assumes that a module has a single inference method, and RPCs are built by collecting those inference methods.

## Changes

This PR takes into account all the @taskmethods on each module. RPCs are generated for each (task, supported streaming flavor) combo!

This PR also generates a request message for the task that only includes the type for the parameter given by the task definition

## Issues fixed
fix https://github.com/caikit/caikit/issues/284
fix https://github.com/caikit/caikit/issues/217

### DIFF
--- a/caikit/core/task.py
+++ b/caikit/core/task.py
@@ -173,10 +173,17 @@ class TaskBase:
                     signature_type
                 ) == typing.Union and parameter_type in typing.get_args(signature_type):
                     continue
-                type_mismatch_errors.append(
-                    f"Parameter {parameter_name} has type {signature_type} but type \
-                        {parameter_type} is required"
-                )
+                if input_streaming and cls._is_iterable_type(parameter_type):
+                    streaming_type = typing.get_args(parameter_type)[0]
+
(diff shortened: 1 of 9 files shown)


caikit/core/task.py
```diff
@@ -173,10 +173,17 @@ def validate_run_signature(
...
-                type_mismatch_errors.append(
-                    f"Parameter {parameter_name} has type {signature_type} but type \
-                        {parameter_type} is required"
-                )
+                if input_streaming and cls._is_iterable_type(parameter_type):
+                    streaming_type = typing.get_args(parameter_type)[0]
+
+                    for iterable_type in typing.get_args(parameter_type):
+                        if cls._subclass_check(iterable_type, streaming_type):
+                            pass
```
joerunde, line 181: I don't think this will ever actually fail if `cls._subclass_check` fails...?

joerunde (approved):
lpgtm, thanks for the PR!

PR #304, opened 2023-07-07
Reversible Hasher

gabe-l-hart:
This will be used to obfuscate the trainer name in the ID given when submitting a training

Supports #308

### DIFF
--- a/caikit/core/toolkit/reversible_hasher.py
+++ b/caikit/core/toolkit/reversible_hasher.py
@@ -25,43 +25,26 @@ WARNING: This utility is intentionally not a cryptographically secure hash! It
 import random
 import string
 
+# Create a statically seeded instance of Random to perform shuffling in a
+# consistent way
+_RAND_INST = random.Random()
+_RAND_INST.seed(42)
+_CHARSET = string.ascii_letters + string.digits + "-_"
+_SHUFFLED = list(_CHARSET)
+_RAND_INST.shuffle(_SHUFFLED)
 
-class _ReversibleHasherMeta(type):
-    """The _ReversibleHasherMeta is used here to ensure that the random state is
-    only temporarily seeded at import time.
-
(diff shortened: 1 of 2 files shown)


caikit/core/toolkit/reversible_hasher.py
```diff
@@ -0,0 +1,67 @@
...
+
+    WARNING: If this class is dynamically imported in a multi-threaded
+        application, other threads that share the random state will be effected
+        by the temporary seed!
+    """
+
+    def __new__(mcs, name, bases, attrs):
+        # Temporarily seed random
+        random_state = random.getstate()
+        random.seed(42)
```
gkumbhat, line 41: nit: can we move 42 as constant at the top ?

caikit/core/toolkit/reversible_hasher.py
```diff
@@ -0,0 +1,67 @@
...
+import string
+
+
+class _ReversibleHasherMeta(type):
+    """The _ReversibleHasherMeta is used here to ensure that the random state is
+    only temporarily seeded at import time.
+
+    WARNING: If this class is dynamically imported in a multi-threaded
+        application, other threads that share the random state will be effected
+        by the temporary seed!
```
gkumbhat, line 35: Is this because of setting the seed via `random.seed` ? Since quite a bit of the applications rely on random number, like in NLP, having a different random seed can have unprecedented side-effects.

Wondering if this can be resolved by using `random.Random` and essentially doing the random-number generation for this class in a more localized manner?

gabe-l-hart: 🤦🤦🤦 Yep, `random.Random` is a MUCH better solution. I was doing all of this metadata junk to try to accomplish what `random.Random` is explicitly designed to do. Will refactor accordingly

gkumbhat (approved):
LGTM

Issue #235, opened 2023-06-14
Improve text-sentiment example


simona-rc:
## Is your feature request related to a problem? Please describe.

The text-sentiment example as is written now can run successfully only if you execute start_runtime.py from the specific directory where start_runtime.py resides. Otherwise, it doesn't load the model and cannot successfully answer inference requests. 
If you want to execute it from a different directory, you need to add to text-sentiment/text_sentiment/config.yml a parameter as the below that includes the relative path from the directory in which you execute to the models directory:
runtime:
   library: text_sentiment
   local_models_dir: text-sentiment/models

I think you should at least add the above information to the text-sentiment example Readme file.
A better solution is to fix the example, so that it doesn't depend on the directory from which you execute start_runtime.py

Thanks, Simona.

PR #307, opened 2023-07-07
Update example and documentation

rafvasq:
Along with a couple of doc edits for readability, this PR addresses #235 to ensure that the text-sentiment example can be followed successfully without specifying a directory to run the `start_runtime.py` script from by overwriting the default value for `local_models_dir` to be a deduced absolute path of the script.

### DIFF
--- a/examples/text-sentiment/start_runtime.py
+++ b/examples/text-sentiment/start_runtime.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 # Standard
-from os import path, chdir
+from os import chdir, path
 import sys
 
 # First Party
--- a/examples/text-sentiment/start_runtime.py
+++ b/examples/text-sentiment/start_runtime.py
@@ -13,13 +13,22 @@
 # limitations under the License.
 
 # Standard
-from os import chdir, path
+from os import path
 import sys
 
 # First Party
 import alog
 
-chdir(path.dirname(__file__))
+# Local
+import caikit
+
+models_directory = path.abspath(path.join(path.dirname(__file__), "models"))
+caikit.config.configure(
+    config_dict={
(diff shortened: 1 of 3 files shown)


examples/text-sentiment/start_runtime.py
```diff
@@ -13,12 +13,14 @@
...
 
 # Standard
-from os import path
+from os import path, chdir
 import sys
 
 # First Party
 import alog
 
+chdir(path.dirname(__file__))
```
gabe-l-hart, line 22: I'm curious why this is needed. If this is patching around the fact that we are expecting some assets to be referenced by a relative path, I think we should fix the places where that's true rather than doing this.

gabe-l-hart (commented):
Thanks for the updates! One question about the `os.chdir`

examples/text-sentiment/start_runtime.py
gabe-l-hart: Ah, so the issue is specifically with the fact that [the default value for `local_models_dir`](https://github.com/caikit/caikit/blob/main/caikit/config/config.yml#L70) is relative and there's no reliable way to ensure that we know the absolute path. I think the better solution then is to overwrite the config default to be the deduced absolute path as follows:

```py
import caikit

models_directory = os.path.abspath(
    os.path.join(os.path.dirname(__file__), "models")
)
caikit.config.configure(
    config_dict={
        "merge_strategy": "merge",
        "runtime": {"local_models_dir": models_directory},
    }
)
```

rafvasq: Made the changes (and learned about `configure()`) 👍

gabe-l-hart (approved):
@rafvasq looks good, thanks for the updates!

Issue #308, opened 2023-07-07
Implement Trainer abstraction in caikit.core.MODEL_MANAGER


gabe-l-hart:
## Is your feature request related to a problem? Please describe.

Similar to how `caikit.core.MODEL_MANAGER.load` is used to load arbitrary models using the [`finder` and `initializer` abstractions](https://github.com/caikit/caikit/blob/main/docs/adrs/020-model-management-abstractions.md), we want a `caikit.core.MODEL_MANAGER.train` (and correspondingly `caikit.train`) that uses the `trainer` abstraction. This will later be used behind the `caikit.runtime` train management endpoints.

## Describe the solution you'd like

```py
model_future = caikit.train(MyModule, first_arg="foo", second_arg=123)
print(model_future.get_status())
# Block until complete
model = model_future.load()
print(model_future.get_status())
```

PR #309, opened 2023-07-07
Destroyables

gabe-l-hart:
## Description

This PR moves the `DestroyableThread` class into `caikit.core.toolkit` so that it can be used in `caikit.train`. It also introduces `DestroyableProcess` and a shared `Destroyable` base interface that allows the two to be used generically.

Supports #308

**Upstream PRs**


caikit/core/toolkit/destroyable_process.py
```diff
@@ -0,0 +1,193 @@
...
+import alog
+
+# Local
+from .destroyable import Destroyable
+
+log = alog.use_channel("DESTROY-PROC")
+
+FORK_CTX = multiprocessing.get_context("fork")
+
+
```
gkumbhat, line 39: It seems we have removed the patch required for python forks to work correctly inside a Threadpoolexecutor. Ref: https://github.com/caikit/caikit/blob/main/caikit/runtime/work_management/train_executors.py#L50

caikit/core/toolkit/destroyable.py
```diff
@@ -0,0 +1,54 @@
...
+    @property
+    @abc.abstractmethod
+    def destroyed(self) -> bool:
+        """Return True if destroy was called, regardless of whether the
+        destroyable was alive at the time
+        """
+
+    @property
+    @abc.abstractmethod
+    def canceled(self) -> bool:
```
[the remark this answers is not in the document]
gabe-l-hart: This was actually my way of getting around the need to expose `events` like you had in your train executors. The crux of that issue was that the function that would call `destroy` couldn't differentiate between being called because the operation had terminated naturally, or because the user had hung up. As a result, `destroy` would be called in either case and the only way to tell the difference was to see if the `cancel_event` was set.

Having `canceled` vs `destroyed` allows us to differentiate this by interface rather than implementation and therefore allows us to not have these events have shared ownership (in fact, events are no longer needed at all).

caikit/core/toolkit/destroyable_process.py
gabe-l-hart: This is actually included in the `LocalTrainer` (next PR): https://github.com/caikit/caikit/pull/310/files#diff-01866f980d3c84ae181021e96cf0ccd560747994902c657fb5fad90e3edde31fR52

caikit/core/model_management/model_trainer_base.py
```diff
@@ -75,14 +114,52 @@ def wait(self):
...
+                *list(path_parts[:-1] + (training_id,) + path_parts[-1:])
+            )
+
     @abc.abstractmethod
     def train(
         self,
         module_class: Type[ModuleBase],
         *args,
         save_path: Optional[str] = None,
+        save_with_id: bool = False,
```
gkumbhat, line 147: Currently we were not using / providing `save_path` to `.train` function when training and essentially considering `save` a totally different and explicit operations. However, this has a limitation that we can't have a way to save or do something with iterations of the model checkpoints that are getting generated at train time. This could be useful for couple of purposes like when training is taking longer time, so we can get subsequent and iterative checkpoints,  these can also be used in some trustworthy AI pipelines, could be a fail safe mechanism, can be used to allow evaluation at multiple epochs at a later stage. 

Since now we are adding `save_path` and `save_with_id` both in overall `.train` function of model-management, do you think this will facilitate the usecases I mentioned above, essentially that allows persisting artifacts at the time of `.train`  function itself?

caikit/core/model_manager.py
```diff
@@ -67,12 +69,119 @@ class ModelManager:
...
+
+        Kwargs:
+            trainer (Union[str, ModelTrainerBase]): The trainer to use. If given
+                as a string, this is a key in the global config at
+                model_management.trainers.
+            save_path (Optional[str]): Path on disk where the model should be
+                saved (may be relative to a remote trainer's filesystem)
+            save_with_id (bool): Inject the training ID into the save path for
+                the output model
+            wait (bool): Wait for training to complete before returning
```
gkumbhat, line 113: question: wait would make the training synchronous right?

caikit/core/model_manager.py
```diff
@@ -67,12 +69,119 @@ class ModelManager:
...
+
+        Returns:
+            model_future (ModelTrainerBase.ModelFutureBase): The future handle
+                to the model which holds the status of the in-flight training.
+        """
+        try:
+            trainer = self._get_trainer(ModelTrainerBase.get_trainer_name(training_id))
+
+        # Fall back to the default trainer to try to find this ID
+        except ValueError:
```
gkumbhat, line 180: relying on `ValueError` seems a bit risky since it can be raised by other value related scenarios, can use value of `ModelTrainerBase.get_trainer_name` to determine if we need to use `default` instead?

caikit/core/toolkit/destroyable_process.py
```diff
@@ -0,0 +1,193 @@
...
+        return self._completion_event
+
+    # NOTE: This functionality is covered by the unit tests, but pytest-cov is
+    #   not correctly collecting the coverage info since it executes inside the
+    #   subprocess
+    def _target_wrapper(self, target, args, kwargs):  # pragma: no cover
+        result = target(*(args or []), **(kwargs or {}))
+        log.debug3("Process target result: %s", result)
+        if self._return_result:
+            self._child_conn.send(result)
```
gkumbhat, line 192: can this create issue since size of the `result` (pickled instance of the trained class) here can be bigger and pipe would have limitation on the size of data transfer, so either it can take quite a while and meanwhile `join` and `close` can create deadlocks ?

gabe-l-hart: Ah, yeah, this is actually explicitly why I added the `return_results` flag: When this class is used for training, [it will NOT return the result at all](https://github.com/caikit/caikit/pull/310/files#diff-01866f980d3c84ae181021e96cf0ccd560747994902c657fb5fad90e3edde31fR104). I added this whole result-returning capability to make this a more generic tool, similar to `DestroyableThread`. Technically, this capability is not immediately needed, but I liked the symmetry and I think it likely will be useful down the road (possibly for inference subprocess wrapping).

gkumbhat: oh gotcha. Make sense. Thanks!

caikit/core/toolkit/destroyable_process.py
```diff
@@ -0,0 +1,193 @@
...
+    ## Impl ##
+
+    def _update_result(self):
+        if self._parent_conn.poll():
+            self.__result = self._parent_conn.recv()
+
+    @property
+    def error(self) -> Optional[Exception]:
+        self._update_result()
+        if isinstance(self.__result, Exception):
```
gkumbhat, line 178: hmm. `self.__result` will not be an instance of Exception since it will happen in the subprocess right?

caikit/runtime/work_management/train_executors.py
```diff
@@ -225,13 +225,15 @@ def error(self):
...
             return self._error
 
-    def __init__(self, cancel_event) -> None:
-
+    def __init__(self):
+        self.__complete_event = multiprocessing.Event()
         self.complete_event = multiprocessing.Event()
-        self.events = [cancel_event, self.complete_event]
-
+        self.__cancel_event = multiprocessing.Event()
```
gkumbhat, line 231: line 230 can be removed?

gkumbhat (commented):
DestroyableProcess would simplify the process executor. Thanks! 

I left some more comments that I missed in previous round. Other than that it looks good to me!

gkumbhat (approved):
Thanks Gabe! I think there is a small change required [here](https://github.com/caikit/caikit/pull/309/files#r1258985026) but other than that, it looks good to me!

caikit/runtime/work_management/train_executors.py
gabe-l-hart: Yep, good catch!

PR #310, opened 2023-07-07
Local trainer

### LINKED ISSUE #308: Implement Trainer abstraction in caikit.core.MODEL_MANAGER

gabe-l-hart:
## Description

This PR adds the `LocalModelTrainer` implementation of the `ModelTrainerBase` abstraction so that `caikit.train` can use it. It is configurable to use either a `DestroyableThread` or a `DestroyableProcess` to perform the asynchronous training job.

Closes #308 

**Upstream PRs**

* https://github.com/caikit/caikit/pull/309

### DIFF
--- a/caikit/core/model_management/local_model_trainer.py
+++ b/caikit/core/model_management/local_model_trainer.py
@@ -101,9 +101,12 @@ class LocalModelTrainer(ModelTrainerBase):
                 log.debug2("Running training %s as a SUBPROCESS", self.id)
                 self._worker = DestroyableProcess(
                     target=self._train_and_save,
-                    return_result=False,
+                    return_result=True,
                     args=args,
-                    kwargs=kwargs,
+                    kwargs={
+                        "return_completion_time": True,
+                        **kwargs,
+                    },
                 )
                 # If training in a subprocess without a save path, the model
                 # will be unreachable once trained!
@@ -169,6 +172,7 @@ class LocalModelTrainer(ModelTrainerBase):
 
             # If running a subprocess, handle abnormal exit codes
             if self._use_subprocess:
+                self._completion_time = self._worker.get_or_throw()
                 if self._worker.exitcode and self._worker.exitcode != os.EX_OK:
                     if self._worker.exitcode == OOM_EXIT_CODE:
                         raise MemoryError("Training process died with OOM error!")
@@ -182,7 +186,6 @@ class LocalModelTrainer(ModelTrainerBase):
             model or raise any errors that happened during training.
             """
             self.wait()
-            result = self._worker.get_or_throw()
             if self._use_subprocess:
                 log.debug2("Loading model saved in subprocess")
                 error.value_check(
@@ -197,11 +200,13 @@ class LocalModelTrainer(ModelTrainerBase):
                     self.save_path,
                 )
                 result = caikit.load(self.save_path)
+            else:
+                result = self._worker.get_or_throw()
             return result
 
         ## Impl ##
 
-        def _train_and_save(self, *args, **kwargs):
+        def _train_and_save(self, *args, return_completion_time=False, **kwargs):
             """Function that will run in the worker thread"""
             with alog.ContextTimer(log.debug, "Training %s finished in: ", self.id):
                 trained_model = self._module_class.train(*args, **kwargs)
@@ -210,26 +215,32 @@ class LocalModelTrainer(ModelTrainerBase):
                 with alog.ContextTimer(log.debug, "Training %s saved in: ", self.id):
                     trained_model.save(self.save_path)
             self._completion_time = datetime.now()
+            log.debug2("Completion time for %s: %s", self.id, self._completion_time)
+            if return_completion_time:
+                return self._completion_time
             return trained_model
 
     ## Interface ##
 
     # Expression for parsing retention policy
     _timedelta_expr = re.compile(
-        r"^((?P<days>\d+?)d)?((?P<hours>\d+?)hr)?((?P<minutes>\d+?)m)?((?P<seconds>\d*\.?\d*?)s)?$"
+        r"^((?P<days>\d+?)d)?((?P<hours>\d+?)h)?((?P<minutes>\d+?)m)?((?P<seconds>\d*\.?\d*?)s)?$"
     )
 
     def __init__(self, config: aconfig.Config, instance_name: str):
         """Initialize with a shared dict of all trainings"""
         self._instance_name = instance_name
-        self._retention_duration = config.get("retention_duration")
         self._use_subprocess = config.get("use_subprocess", False)
+        self._retention_duration = config.get("retention_duration")
         if self._retention_duration is not None:
             try:
+                log.debug2("Parsing retention duration: %s", self._retention_duration)
                 self._retention_duration = timedelta(
                     **{
                         key: float(val)
-                        for key, val in self._timedelta_expr.match("23d")
+                        for key, val in self._timedelta_expr.match(
+                            self._retention_duration
+                        )
                         .groupdict()
                         .items()
                         if val is not None
@@ -238,7 +249,9 @@ class LocalModelTrainer(ModelTrainerBase):
             except AttributeError:
                 error(
                     "<COR63897671E>",
-                    f"Invalid retention_duration: {self._retention_duration}",
+                    ValueError(
+                        f"Invalid retention_duration: {self._retention_duration}"
+                    ),
                 )
 
         # The shared dict of futures and a lock to serialize mutations to it
--- a/caikit/core/model_management/local_model_trainer.py
+++ b/caikit/core/model_management/local_model_trainer.py
@@ -65,7 +65,7 @@ class LocalModelTrainer(ModelTrainerBase):
 
         def __init__(
             self,
-            parent_name: str,
+            trainer_name: str,
             module_class: Type[ModuleBase],
             *args,
             save_path: Optional[str],
@@ -75,7 +75,7 @@ class LocalModelTrainer(ModelTrainerBase):
             **kwargs,
         ):
             super().__init__(
-                parent_name=parent_name,
+                trainer_name=trainer_name,
                 training_id=str(uuid.uuid4()),
                 save_with_id=save_with_id,
                 save_path=save_path,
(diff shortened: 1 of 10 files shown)


caikit/core/model_management/local_model_trainer.py
```diff
@@ -0,0 +1,313 @@
...
+        """Initialize with a shared dict of all trainings"""
+        self._instance_name = instance_name
+        self._retention_duration = config.get("retention_duration")
+        self._use_subprocess = config.get("use_subprocess", False)
+        if self._retention_duration is not None:
+            try:
+                self._retention_duration = timedelta(
+                    **{
+                        key: float(val)
+                        for key, val in self._timedelta_expr.match("23d")
```
gkumbhat, line 230:
```suggestion
                        for key, val in self._timedelta_expr.match(self._retention_duration)
```

gabe-l-hart: Good catch. Fixed with my latest changes

tests/core/toolkit/test_destroyable_thread.py
```diff
@@ -38,11 +38,22 @@ def infinite_wait():
             time.sleep(0.1)
```
joerunde, line 38: While we're here it might be nice to make this a `time.sleep(0.001)` to make the test a bit more snappy?

tests/core/toolkit/test_destroyable_thread.py
```diff
@@ -38,11 +38,22 @@ def infinite_wait():
...
+
+
+def test_threads_canceled_when_interrupt_fails():
+    def long_sleep():
+        time.sleep(0.2)
 
+    thread = DestroyableThread(long_sleep)
+    thread.start()
+    thread.destroy()
+    assert thread.canceled
```
joerunde, line 55: Is this trying to assert that the `canceled` flag is set even when the thread itself is not interrupted?
I wonder if this can be done with a wait on an event so that we don't have to jam a `sleep` in here and have the race conditions of:
- Thread has not started yet when `destroy` is called
- Thread has already finished when `destroy` is called
...and also have to wait the entire `0.2` seconds for the test to finish. That's like a solid 5% of our entire unit test runtime atm

gabe-l-hart: Interesting, good call on using events here. I'll experiment a bit

gabe-l-hart: So I think this is a good approach to eliminating the possibility that this test is flakey, but I also think we need to keep the `sleep(0.2)` in there since the whole point of this test is to simulate `canceled` being true when the thread has not yet died completely. If it's sitting waiting on an event, it will happily raise the internal exception and die as soon as `destroy` is called.

tests/core/model_management/test_local_model_trainer.py
```diff
@@ -0,0 +1,209 @@
...
+    )
+    assert model_future.get_status() == ModelTrainerBase.TrainingStatus.RUNNING
+    assert not model_future.get_status().is_terminal
+
+    # Cancel the future and make sure it reports canceled, even though the
+    # function is still sleeping
+    model_future.cancel()
+    assert model_future.get_status() == ModelTrainerBase.TrainingStatus.CANCELED
+    assert model_future.get_status().is_terminal
+    model_future.wait()
```
joerunde, line 172: Thoughts on not doing `model_future.wait()` here?
This will add a full second to the test runs, and I had just gone and removed all of these sleepy cases from the old training tests to take the suite from 30s to 5s.

If it's possible to add a global teardown fixture that can `.wait()` on a big old list of arbitrary futures that could be nice, I really just don't want this to have to happen serially here and bog everything down

tests/core/model_management/test_local_model_trainer.py
```diff
@@ -0,0 +1,209 @@
...
+
+
+def test_purge_retention_time(trainer_type_cfg):
+    """Test that purging old models works as expected"""
+    trainer = local_trainer(retention_duration="0.1s", **trainer_type_cfg)
+    model_future = trainer.train(SampleModule, DataStream.from_iterable([]))
+    model_future.wait()
+    retrieved_future = trainer.get_model_future(model_future.id)
+    assert retrieved_future is model_future
+    time.sleep(0.2)
```
joerunde, line 193: oof, there's no way that this doesn't end up flaking out at some point though.
Is it possible to change a timestamp here instead to make this thing purge?
e.g.
```
    model_future = trainer.train(SampleModule, DataStream.from_iterable([]))
    model_future.wait()
    retrieved_future = trainer.get_model_future(model_future.id)
    assert retrieved_future is model_future
    model_future._completion_time = now() - 1 day
    with pytest.raises(ValueError):
        trainer.get_model_future(model_future.id)
```

caikit/core/model_management/local_model_trainer.py
```diff
@@ -0,0 +1,328 @@
...
+                log.debug2("Loading model saved in subprocess")
+                error.value_check(
+                    "<COR16745216E>",
+                    self.save_path,
+                    "Unable to load model trained in subprocess without a save_path",
+                )
+                error.value_check(
+                    "<COR59551640E>",
+                    os.path.exists(self.save_path),
+                    "Unable to load model saved in subprocess at {}",
```
joerunde, line 199: If the path doesn't exist I might expect a slightly more specific error message, maybe
"Unable to load model saved in subprocess, path does not exist: {}"

joerunde: Also, should all of these logs contain `self.id` for traceability?

caikit/core/model_management/local_model_trainer.py
```diff
@@ -0,0 +1,328 @@
...
+        save_path: Optional[str] = None,
+        save_with_id: bool = False,
+        external_training_id: Optional[str] = None,
+        **kwargs,
+    ) -> "LocalModelFuture":
+        """Start training the given module and return a future to the trained
+        model instance
+        """
+        # Always purge old futures
+        self._purge_old_futures()
```
joerunde, line 274: Why do these need purging at train time?

caikit/core/model_management/local_model_trainer.py
```diff
@@ -0,0 +1,328 @@
...
+            external_training_id=external_training_id,
+            use_subprocess=self._use_subprocess,
+            **kwargs,
+        )
+
+        # Lock the global futures dict and add it to the dict
+        with self._futures_lock:
+            assert (
+                model_future.id not in self._futures
+            ), f"UUID collision for model future {model_future.id}"
```
joerunde, line 292: I smell a line that will never be unit test covered!

caikit/core/model_management/local_model_trainer.py
gabe-l-hart: Hm, good question. I think I threw that in there just to say "any time the map is touched, purge," but I think you're right that this should be removed

caikit/core/model_management/local_model_trainer.py
gabe-l-hart: That's why it's an `assert`! It _will_ be covered, but the branch will never be followed...

joerunde: > Only after generating 1 billion UUIDs every second for the next 100 years, the probability of creating just one duplicate would be about 50%. Or, to put it another way, the probability of one duplicate would be about 50% if every person on earth owned 600 million UUIDs.

I don't think we'll need this line, but I won't block the PR on it existing

gabe-l-hart: IIRC, the real reason I added this was the presence of `external_training_id`. We have no control over how those are generated, so it could be `lambda: "foo"`

caikit/core/model_management/local_model_trainer.py
```diff
@@ -0,0 +1,328 @@
...
+    name = "LOCAL"
+
+    class LocalModelFuture(ModelTrainerBase.ModelFutureBase):
+        """A local model future manages an execution thread for a single train
+        operation
+        """
+
+        def __init__(
+            self,
+            parent_name: str,
```
joerunde, line 68: unclear how many docstrings we need on this but `parent_name` is a bit opaque here.
Is this just the name of the trainer for the reversible hashing bit of the id? Would `trainer_name` be more understandable?

gabe-l-hart: Yeah, this is just the name of the trainer. I'll make that change (and in the base class)

tests/core/model_management/test_local_model_trainer.py
gabe-l-hart: I'm fine with that. It will wrap up in the background

tests/core/model_management/test_local_model_trainer.py
gabe-l-hart: Yep, much cleaner

tests/core/toolkit/test_destroyable_thread.py
gabe-l-hart: I'll go ahead and remove the `assert thread.is_alive()` as that's the one that could fail as currently written if the races come up

caikit/core/model_management/local_model_trainer.py
gabe-l-hart: Actually, I think the reason I did it here was so that we could guarantee that the in-memory dict wouldn't grow without bound. This is all pretty far-fetched since this implementation won't be used in a production environment anyway, but if we only purge on lookup, we could theoretically just keep submitting training jobs and having the map keep growing.

joerunde (approved):
lpgtm, thanks for cleaning up some sleeps!

gkumbhat (approved):
LGTM. Thanks Gabe!

Issue #311, opened 2023-07-07
Use caikit.train in caikit.runtime


gabe-l-hart:
## Is your feature request related to a problem? Please describe.

Once `caikit.train` is fully supported (#308), we want to use it in `caikit.runtime` when performing training jobs. This will replace the current [train_executors](https://github.com/caikit/caikit/blob/main/caikit/runtime/work_management/train_executors.py) and [training_manager](https://github.com/caikit/caikit/blob/main/caikit/runtime/model_management/training_manager.py).

## Describe the solution you'd like

* Use `caikit.train` in [global_train_servicer](https://github.com/caikit/caikit/blob/main/caikit/runtime/servicers/global_train_servicer.py#L180)
* Use `caikit.get_model_future` in [training_management_servicer](https://github.com/caikit/caikit/blob/main/caikit/runtime/servicers/training_management_servicer.py#L38)

PR #312, opened 2023-07-07
Train futures in runtime

gabe-l-hart:
## Description

This PR overhauls the `caikit.runtime` servicers to use `caikit.train` and `caikit.get_model_future`, thus delegating responsibility for all training management to the `ModelTrainer` abstraction.

Cloases: #311 

**Upstream PRs**

* https://github.com/caikit/caikit/pull/309
* https://github.com/caikit/caikit/pull/310


caikit/runtime/servicers/global_train_servicer.py
```diff
@@ -59,20 +55,8 @@ class GlobalTrainServicer:
     def __init__(self, training_service: ServicePackage):
         self._training_service = training_service
         self._model_manager = ModelManager.get_instance()
```
joerunde, line 57: Not related to this PR but this class really needs to get renamed

caikit/runtime/servicers/global_train_servicer.py
```diff
@@ -115,31 +104,30 @@ def Train(self, request, context, *_, **__) -> TrainingJob:
...
             with alog.ContextLog(log.debug, outer_scope_name):
+                module = None
                 for mod in caikit.core.registries.module_registry().values():
                     if mod.TASK_CLASS:
                         train_request_for_mod = (
                             ModuleClassTrainRPC.module_class_to_req_name(mod)
                         )
                         if train_request_for_mod == desc_name:
-                            model = mod
+                            module = mod
```
joerunde, line 114: nit that totally doesn't have to be done now but the `ModuleClassTrainRPC` for this particular request is passed in as `caikit_rpc` so this entire block can be replaced with `module = caikit_rpc.module`

(I think it might actually be called `caikit_rpc.clz` right now though, which, 🤢 )

caikit/runtime/servicers/training_management_servicer.py
```diff
@@ -32,16 +38,32 @@ class TrainingManagementServicerImpl:
     """This class contains the implementation of all of the RPCs that are required to run a
     service in Model Mesh as a Model-Runtime."""
 
-    def __init__(self):
-        self.training_manager = TrainingManager.get_instance()
+    # TODO: Align training status enums!
```
joerunde, line 41: yeah... do we want to leave two status enums hanging around or just expose the `ModelTrainerBase.TrainingStatus` enum as the external status enum?

joerunde: I'm all for not exposing internal state on external apis but this is just a single enum and I doubt it'll bloat up into way more states, given our previous experience with similar ml training systems

caikit/runtime/servicers/global_train_servicer.py
joerunde: oof, actually this does pull in the `MODEL_MANAGER` now too so it is a bit confusing to see `ModelManager` and `MODEL_MANAGER` and have them be completely different things.

`ModelManager` -> `RuntimeModelCache`?

caikit/runtime/servicers/training_management_servicer.py
```diff
@@ -32,16 +38,32 @@ class TrainingManagementServicerImpl:
...
+                training_id=training_info.training_id,
+                status=self.STATUS_MAPPING[model_future.get_status()].value,
+            ).to_proto()
+        except ValueError as err:
+            raise CaikitRuntimeException(
+                grpc.StatusCode.NOT_FOUND,
+                "{} not found in the list of currently running training jobs".format(
+                    training_info.training_id,
+                ),
+            ) from err
```
joerunde, line 69: Is it time to implement a `def CancelTraining` here as well and put it in the training management api?

tests/runtime/test_grpc_server.py
```diff
@@ -556,6 +577,9 @@ def test_train_fake_module_ok_response_with_datastream_csv_file(
     is_good_train_response(
         actual_response, HAPPY_PATH_TRAIN_RESPONSE, model_name, training_management_stub
     )
+    register_trained_model(
+        runtime_grpc_server, actual_response.model_name, actual_response.training_id
+    )
```
joerunde, line 582: oof, sorry about all these changes coming in from killing `autoload`

joerunde (commented):
lpgtm, I think we should probably just go ahead and clean up that module lookup in the trainer to use the passed-in rpc and make sure that's all hooked up correctly.

I'll take another super quick peek once the local trainer pr is merged and this is self-contained

caikit/runtime/servicers/global_train_servicer.py
gabe-l-hart: Ug, yeah, I really don't like this and it would be good go fix. Then again, I'm not sure I want to rock the boat that much (and I _just_ started another branch that touches `runtime/model_management/model_manager.py` 😢)

caikit/runtime/servicers/global_train_servicer.py
gabe-l-hart: Making this change breaks a lot of tests (though it seems to work in the end-to-end ones that aren't testing the function directly). I'll make an issue to clean it up later

gabe-l-hart: https://github.com/caikit/caikit/issues/326

caikit/runtime/servicers/training_management_servicer.py
gabe-l-hart: Yeah, this is the grossest corner. I think we should probably align on a common enum and use it everywhere. I'll make that change

tests/runtime/conftest.py
```diff
@@ -235,6 +237,29 @@ def other_task_model_id(other_good_model_path) -> str:
...
+):
+    """Helper to auto-load a model that has completed training. This replaces
+    the old auto-load feature which was only needed for unit tests
+    """
+    model_future = MODEL_MANAGER.get_model_future(training_id)
+    model = model_future.load()
+    loaded_model = (
+        LoadedModel.Builder()
+        .id(model_id)
+        .type("trained")
```
joerunde, line 253: oh this is yet another gross corner where `model_type` is still just some user defined string :(

This information _should_ be all contained in the `model` here, but this is also only relevant for model-mesh so probably not worth refactoring it out right now

joerunde (approved):
lpgtm, fair on punting those changes. I can probably go get them in on a branch based from here

caikit/runtime/servicers/global_train_servicer.py
```diff
@@ -177,57 +165,49 @@ def Train(self, request, context, *_, **__) -> TrainingJob:
...
+        # Figure out where this model will be saved
+        model_path = self._get_model_path(training_output_dir, request.model_name)
+
+        # Build the full set of kwargs for the train call
+        kwargs.update(
+            {
+                "module": module,
+                "save_path": model_path,
+                "save_with_id": True,
+                **build_caikit_library_request_dict(request, module.TRAIN_SIGNATURE),
```
prashantgupta24, line 205: doesn't this assume that the `request` type is a `ProtoMessageType`? How will this work with REST calls?

prashantgupta24: On the inference side, the call to `Predict` does the `build_caikit_library_request_dict`, which creates kwargs and passes to `predict_model`.

`predict_model` takes in these kwargs which are used to actually call the `model_run_fn`. Should we do the same thing here?

caikit/runtime/servicers/global_train_servicer.py
gabe-l-hart: I'm going to hold on this since I think renaming this would be an additional big change and it _might_ intersect with some of our composition/orchestration discussions.

caikit/runtime/servicers/training_management_servicer.py
gabe-l-hart: Fixed

caikit/runtime/servicers/training_management_servicer.py
gabe-l-hart: Ooh, good call. Let's do that as a next story

gabe-l-hart: https://github.com/caikit/caikit/issues/332

tests/runtime/test_grpc_server.py
gabe-l-hart: Eh, it was the right thing to do!

tests/runtime/conftest.py
gabe-l-hart: Hm, good point. That's a good one to clean up down the road

caikit/runtime/servicers/global_train_servicer.py
gabe-l-hart: Hm, good point! Let me see if that can be refactored to match GPS better

gabe-l-hart: I looked into this a bit and I think I'm going to punt it. Shifting this out of `run_training_job` requires `ModelTrainServicer` to also use `build_caikit_library_request_dict` which seems to have some nuance. Due to the `DataStreamSource` stuff, this is a bit complex and does need to be coerced into the data model object to work cleanly rather than just a pile of `**kwargs`. I think we'll want to dig into this as part of getting train working with REST.

prashantgupta24: Sounds good!

Issue #305, opened 2023-07-07
Flakey WIP decorator test


gabe-l-hart:
## Describe the bug

This test fails randomly with `RuntimeError: Event loop is closed` which triggers [this assertion](https://github.com/caikit/caikit/blob/main/tests/core/toolkit/test_wip_decorator.py#L180) to fail

## Platform

Running in Github Actions. Seems to be independent of python version

## Observed behavior

Example failed run: https://github.com/caikit/caikit/actions/runs/5488998385/jobs/10002586022?pr=304

gabe-l-hart:
[logs_2080.zip](https://github.com/caikit/caikit/files/11984510/logs_2080.zip)

PR #313, opened 2023-07-07
FlakeyWipTest: Convert tests to pytest and loosen regex match checks

gabe-l-hart:
I'm pretty sure that these sporatic errors are caused by the introduction of pytest-asyncio as a test plugin. The tests themselves were explicitly checking caplog.records[0], so if some other warning pops up during the capture, the regex check was failing. Since the consistent error has to do with the asyncio pool being closed, I suspect this is just a bad coincidence plus a test that's assuming it's more isolated than it is.

Closes https://github.com/caikit/caikit/issues/305

### DIFF
--- a/tests/core/toolkit/test_wip_decorator.py
+++ b/tests/core/toolkit/test_wip_decorator.py
@@ -14,11 +14,13 @@
 
 # Standard
 import logging
+import re
 
 # Third Party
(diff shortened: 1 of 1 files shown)


evaline-ju (approved):
LGTM! Thanks for fixing

Issue #316, opened 2023-07-10
Handle output streams based on rpc metadata


joerunde:
See comment on https://github.com/caikit/caikit/pull/314

PR #317, opened 2023-07-10
🐛 fix output streams based on rpc metadata

prashantgupta24:
fix https://github.com/caikit/caikit/issues/316


joerunde (approved):
super slick!

Issue #323, opened 2023-07-12
http-server in the example/text-sentiment returns "Model ... not loaded" 


kingaj12:
## Describe the bug

I am experimenting with the new http_server (thanks @gabe-l-hart!) by replacing the `grpc_server.main()` call in the text-sentiment example server `examples/text-sentiment/start_runtime.py`.

The api from the swagger interface `127.0.0.1:8080/docs` asks for a `model_id`.  I have tried the `module_id` from the file `examples/text-sentiment/models/text_sentiment/config.yml` and also the `module_id` from the `caikit_huggingface_demo/example_models/sentiment/config.yml`.  I also just tried putting in random integers and strings.  But the server reports model not loaded.  

## Platform

Please provide details about the environment you are using, including the following:

- Interpreter version: Python 3.11.4
- Library version: main

## Sample Code

Replace `grpc_server.main()` with `http_server.main()` in the line 
https://github.com/caikit/caikit/blob/5644440249c30bc62aede294bde1860b703e58f6/examples/text-sentiment/start_runtime.py#L41

## Expected behavior

Executing the curl command with the id from the text-sentiment config file should return `200`
```
curl -X 'POST' \
  'http://127.0.0.1:8080/api/v1/8f72161-c0e4-49b0-8fd0-7587b3017a35/task/hugging-face-sentiment' \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -d '{
  "inputs": {
    "text_input": {
      "text": "I am happy!"
    }
  }
}'

```
## Observed behavior
The curl command returns

```

{
    "details": "Model '8f72161-c0e4-49b0-8fd0-7587b3017a35' not loaded", 
    "code": 404, 
    "id": "68a6c0fb0cc24e6b89be0c6cfa4d6eb4"
}

```

## Additional context

I guess it's not really clear to me what `model_id` is asking for.

kingaj12:
It looks like the `model_id` expected is `text_sentiment`.   But now there is another issue:
```

curl -X 'POST' \
  'http://127.0.0.1:8080/api/v1/text_sentiment/task/hugging-face-sentiment' \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -d '{
  "inputs": {
    "text_input": {
      "text": "I am happy"
    }
  }
}'
```
returns a Null response:
```

{
  "classes": [
    {
      "class_name": null,
      "confidence": null
    }
  ]
}
```

[...]

PR #327, opened 2023-07-12
:sparkles: add http example

joerunde:
For the non-bug part of #323 

This adds a `--protocol` flag to the example scripts to run the http server as well.
This does not yet fix the bug in the http server itself, where the response for this huggingface model is not returned. I hope to get to that soon, just wanted to get this example out here and then come in with a fix and tests.

### DIFF
--- a/examples/text-sentiment/client.py
+++ b/examples/text-sentiment/client.py
@@ -13,30 +13,54 @@
 # limitations under the License.
 
 # Third Party
+from start_runtime import protocol_arg
 import grpc
+import requests
 
 # Local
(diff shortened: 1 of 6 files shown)


examples/text-sentiment/client.py
```diff
@@ -13,30 +13,54 @@
 # limitations under the License.
 
 # Third Party
+from start_runtime import protocol_arg
```
gabe-l-hart, line 16: Why isn't this using a relative import? Strange that it works

joerunde: there's no actual package to import! (no __init__.py in examples/text-sentiment)

PR #330, opened 2023-07-13
🐛  Pydantic bug

### LINKED ISSUE #323: http-server in the example/text-sentiment returns "Model ... not loaded"

joerunde:
depends on https://github.com/caikit/caikit/pull/327

fixes #323 

This adds enum and list support to the rest server, so that it can return data models with enums or lists!

### DIFF
--- a/caikit/runtime/http_server.py
+++ b/caikit/runtime/http_server.py
@@ -36,7 +36,10 @@ import pydantic
 import uvicorn
 
 # First Party
-from py_to_proto.dataclass_to_proto import Annotated, get_origin  # Imported here for 3.8 compat
+from py_to_proto.dataclass_to_proto import (  # Imported here for 3.8 compat
+    Annotated,
+    get_origin,
+)
 import aconfig
 import alog
 
--- a/caikit/runtime/http_server.py
+++ b/caikit/runtime/http_server.py
@@ -18,7 +18,7 @@ API based on the task definitions available at boot.
 """
 # Standard
 from functools import partial
-from typing import Iterable, List, Optional, Type, Union, get_args, get_origin
+from typing import Iterable, List, Optional, Type, Union, get_args
 import asyncio
 import enum
 import json
@@ -36,7 +36,7 @@ import pydantic
 import uvicorn
 
 # First Party
-from py_to_proto.dataclass_to_proto import Annotated  # Imported here for 3.8 compat
+from py_to_proto.dataclass_to_proto import Annotated, get_origin  # Imported here for 3.8 compat
 import aconfig
 import alog
 
--- a/caikit/runtime/http_server.py
+++ b/caikit/runtime/http_server.py
@@ -233,7 +233,7 @@ class RuntimeHTTPServer(RuntimeServerBase):
         # pylint: disable=unused-argument
         async def _handler(
             model_id: str, request: pydantic_request, context: Request
-        ) -> pydantic_response | Response:
+        ) -> Union[pydantic_response, Response]:
             log.debug("In unary handler for %s for model %s", rpc.name, model_id)
             loop = asyncio.get_running_loop()
             request_kwargs = {
--- a/caikit/runtime/http_server.py
+++ b/caikit/runtime/http_server.py
@@ -422,6 +422,7 @@ class RuntimeHTTPServer(RuntimeServerBase):
         return dm_obj
 
     @classmethod
+    # pylint: disable=too-many-return-statements
     def _get_pydantic_type(cls, field_type: type) -> type:
         """Recursive helper to get a valid pydantic type for every field type"""
         if get_origin(field_type) is Annotated:
--- a/caikit/runtime/http_server.py
+++ b/caikit/runtime/http_server.py
@@ -16,13 +16,13 @@ This module holds the implementation of caikit's primary HTTP server entrypoint.
 The server is responsible for binding caikit workloads to a consistent REST/SSE
 API based on the task definitions available at boot.
 """
-import enum
 # Standard
 from functools import partial
 
 # Standardfrom functools import partial
-from typing import Iterable, Optional, Type, Union, get_args, get_origin, List
+from typing import Iterable, List, Optional, Type, Union, get_args, get_origin
 import asyncio
+import enum
 import json
 import re
 import ssl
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -79,6 +79,7 @@ dev-fmt = [
     "pre-commit>=3.0.4,<4.0",
     "pylint>=2.16.2,<3.0",
     "pydeps==1.12.10",
+    "requests==2.31.0"
```
prashantgupta24, line 82: any reason for having this version here instead of `requests>=2.28.2,<3",` in `dev-test`?

joerunde: it's just needed for linting now, I can go ahead and throw `dev-test` into the bucket of extras for linting

joerunde: (though this bit is from the branch PRed in #327)

tests/runtime/test_http_server.py
```diff
@@ -267,6 +271,42 @@ def test_model_not_found():
...
+
+def test_pydantic_wrapping_with_lists():
+    """Check that pydantic wrapping works on data models with lists"""
+
+    @dataobject(package="test.test")
+    class Bar(DataObjectBase):
+        baz: int
+
+    @dataobject(package="test.test")
+    class Foo(DataObjectBase):
```
prashantgupta24, line 298: Uhh the error (from what I understand) is that we're trying to use `class Foo` within the same descriptor pool as other tests, specifically `test_data_stream_source` (we create a different descriptor pool for each of the the `dataobject` tests).

prashantgupta24: I guess a simple way would be to just rename this class, the ideal way would be to make the `temp_dpool` as a fixture that can be reused by multiple test files?

prashantgupta24: Not worth fixing for this test imo, although I _would_ like to do that as an extra action item so that we can freely declare same-name `dataobjects` in the future in other test classes?

joerunde: @prashantgupta24 yeah, it would also be nice if py-to-proto allowed us to do this by package instead of requiring a temporary descriptor pool. I think this is potentially a problem if two dataobjects in different packages have the same name. I don't know if I'm convinced that we should say that datamodel classes need to have globally unique names.

prashantgupta24: Yeah that's a good point as well!

caikit/runtime/http_server.py
```diff
@@ -293,7 +290,7 @@ async def _handler(
         ) -> EventSourceResponse:
             log.debug("In streaming handler for %s", rpc.name)
             request_kwargs = {
-                field: getattr(request, field) for field in request.__fields__
+                field: getattr(request, field) for field in request.model_fields
```
prashantgupta24, line 293: could this be just `request_kwargs = dict(request)`?

prashantgupta24: intentionally not call `model_dump` coz we  need flat dict

joerunde: probably? I was just getting rid of a deprecation warning here and I want to limit the scope of this PR

prashantgupta24: cool, either way I already have that change in my rest-train PR 👍

joerunde: hot

prashantgupta24:
Issue: In py38, 
for `field_type`
```
typing_extensions.Annotated[caikit.interfaces.nlp.data_model.text_generation.GeneratedToken, 1]
```
I get:
```
get_origin(field_type)
<class 'caikit.interfaces.nlp.data_model.text_generation.GeneratedToken'>
```

joerunde:
ah thanks @prashantgupta24 , that is mildly gross but definitely required 👍

gabe-l-hart (approved):
Thanks for the fix!

caikit/runtime/http_server.py
```diff
@@ -229,11 +230,11 @@ def _add_unary_input_unary_output_handler(self, rpc: CaikitRPCBase):
             self._get_response_dataobject(rpc)
         )
 
-        @self.app.post(self._get_route(rpc))
+        @self.app.post(self._get_route(rpc), response_model=pydantic_response)
         # pylint: disable=unused-argument
         async def _handler(
             model_id: str, request: pydantic_request, context: Request
-        ) -> pydantic_response:
+        ) -> Union[pydantic_response, Response]:
```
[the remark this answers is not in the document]
joerunde: The error return case returns a plain old `Response` which isn't of type `pydantic_response`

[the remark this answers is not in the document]
gabe-l-hart: Gotcha! Thanks

Issue #334, opened 2023-07-14
Make it possible to override the domain and package names


sboagibm:
I have an older version of a gRPC client that I want backwards compatibility with.  If I can control the domain and package names, I can achieve that.

So I'd like to be able to do:
Example:
```
runtime:
    service_generation:
        package: whatever.runtime.foofoo.v0
        domain: WhatEver
```

Changes would need to be made in `caikit.runtime.service_factory.ServicePackageFactory.get_service_package`.

(PR is coming)

PR #335, opened 2023-07-14
feat: Make it possible to override the domain and package names

sboagibm:
In order to achieve backwards compatibility with an earlier gRPC client, I need to be able to override the domain and package names in the config file:

Example:
```
runtime:
    service_generation:
        package: whatever.runtime.foofoo.v0
        domain: WhatEver
```

Fixes https://github.com/caikit/caikit/issues/334

### DIFF
--- a/tests/runtime/test_service_factory.py
+++ b/tests/runtime/test_service_factory.py
@@ -13,15 +13,18 @@
 # limitations under the License.
 
 """Unit tests for the service factory"""
+import tempfile
+from pathlib import Path
 
-# Third Party
-import pytest
+import caikit
+from caikit.runtime.dump_services import dump_grpc_services
 
 # Local
 from caikit.runtime.service_factory import ServicePackageFactory
 from sample_lib.modules.sample_task import ListModule
 from tests.conftest import temp_config
-import caikit
+
+# Third Party
 
 ### Private method tests #############################################################
 
@@ -31,6 +34,7 @@ MODULE_LIST = [
     for module_class in caikit.core.registries.module_registry().values()
     if module_class.__module__.partition(".")[0] == "sample_lib"
 ]
+
 
 ### Test ServicePackageFactory._get_and_filter_modules function
 def test_get_and_filter_modules_respects_excluded_task_type():
@@ -56,7 +60,6 @@ def test_get_and_filter_modules_respects_excluded_modules():
             }
         }
     ) as cfg:
-
         clean_modules = ServicePackageFactory._get_and_filter_modules(cfg, "sample_lib")
         assert "ListModule" not in str(clean_modules)
         assert "SampleModule" in str(clean_modules)
@@ -75,7 +78,6 @@ def test_get_and_filter_modules_respects_excluded_modules_and_excluded_task_type
             }
         }
     ) as cfg:
-
         clean_modules = ServicePackageFactory._get_and_filter_modules(cfg, "sample_lib")
         assert "ListModule" not in str(clean_modules)
         assert "OtherModule" not in str(clean_modules)
@@ -93,7 +95,6 @@ def test_get_and_filter_modules_respects_included_modules_and_included_task_type
             }
         }
     ) as cfg:
-
         clean_modules = ServicePackageFactory._get_and_filter_modules(cfg, "sample_lib")
         assert len(clean_modules) == 2
         assert "OtherModule" in str(clean_modules)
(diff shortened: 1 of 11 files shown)


sboagibm:
(Sorry about the type-tweak revert commits.)

joerunde (commented):
Thanks @sboagibm!

It'd be nice to get an example usage of this in somewhere, either in `tests/fixtures/config` or `tests/fixtures/sample_lib/config`, and it would be super great if there were one small unit test for this so that we don't end up breaking this accidentally in the future.

Or, maybe two unit tests since it looks like you can specify a `domain` but not a `package` to get `caikit.runtime.{domain}`, or you can specify just a package to override the whole thing

tests/runtime/test_service_factory.py
```diff
@@ -146,7 +148,146 @@ def test_get_and_filter_modules_respects_included_task_types_and_excluded_module
             }
         }
     ) as cfg:
-
         clean_modules = ServicePackageFactory._get_and_filter_modules(cfg, "sample_lib")
         assert "SampleModule" in str(clean_modules)
         assert "ListModule" not in str(clean_modules)
+
+
+def get_packages_with_override(
```
gabe-l-hart, line 156: NIT: We usually try to co-locate helper files like this into a `## Helpers ##` section header at the top

gabe-l-hart (dismissed):
Thanks Scott! One NIT suggestion to move the utility function, but I'm good with this any time

sboagibm:
@gabe-l-hart @joerunde I'm really pulling my hair out, and need help.

The problem is the new tests run fine when run individually, the when run together they get [this error](https://github.com/caikit/caikit/actions/runs/5763354156/job/15624969518?pr=335#step:5:307):

```python
            if not _are_same_file_descriptors(fd_proto, existing_proto):
                # NOTE: This is a TypeError because that is what you get most of the time when you
                # have conflict issues in the descriptor pool arising from JTD to Proto followed by
                # importing differing defs for the same top level message type using different file
                # names (i.e., skipping this validation) compiled by protoc. Raising TypeError here
                # ensures that we at least usually raise the same error type regardless of
                # import / operation order.
>               raise TypeError(
                    f"Cannot add new file {fd_proto.name} to descriptor pool, file already exists with different content"
                )
E               TypeError: Cannot add new file sampletaskrequest.proto to descriptor pool, file already exists with different content
```

Which seems to be a problem that it isn't cleaning up from the call to `ServicePackageFactory.get_service_package(service_type)`.  I tried writing a function to clean up:

```python
def clear_sample_lib():
    import caikit.core.registries

    """Clear the registry of side effects from creating the sample_lib package(s)"""

    module_ids = [
        module_id
        for module_id, module_class in caikit.core.registries.module_registry().items()
        if module_class.__module__.partition(".")[0] == "sample_lib"
    ]
    for module_id in module_ids:
        del caikit.core.registries.module_registry()[module_id]
```

Which _wants_ to work, but in the next test the `clean_modules = ServicePackageFactory._get_and_filter_modules` fails, so clearly the deletion from the registry is not enough, and, anyway, why would I need to go through this much trouble if that's what you're supposed to do?  I was wishing there was a simple `ServicePackageFactory.clear_service_package(...)` inverse of `ServicePackageFactory.get_service_package(...)` since clearly the latter has side effects.  

The code is complex enough that it would take me another day to figure out how to get this to pass, and I have this sneaking suspicion that I'm just doing something dumb.  

Please advise!

gabe-l-hart:
Oof, yeah, this is a really annoying corner of `protobuf`. Everything is held in a `DescriptorPool` which requires that file names be globally unique. There are APIs for creating ephemeral descriptor pools, but all generated protos use a shared global `DEFAULT_POOL`. The key in unit tests is to use a temporary descriptor pool. You can do this in two ways:

1. You can manually create an instance and pass it to all of the `py_to_proto` functions (like is done [here](https://github.com/IBM/py-to-proto/blob/main/tests/conftest.py#L25))
2. If you need to rely on using the default pool (which I believe we do with `@dataobject`), you need to do the above and then patch the global one like [here](https://github.com/caikit/caikit/blob/main/tests/core/data_model/test_dataobject.py#L56)

In your case, we probably want (2) which means that `fixture` should probably get hoisted to `tests/conftest.py`, though if that's done we probably need an `autouse=False` version so that other test modules don't automatically opt into it.

joerunde:
We have also been talking about fixing up py-to-proto to package-scope the generated files so that when you change the package name, it doesn't fail on a file collision

PR #336, opened 2023-07-14
Lazy runtime model load 324

gabe-l-hart:
## Description

This PR adds a new feature to `caikit.runtime` to use a local directory on disk (which can be mounted) to act as a shared-state cache. This will allow replicas of `caikit.runtime` to host an eventually-consistent set of models that changes dynamically. This is a significantly weaker value proposition than `model-mesh`, but for many model types, especially those that use a `module_backend` to delegate the work to another engine (e.g. `caikit-tgis-backend`), it's fine to require that all models be loaded on all replicas.

### DIFF
--- a/caikit/runtime/http_server.py
+++ b/caikit/runtime/http_server.py
@@ -451,7 +451,7 @@ class RuntimeHTTPServer(RuntimeServerBase):
         if get_origin(field_type) is Annotated:
             return cls._get_pydantic_type(get_args(field_type)[0])
         if get_origin(field_type) is Union:
-            return Union.__getitem__(
+            return Union.__getitem__(  # pylint: disable=unnecessary-dunder-call
                 tuple(
                     (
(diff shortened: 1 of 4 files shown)


caikit/runtime/http_server.py
```diff
@@ -451,7 +451,7 @@ def _get_pydantic_type(cls, field_type: type) -> type:
         if get_origin(field_type) is Annotated:
             return cls._get_pydantic_type(get_args(field_type)[0])
         if get_origin(field_type) is Union:
-            return Union.__getitem__(
+            return Union.__getitem__(  # pylint: disable=unnecessary-dunder-call
```
gabe-l-hart, line 454: This one always failed for me locally, but didn't seem to fail on Actions

gabe-l-hart:
Closing in favor of https://github.com/caikit/caikit/pull/341

PR #338, opened 2023-07-17
♻️ align REST with HF API

prashantgupta24:


examples/text-sentiment/client.py
```diff
@@ -12,16 +12,27 @@
...
 import grpc
 import requests
 
 # Local
 from caikit.runtime.service_factory import ServicePackageFactory
-from text_sentiment.data_model import TextInput
+import caikit
 
 if __name__ == "__main__":
+    caikit.config.configure(
```
prashantgupta24, line 28: For some reason I wasn't able to run the client code without this snippet. Without this, it was trying to find `sample_lib`? I'm hoping it's a config issue on my end

caikit/runtime/http_server.py
```diff
@@ -237,16 +237,22 @@ async def _handler(
...
             # would have been useful to call dataobject.to_dict()
             # but unfortunately we now have converted pydantic objects
             combined_dict = {}
             for field in request_kwargs:
                 if request_kwargs[field]:
-                    combined_dict.update(**dict(request_kwargs[field]))
+                    if field == REQUIRED_INPUTS_KEY and input_name:
+                        combined_dict.update({input_name: request_kwargs[field]})
+                    else:
+                        combined_dict.update(**dict(request_kwargs[field]))
```
prashantgupta24, line 255: to test this thoroughly, we should ideally have a module that has multiple required params?

examples/text-sentiment/client.py
joerunde: @prashantgupta24 previously there was a `from text_sentiment.data_model import TextInput` bit that would import the `text_sentiment` "library" which would inject its configuration.

I like this explicit configuration better

caikit/runtime/http_server.py
joerunde: probably!
For now I'm fine with leaving things only tested with 1 required input parameter and punting to a discussion about what that api should look like. I bet it'll end up looking suspiciously like what you have implemented here, and in the meantime our stance can be "only tasks with 1 required input have HF-aligned rest support"

tests/runtime/test_http_server.py
```diff
@@ -229,7 +229,7 @@ def test_inference(sample_task_model_id):
         assert json_response["greeting"] == "Hello world"
 
 
-def test_inference_optional_field(sample_task_model_id):
+def test_inference_sample_task_optional_field(sample_task_model_id):
```
joerunde, line 232: should this test change the `inputs` bit to be
`"inputs": {"name": "world"},`
in order to make sure that that bit is correct and isn't causing the error?

tests/runtime/test_http_server.py
```diff
@@ -215,11 +215,11 @@ def test_docs():
         assert response.status_code == 200
 
 
-def test_inference(sample_task_model_id):
+def test_inference_sample_task(sample_task_model_id):
     """Simple check that we can ping a model"""
     server = http_server.RuntimeHTTPServer()
     with TestClient(server.app) as client:
-        json_input = {"inputs": {"sample_input": {"name": "world"}}}
+        json_input = {"inputs": {"name": "world"}}
```
joerunde, line 222: Do we have a test that checks that `{"inputs": {"sample_input": {"name": "world"}}}` gives a 400?

prashantgupta24: I thought about that - the issue is that the above doesn't give a 400 since it's technically not an incorrect request right? The only thing that happens is that the arg `sample_input` in the `run` function gets a `None` value passed to it

prashantgupta24: Something like `json_input = {"blah": {"sample_input": {"name": "world"}}}` would give a 400, I can add a test for that

joerunde: > the issue is that the above doesn't give a 400 since it's technically not an incorrect request right? The only thing that happens is that the arg sample_input in the run function gets a None value passed to it

hmm....  I'd say it is an invalid request since every request should have all of the required parameters set, and those should be correctly passed along to the model. If we ever try to run a model with a required parameter set to `None`, then we've messed up somewhere

joerunde (approved):
lgtm, I think this does what we want with the API. We could add more up-front request validation but that doesn't need to happen within this one PR

markstur:
What HF API is this aligning with?  I'd think:

`<inference api base url>/pipeline/<TASK_NAME>/model/<MODEL_ID>`

and if we had a default/implied task then just:

`<inference api base url>/models/<MODEL_ID>`

would match the inference API.  Maybe I'm looking at the wrong API.  Based on here: [hfapi](https://github.com/huggingface/hfapi/blob/master/hfapi/__init__.py#L19) and this questionable source: [here](https://discuss.huggingface.co/t/how-to-force-the-task-of-inference-api/15853)

gabe-l-hart:
Good question Mark, we need to define that better. For the time being, we're not worrying about the url endpoint, only the payloads

PR #341, opened 2023-07-18
Lazy local model loading 324

gabe-l-hart:
## Description

This is a better implementation of https://github.com/caikit/caikit/pull/336

This PR adds a new feature to `caikit.runtime` to use a local directory on disk (which can be mounted) to act as a shared-state cache. This will allow replicas of `caikit.runtime` to host an eventually-consistent set of models that changes dynamically. This is a significantly weaker value proposition than `model-mesh`, but for many model types, especially those that use a `module_backend` to delegate the work to another engine (e.g. `caikit-tgis-backend`), it's fine to require that all models be loaded on all replicas.

### DIFF
--- a/caikit/config/config.yml
+++ b/caikit/config/config.yml
@@ -73,6 +73,9 @@ runtime:
     library: sample_lib # TODO: replace with libraries below when runtime can support multiple libraries
     local_models_dir: models
 
+    # Number of threads to make available for load jobs (null => all)
+    load_threads: null
+
     # TLS configs
     tls:
         server:
--- a/caikit/config/config.yml
+++ b/caikit/config/config.yml
@@ -73,7 +73,14 @@ runtime:
     library: sample_lib # TODO: replace with libraries below when runtime can support multiple libraries
     local_models_dir: models
 
+    # If enabled, the models in local_models_dir will be periodically sync'ed
+    # with the in-memory models. New models that are not in-memory that are
+    # found in local_models_dir will be loaded and existing models that were
+    # previously loaded from local_models_dir, but are no longer present will be
+    # unloaded.
     lazy_load_local_models: false
+    lazy_load_poll_period_seconds: 10
+
     # Number of threads to make available for load jobs (null => all)
     load_threads: null
 
--- a/caikit/config/config.yml
+++ b/caikit/config/config.yml
@@ -73,6 +73,7 @@ runtime:
     library: sample_lib # TODO: replace with libraries below when runtime can support multiple libraries
     local_models_dir: models
 
+    lazy_load_local_models: false
     # Number of threads to make available for load jobs (null => all)
     load_threads: null
 
--- a/caikit/runtime/http_server.py
+++ b/caikit/runtime/http_server.py
@@ -472,7 +472,7 @@ class RuntimeHTTPServer(RuntimeServerBase):
         if get_origin(field_type) is Annotated:
             return cls._get_pydantic_type(get_args(field_type)[0])
         if get_origin(field_type) is Union:
-            return Union.__getitem__(
+            return Union.__getitem__(  # pylint: disable=unnecessary-dunder-call
                 tuple(
                     (
                         cls._get_pydantic_type(arg_type)
--- a/caikit/runtime/http_server.py
+++ b/caikit/runtime/http_server.py
@@ -472,14 +472,14 @@ class RuntimeHTTPServer(RuntimeServerBase):
         if get_origin(field_type) is Annotated:
             return cls._get_pydantic_type(get_args(field_type)[0])
         if get_origin(field_type) is Union:
-            return Union.__getitem__(  # pylint: disable=unnecessary-dunder-call
+            return Union[  # type: ignore
                 tuple(
                     (
                         cls._get_pydantic_type(arg_type)
                         for arg_type in get_args(field_type)
                     )
                 )
-            )
+            ]
         if get_origin(field_type) is list:
             return List[cls._get_pydantic_type(get_args(field_type)[0])]
 
--- a/caikit/runtime/model_management/model_manager.py
+++ b/caikit/runtime/model_management/model_manager.py
@@ -66,6 +66,8 @@ class ModelManager:
 
     __model_size_gauge_lock = threading.Lock()
 
+    _LOCAL_MODEL_TYPE = "standalone-model"
+
     def __init__(self):
         """Initialize a ModelManager instance."""
         # Re-instantiating this is a programming error
@@ -90,6 +92,9 @@ class ModelManager:
         ):
             log.info("<RUN44739400I>", "Loading local models into Caikit Runtime...")
             self.initialize_local_models()
+
+        # Keep track of whether lazy loading is enabled
+        self._lazy_load_local_models = get_config().runtime.lazy_load_local_models
 
     def load_model(
         self,
@@ -208,7 +213,9 @@ class ModelManager:
             try:
                 # Use the file name as the model id
                 model_path = os.path.join(self._local_models_dir, model_id)
-                self.load_model(model_id, model_path, "standalone-model", wait=False)
+                self.load_model(
+                    model_id, model_path, self._LOCAL_MODEL_TYPE, wait=False
+                )
             except CaikitRuntimeException as err:
                 log.warning(
                     "<RUN56627484W>",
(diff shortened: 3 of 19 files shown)


caikit/runtime/http_server.py
```diff
@@ -451,7 +451,7 @@ def _get_pydantic_type(cls, field_type: type) -> type:
         if get_origin(field_type) is Annotated:
             return cls._get_pydantic_type(get_args(field_type)[0])
         if get_origin(field_type) is Union:
-            return Union.__getitem__(
+            return Union.__getitem__(  # pylint: disable=unnecessary-dunder-call
```
gabe-l-hart, line 454: This only failed on certain python versions, but I figured it was worth throwing in there

caikit/runtime/model_management/model_manager.py
```diff
@@ -192,16 +334,11 @@ def unload_model(self, model_id) -> int:
 
         return model_size
 
-    def unload_all_models(self) -> None:
```
gabe-l-hart, line 195: This is only ever used in unit tests, so you could argue that it should be removed entirely. As it was, it was missing logic from `unload_model` and was doing too much exception handling

caikit/runtime/model_management/model_manager.py
```diff
@@ -222,22 +359,6 @@ def get_model_size(self, model_id) -> int:
             )
 
         loaded_model = self.loaded_models[model_id]
-
-        # Model sizes should all be cached
```
gabe-l-hart, line 226: This was code that would never be hit unless someone mutated `loaded_models` outside of `load_model` which should never happen except in a nefarious unit test.

caikit/runtime/model_management/model_manager.py
```diff
@@ -252,33 +373,49 @@ def estimate_model_size(self, model_id, local_model_path, model_type) -> int:
...
+        if not model_id or not isinstance(model_id, str):
             raise CaikitRuntimeException(
                 StatusCode.INVALID_ARGUMENT, "Missing required model ID"
             )
 
-        # Now retrieve the model
-        if model_id not in self.loaded_models:
-            # We should not encounter this scenario, so if it happens, log
-            # it as an error-level log
+        # Now retrieve the model and fall back to lazy loading
```
gabe-l-hart, line 390: Here's the real guts of it all! This is where the lazy loading fallback happens.

caikit/runtime/model_management/model_manager.py
```diff
@@ -252,33 +373,49 @@ def estimate_model_size(self, model_id, local_model_path, model_type) -> int:
...
         """
-        if not model_id:
+        if not model_id or not isinstance(model_id, str):
             raise CaikitRuntimeException(
                 StatusCode.INVALID_ARGUMENT, "Missing required model ID"
             )
 
-        # Now retrieve the model
-        if model_id not in self.loaded_models:
-            # We should not encounter this scenario, so if it happens, log
```
gabe-l-hart, line 271: This was true when the only way to access an inference call was behind MM, but it's not at all true when `caikit.runtime` is exposed as a frontend API and could easily have resulted in logs full of error-level messages from DOS-style requests with bad model ids

caikit/runtime/servicers/model_runtime_servicer.py
```diff
@@ -57,12 +57,14 @@ def loadModel(self, request, context):
             )
             caikit_config = get_config()
             if caikit_config.runtime.use_abortable_threads:
+                aborter = CallAborter(context)
```
gabe-l-hart, line 60: This aborter is now used to abort the top-level action as well as the internal thread doing the loading

caikit/runtime/work_management/abortable_action.py
```diff
@@ -25,6 +27,21 @@
 log = alog.use_channel("ABORT-ACTION")
 
 
+class ActionAborter(abc.ABC):
```
gabe-l-hart, line 30: Adding this new abstract class was not strictly necessary. I thought I might need an alternate implementation for aborting the internal load thread and found that I didn't. That said, having this as an abstraction will allow a non-grpc-context implementation for HTTP, so I kept it.

tests/runtime/conftest.py
```diff
@@ -246,13 +247,12 @@ def register_trained_model(
     the old auto-load feature which was only needed for unit tests
     """
     model_future = MODEL_MANAGER.get_model_future(training_id)
-    model = model_future.load()
     loaded_model = (
         LoadedModel.Builder()
         .id(model_id)
         .type("trained")
         .path("")
-        .module(model)
```
gabe-l-hart, line 255: There are a lot of places where the words `module` and `model` are intermixed. In this case, `module` felt distinctly incorrect because `module` should be the `class` while `model` should be the `instance` and the _whole_ point here is to create an `instance`.

tests/runtime/model_management/test_model_manager.py
gabe-l-hart, line 1: The diff here is going to suck because I ditched `unittest` style. Sorry about that!

caikit/runtime/http_server.py
gabe-l-hart: So, it _does_ work if you write it that way, but my inline linter in VS Code goes crazy for it (lots of red squiggles) because `Union[...]` is supposed to have more than one arg and the args are supposed to be types. The fact that `Union` has special semantics for `tuple` appears to not be handled. If it's just me, we can switch it and I can try to find out how to disable this in whatever linter system VS code auto-installed for me.

tests/runtime/model_management/test_model_manager.py
joerunde: Don't be!

caikit/runtime/work_management/call_aborter.py
```diff
@@ -26,11 +26,12 @@
 
 # Local
 from caikit.runtime.types.caikit_runtime_exception import CaikitRuntimeException
+from caikit.runtime.work_management.abortable_action import ActionAborter
 
 log = alog.use_channel("CALL-ABORTER")
 
 
-class CallAborter:
+class CallAborter(ActionAborter):
```
joerunde, line 34: Maybe we should rename this class to `RPCAborter`? Since it specifically uses a grpc context

caikit/runtime/model_management/model_manager.py
```diff
@@ -65,92 +70,223 @@ class ModelManager:
 
     __model_size_gauge_lock = threading.Lock()
 
+    _LOCAL_MODEL_TYPE = "standalone-model"
```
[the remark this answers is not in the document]
joerunde: oof, well it shouldn't have been!

Having both a `LOCAL_MODEL_TYPE` and `_LOCAL_MODEL_TYPE` might be a bit confusing here

caikit/runtime/http_server.py
joerunde: Oh super interesting. Does star-expansion fix your lint issue? Both of these seem to work and neither gets a complaint from Pycharm:
```
some_types = (int, str, bool)
thing = Union[some_types]
thing2 = Union[*some_types]

thing == thing2 # True
thing is thing2 # True
```

caikit/runtime/model_management/model_manager.py
```diff
@@ -65,92 +70,230 @@ class ModelManager:
...
+                    log.warning(
+                        "<RUN56627485W>",
+                        "Failed to load model %s: %s",
+                        model_id,
+                        repr(err),
+                        exc_info=True,
+                    )
+
+        # If running periodically, kick off the next iteration
+        if self._enable_lazy_load_poll:
```
joerunde, line 284: Should we stuff the above implementation details into a private function and wrap this up a bit neater into something like:

```
def sync_local_models():

    try:
        self._sync()
    except Exception as e:
        log.warning("stuff went wrong...")
        
    if self._lazy_sync_timer is None:
        ...
    self._lazy_sync_timer.start()
```

I'd like to both:
- make it a bit more clear that this function is responsible for re-invoking itself via the timer
- make really certain that we're wrapping all of the work in an exception handler so that a failure during one iteration doesn't end up killing the bit that queues up the next call

tests/runtime/model_management/test_model_manager.py
```diff
@@ -34,370 +38,739 @@
...
+            shutil.copytree(good_model_path, model_cache_path)
+            assert manager_one.retrieve_model(model_id)
+            assert model_id in manager_one.loaded_models
+
+            # Wait for the sync period and make sure it's available in the
+            # second manager
+            start = time.time()
+            mgr_two_loaded = False
+            while (time.time() - start) < (purge_period * 1000):
+                time.sleep(purge_period)
```
joerunde, line 458: sleeeeeps in tests :((((

I guess this one isn't too egregious, though if it ever starts failing it'll kill a second waiting to do so. I don't think there's really a better way to do this though unless you were to explicitly invoke the second model-manager's `sync`, which I guess defeats the purpose of checking that it is periodically syncing

caikit/config/config.yml
```diff
@@ -73,6 +73,17 @@ runtime:
...
+    # If enabled, the models in local_models_dir will be periodically sync'ed
+    # with the in-memory models. New models that are not in-memory that are
+    # found in local_models_dir will be loaded and existing models that were
+    # previously loaded from local_models_dir, but are no longer present will be
+    # unloaded.
+    lazy_load_local_models: false
+    lazy_load_poll_period_seconds: 10
+
+    # Number of threads to make available for load jobs (null => all)
+    load_threads: null
```
joerunde, line 85: nit that most integer-type args I've seen in the wild do something like `(-1 for unlimited)`

but I don't really care either way

joerunde (approved):
I think this addresses all the high-level concerns I had about synchronization so I'd be okay to merge and keep iterating on this as we try to integrate it into projects and see what comes up.

Mostly just had some style nits, I kinda would like to break up the big `sync` method into the impl + a safe wrapper that re-invokes itself but not super urgent to get to atm.

caikit/config/config.yml
gabe-l-hart: I did this to mirror how `ThreadPoolExecutor`'s `__init__` args look. If you pass `None`, it's "all"

tests/runtime/model_management/test_model_manager.py
gabe-l-hart: Heh, I was thinking of _exactly_ this comment when I put this loop in rather than just a static sleep. I figured a tight loop that knows when the condition is met was better!

caikit/runtime/model_management/model_manager.py
gabe-l-hart: Very good point. Let me see how quickly I can bang that out

gabe-l-hart: Done

Issue #340, opened 2023-07-17
Put in tests for unary-in stream-out case for REST


prashantgupta24:
## Description

As a Caikit user, I want to put in tests for the unary-in stream-out case so that it's well tested against new code.

## Additional context

- https://sysid.github.io/server-sent-events/ (SSE package)
- https://github.com/sysid/sse-starlette/blob/master/tests/test_event_source_response.py (tests for SSE)

## Acceptance Criteria

- [x] Unit tests cover new/changed code

PR #342, opened 2023-07-18
✅ Add REST server streaming test 

prashantgupta24:
On top of https://github.com/caikit/caikit/pull/338

Taken inspiration from https://github.com/sysid/sse-starlette/blob/master/tests/test_event_source_response.py

fixes https://github.com/caikit/caikit/issues/340

### DIFF
--- a/caikit/runtime/http_server.py
+++ b/caikit/runtime/http_server.py
@@ -222,7 +222,7 @@ class RuntimeHTTPServer(RuntimeServerBase):
                 self._add_unary_input_unary_output_handler(rpc)
 
     def _get_request_params(
-        self, rpc: CaikitRPCBase, request: "pydantic_request"
+        self, rpc: CaikitRPCBase, request: Type[pydantic.BaseModel]
     ) -> Dict[str, Any]:
         """get the request params based on the RPC's req params"""
         request_kwargs = dict(request)
--- a/caikit/runtime/http_server.py
+++ b/caikit/runtime/http_server.py
@@ -18,7 +18,7 @@ API based on the task definitions available at boot.
 """
 # Standard
 from functools import partial
-from typing import Iterable, List, Optional, Type, Union, get_args
+from typing import Any, Dict, Iterable, List, Optional, Type, Union, get_args
 import asyncio
 import enum
 import json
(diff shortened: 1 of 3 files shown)


caikit/runtime/http_server.py
```diff
@@ -221,6 +221,30 @@ def _bind_routes(self, service: ServicePackage):
             else:
                 self._add_unary_input_unary_output_handler(rpc)
 
+    def _get_request_params(
+        self, rpc: CaikitRPCBase, request: "pydantic_request"
```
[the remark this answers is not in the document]
joerunde: Yeah I'm not sure that will work... 🤔 

I think this should be `Type[pydantic.BaseModel]` since it's a type that extends a pydantic model

tests/runtime/test_http_server.py
```diff
@@ -260,6 +257,25 @@ def test_inference_other_task(other_task_model_id):
...
+            f"/api/v1/{sample_task_model_id}/task/server-streaming-sample",
+            json=json_input,
+        )
+        assert stream.status_code == 200
+        assert (
+            stream.content.decode(stream.default_encoding).count("SampleOutputType")
+            == 10
+        )
+        assert (
+            b"SampleOutputType(greeting='Hello world stream')\r\n\r\n" in stream.content
```
joerunde, line 275: so many carriage returns!

tests/runtime/test_http_server.py
```diff
@@ -260,6 +257,25 @@ def test_inference_other_task(other_task_model_id):
         assert json_response["farewell"] == "goodbye: world 42 times"
 
 
+def test_inference_streaming_sample_module(sample_task_model_id):
+    """Simple check that we can ping a model"""
```
joerunde, line 261: nit: could maybe use a better docstring re: testing a happy path unary-stream case

tests/runtime/test_http_server.py
prashantgupta24: yeah, even the [example](https://github.com/sysid/sse-starlette/blob/master/tests/test_event_source_response.py) from `test_event_source_response` has them

caikit/runtime/http_server.py
```diff
@@ -221,6 +221,30 @@ def _bind_routes(self, service: ServicePackage):
...
+    ) -> Dict[str, Any]:
+        """get the request params based on the RPC's req params"""
+        request_kwargs = dict(request)
+        required_params = rpc.task.get_required_parameters(False)
```
joerunde, line 229: nit: the rpc itself should hold the info about whether the input is streaming or not, and maybe that should be used here instead of hardcoding false?

Might at least be worth a comment saying
```
# Here we assume the input isn't streaming
```

joerunde (approved):
nice lil' refactor too!

lpgtm, thanks for the test 👍

caikit/runtime/http_server.py
prashantgupta24: Good point!

Issue #344, opened 2023-07-20
Use correct inference method name in rest server


joerunde:
We hardcoded the inference method names in the rest server- instead we need to pull them off the module class for the correct flavor of streaming

PR #345, opened 2023-07-20
:bug: Allow user-defined methods for http server

evaline-ju:
Closes: #344 

The assumption for the http server had been that `run` and `run_stream_out` would be the method names for unary-unary and unary-stream methods. This PR updates the server to check the method on present, which like the grpc server requires retrieving the model. Retrieving is placed in the try/catch since 404s/not founds are also returned as `CaikitRuntimeExceptions`. Further cleanup could be done in the future to make the http server more align with the grpc server.

### DIFF
--- a/caikit/runtime/http_server.py
+++ b/caikit/runtime/http_server.py
@@ -266,11 +266,18 @@ class RuntimeHTTPServer(RuntimeServerBase):
 
             log.debug4("Sending request %s to model id %s", request_params, model_id)
             try:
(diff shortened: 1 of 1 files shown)


joerunde (approved):
👍 🚀 🌶️ 

We need some overhaul of all the stuff that comes before `predict_model` but I assume we'll get there as we shore up the http side of things

prashantgupta24:
Nice!

Issue #347, opened 2023-07-20
Model Management Factory Plugins


gabe-l-hart:
## Is your feature request related to a problem? Please describe.

The new `caikit.core.model_management` abstractions require that new implementations of the given factory's abstraction get registered before they can be instantiated. This works well when the implementations live in `caikit` itself, but it requires extra code to bind an external implementation into the factory before it can be used. For using `caikit.runtime`, this would require that a custom entrypoint be written.

## Describe the solution you'd like

We need a way to tell the factory to dynamically register new implementations based on config! The simplest way would be to provide a new argument in the configuration blob that points to a class in a library that will be dynamically imported before construction.

## Describe alternatives you've considered

This could be handled with a new entrypoint that takes these arguments as configurable, but that would not extend to pure library usage.

PR #348, opened 2023-07-20
ModelManagementFactoryImports

gabe-l-hart:
## Description

Make model management factories support import plugins. In each factory blob, there's now an additional "import_class" field that can be added to force an import of an external library with a predefined class in it that should be registered and constructed.

Closes: https://github.com/caikit/caikit/issues/347


joerunde (approved):
sweeeeet!

Love to see the parameterized tests

Issue #346, opened 2023-07-20
REST server cannot handle null in model response


joerunde:
## Describe the bug

If I have a module which does not set a field in its response data model, and that field is itself a type which accepts None as a default (e.g. another data model type), then response serialization fails in the rest server.

## Platform

- Interpreter version: 3.11
- Library version: 0.11.0

## Sample Code

The simplest thing to do is not set a `ProducerId` where a DM output has one specified

## Expected behavior

Either a `{producer_id: null}` or a `{producer_id: {name: "", version: ""}}` would probably be acceptable.

## Observed behavior

We get a cryptic error from uvicorn with not much help:

> {"channel": "uvicorn.error", "exception": "Traceback (most recent call last):
\n File "/usr/local/lib/python3.11/site-packages/uvicorn/protocols/http/httptools_impl.py", line 426, in run_asgi
\n result = await app( # type: ignore[func-returns-value]
\n ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
\n File "/usr/local/lib/python3.11/site-packages/uvicorn/middleware/proxy_headers.py", line 84, in call
\n return await self.app(scope, receive, send)
\n ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
\n File "/usr/local/lib/python3.11/site-packages/fastapi/applications.py", line 289, in call
\n await super().call(scope, receive, send)
\n File "/usr/local/lib/python3.11/site-packages/starlette/applications.py", line 122, in call
\n await self.middleware_stack(scope, receive, send)
\n File "/usr/local/lib/python3.11/site-packages/starlette/middleware/errors.py", line 184, in call
\n raise exc
\n File "/usr/local/lib/python3.11/site-packages/starlette/middleware/errors.py", line 162, in call
\n await self.app(scope, receive, _send)
\n File "/usr/local/lib/python3.11/site-packages/starlette/middleware/exceptions.py", line 79, in call
\n raise exc
\n File "/usr/local/lib/python3.11/site-packages/starlette/middleware/exceptions.py", line 68, in call
\n await self.app(scope, receive, sender)
\n File "/usr/local/lib/python3.11/site-packages/fastapi/middleware/asyncexitstack.py", line 20, in call
\n raise e
\n File "/usr/local/lib/python3.11/site-packages/fastapi/middleware/asyncexitstack.py", line 17, in call
\n await self.app(scope, receive, send)
[...]

PR #349, opened 2023-07-20
🥅 add better REST error handling

prashantgupta24:
fix https://github.com/caikit/caikit/issues/346

## Changes

- This PR implements `ResponseValidationError` which validates the handler's `response` and enables better error message from the response validation.

    - To test this, this PR also adds `ProducerId` as a required field in `OtherOutputType` and adds a poison pill within `other_implementation` so that we can test error conditions if a required field is not present. 

- This PR also adds other `http` tests to test for incorrect/missing fields in `request` messages.
## Useful links

- Uses [requestvalidationerror](https://fastapi.tiangolo.com/tutorial/handling-errors/#use-the-requestvalidationerror-body) documentation but for `ResponseValidationError`.

### DIFF
--- a/caikit/runtime/http_server.py
+++ b/caikit/runtime/http_server.py
@@ -28,7 +28,7 @@ import threading
 import time
 
 # Third Party
-from fastapi import FastAPI, Request, Response
+from fastapi import FastAPI, Request, Response, status
 from fastapi.encoders import jsonable_encoder
 from fastapi.exceptions import ResponseValidationError
 from fastapi.responses import JSONResponse, PlainTextResponse
@@ -119,10 +119,11 @@ class RuntimeHTTPServer(RuntimeServerBase):
 
         self.app = FastAPI()
 
+        # Response validation
         @self.app.exception_handler(ResponseValidationError)
         async def validation_exception_handler(_, exc: ResponseValidationError):
             return JSONResponse(
-                status_code=500,
+                status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
                 content=jsonable_encoder({"detail": exc.errors(), "body": exc.body}),
             )
 
--- a/caikit/runtime/http_server.py
+++ b/caikit/runtime/http_server.py
@@ -29,7 +29,9 @@ import time
 
 # Third Party
 from fastapi import FastAPI, Request, Response
-from fastapi.responses import PlainTextResponse
+from fastapi.encoders import jsonable_encoder
+from fastapi.exceptions import ResponseValidationError
+from fastapi.responses import JSONResponse, PlainTextResponse
 from grpc import StatusCode
 from sse_starlette import EventSourceResponse, ServerSentEvent
 import numpy as np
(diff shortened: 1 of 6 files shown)


tests/fixtures/sample_lib/data_model/sample.py
```diff
@@ -29,6 +30,7 @@ class OtherOutputType(DataObjectBase):
     """A simple return type for the `other_task` task"""
 
     farewell: str
+    producer_id: ProducerId
```
prashantgupta24, line 33: Ideally I want to be able to test this without changing the actual task definition. Mocks maybe?

caikit/runtime/http_server.py
```diff
@@ -111,6 +113,13 @@ def __init__(self, tls_config_override: Optional[aconfig.Config] = None):
 
         self.app = FastAPI()
 
+        @self.app.exception_handler(ResponseValidationError)
```
gabe-l-hart, line 116: I can't find much documentation on `ResponseValidationError`, but is this validating the handler's response rather than the request? If so, I agree that 500 is the right return value

tests/runtime/test_http_server.py
```diff
@@ -253,9 +253,9 @@ def test_inference_other_task(other_task_model_id):
             f"/api/v1/{other_task_model_id}/task/other",
             json=json_input,
         )
-        assert response.status_code == 200
+        assert response.status_code == 500
```
gabe-l-hart, line 256: Hm, I don't understand this test. It looks like we took a working test and since the output doesn't have `producer_id` in it, that results in a 500 instead of a 200? If so, I agree with the above comment about needing a better way to test this. Maybe some kind of poison pill in the request that causes a bad response object?

gabe-l-hart (changes requested):
Given the description of #346, I'm not sure this actually solves the problem of filling in "nullable" values. It looks like this would cause the example in the issue to return a `500` rather than a `200` with an empty value filled in for `producer_id`

prashantgupta24:
> Given the description of #346, I'm not sure this actually solves the problem of filling in "nullable" values. It looks like this would cause the example in the issue to return a `500` rather than a `200` with an empty value filled in for `producer_id`

In #346, I explained that a `500` makes sense in terms of `pydantic` because we're not providing all mandatory return type fields. So either we make fields `optional`, in which case, we get a `200` back, _or_, we provide a better error message (but the error still remains `500`). 

This PR is trying to address giving a better error message, and not filling in "nullable" values.

tests/runtime/test_http_server.py
prashantgupta24: Good idea! Added a poison pill

caikit/runtime/http_server.py
prashantgupta24: [This](https://fastapi.tiangolo.com/tutorial/handling-errors/#use-the-requestvalidationerror-body) is the only documentation I found which is for `RequestValidationError`, I extrapolated for `ResponseValidationError`

gabe-l-hart (approved):

Issue #249, opened 2023-06-19
Support native python list initialization within a Union


prashantgupta24:
## Describe the bug

Right now, if a user creates a `Dataobject` like this:
```
@dataobject
class Foo(DataObjectBase):
    foo: Union[List[str], List[int]]
```
They have to create an object using
```
foo_int = Foo.IntSequence(values=[1, 2])
or
foo_str = Foo.StrSequence(values=["hello", "world"])
```
instead of the pythonic way of:

```
foo_with_ints = Foo([1, 2, 3])
or
foo_with_strs = Foo(["1", "2", "3"])
```

## Platform

Please provide details about the environment you are using, including the following:

- Interpreter version: 3.9
- Library version: 0.8.0

aluu317:
Update to the current way to create Union of lists (with the merge of PR #294),

users would have to create a `union_list_str_int_dm` first, like so:
```
union_list_str_int_dm = caikit.interfaces.common.data_model.UnionListStrIntSource

request = Foo(
        foo=union_list_str_int_dm(
            union_list=union_list_str_int_dm.StrSequence(values=["str", "sequence"]))

# or

request = Foo(
        foo=union_list_str_int_dm(
            union_list=union_list_str_int_dm.IntSequence(values=[1, 2]))
```
But what we'd like is:
```
request = Foo(
        foo=["str", "sequence"]
        )
        
# or
request = Foo(
        foo=[1, 2]
        )
```

PR #350, opened 2023-07-20
Support native python lists in Union of Lists

aluu317:
issue #249 

This PR changes `base.py` and corresponding tests to accommodate native Python list instantiation for Union of lists:
- converting between proto and data object: `to_proto()` and `from_proto()`
- converting between json and data object: `to_json()` and `from_json()`

Example:
```
class Foo(DataObjectBase):
        foo: Union[List[str], List[int]]
        bar: Union[List[str], List[int]]

foo_int = Foo(foo=[1,2,3], bar=["hello","world"])
```

This will go to proto as: `foo_int_sequence` field has value [1, 2, 3] and `bar_str_sequence` has value ["hello", "world"]
This will go to json as: `{"foo": "[1,2]", "bar": "['hello', 'world'"}`
And to_dict as:  `{"foo_int_sequence": {"values": [1,2]}, "bar_str_sequence": {"values": ['hello', 'world']}}`

Note: I went back and forth between what `to_dict` should return. I did it the way above so we have a true way of returning the representation of `foo_int_sequence` when `to_dict` is called. And` to_json` is a "user-friendly" way of reading the python native lists. I'm happy to change `to_dict` if we dont see value in that direction.


tests/core/data_model/test_dataobject.py
prashantgupta24, line 1: Haven't seen the changes yet, but what would it take to have `foo_int` declared like this?
```
 foo_int = Foo([1, 2])
```

aluu317: I think it'll work. Let me change one test to not explicitly say `foo=[1,2,]` if that's what you mean?

joerunde (approved):
lookin slick!

tests/core/data_model/test_dataobject.py
```diff
@@ -517,6 +517,32 @@ class Foo(DataObjectBase):
     assert foo2.foo_bool2
 
 
+@pytest.mark.skip(reason="wip")
```
prashantgupta24, line 520: worth adding some more context on what the WIP is? Even I can't remember now 🤔

aluu317: yes! I updated the reason to be that we still cannot use static BoolSequence in core (issue https://github.com/caikit/caikit/issues/409). But correct me if that's not the reason. The error seen if this test is run is:
```
...
except TypeError as e:
                # More likely than not, this is a duplicate symbol; the main case in which
                # this could occur is when you've compiled files with protoc, added them to your
                # descriptor pool, and ALSO added the defs in your py_to_proto schema, but the
                # lookup validation with fd_proto.name is skipped because the .proto file fed to
                # protoc had a different name!
>               raise TypeError(
                    f"Failed to add {fd_proto.name} to descriptor pool with error: [{e}]; Hint: if you previously used protoc to compile this definition, you must recompile it with the name {fd_proto.name} to avoid the conflict."
                )
                
                
E               TypeError: Failed to add foo.proto to descriptor pool with error: [Couldn't build proto file into descriptor pool!
E               Invalid proto descriptor for file "foo.proto":
E                 caikit_data_model.Foo.FooBoolSequence.values: "values" is already defined in "caikit_data_model.Foo.FooBoolSequence".
E                 caikit_data_model.Foo.FooBoolSequence: "FooBoolSequence" is already defined in "caikit_data_model.Foo".
E               ]; Hint: if you previously used protoc to compile this definition, you must recompile it with the name foo.proto to avoid the conflict.

venv/lib/python3.9/site-packages/py_to_proto/utils.py:67: TypeError
```

prashantgupta24: Yeah that sounds good (technically it's more of we `don't` use static BoolSequence in core instead of `can't`, at the moment)

prashantgupta24:
Why did this start failing at `isinstance(union_list, List)` in `primitive_party_implementation.py`?

prashantgupta24:
ohh no, it started failing after merging with the REST for train stuff 😢  I'll look into it!

caikit/core/data_model/base.py
```diff
@@ -830,6 +853,14 @@ def fill_proto(self, proto):
                 elif subproto.DESCRIPTOR.full_name == timestamp.TIMESTAMP_PROTOBUF_NAME:
                     timestamp_proto = timestamp.datetime_to_proto(attr)
                     subproto.CopyFrom(timestamp_proto)
+                # check that this is any of the Union of List types
+                elif subproto.DESCRIPTOR.full_name.endswith("Sequence"):
```
prashantgupta24, line 857: This maintains backward compatibility with anyone trying to instantiate a DM object with `StrSequence` . 

So for `SampleTaskSamplePrimitiveModuleTrainRequest`, 

```
union_list=caikit.interfaces.common.data_model.StrSequence(values=["str", "sequence"]),
``` 
and 
```
union_list=["str", "sequence"],
```
both work now.
```suggestion
                elif subproto.DESCRIPTOR.full_name.endswith(
                    "Sequence"
                ) and not issubclass(attr.__class__, DataBase):
```

prashantgupta24: The reason we have to do this is because the way unions of lists is handled is different if they are created in a `DataObject` manually vs if they are present as part of a function signature which gets parsed. #409 will fix this once and for all by making this handling common across both scenarios.

prashantgupta24:
For conflicts with the REST server work:

The issue is that we create a `pydantic` model from our DM class for `SampleTaskSamplePrimitiveModuleTrainRequest`, and that pydantic model has `*Sequence` within it. Once we convert it back to our DM object, the `*sequence` persists.

in test_grpc_server, the `SampleTaskSamplePrimitiveModuleTrainRequest` looks like this:
```
SampleTaskSamplePrimitiveModuleTrainRequest(model_name='primitive_trained_model', output_path=None, sample_input=SampleInputType(name='Gabe'), simple_list=['hello', 'world'], union_list=['str', 'sequence'], union_list2=[1, 2], union_list3=[True, False], union_list4=123, training_params_json_dict_list=[{'foo': {'bar': [1, 2, 3]}}], training_params_json_dict={'foo': {'bar': [1, 2, 3]}}, training_params_dict={'layer_sizes': 100, 'window_scaling': 200}, training_params_dict_int={1: 0.1, 2: 0.01})
```

vs in `http_server`, it looks like:

```
SampleTaskSamplePrimitiveModuleTrainRequest(model_name='primitive_task_train', output_path=None, sample_input=SampleInputType(name='test'), simple_list=['hello', 'world'], union_list=StrSequence(values=['hello', 'world']), union_list2=StrSequence(values=['hello', 'world']), union_list3=StrSequence(values=['hello', 'world']), union_list4=1, training_params_json_dict_list=[{'foo': {'bar': [1, 2, 3]}}], training_params_json_dict={'foo': {'bar': [1, 2, 3]}}, training_params_dict={'layer_sizes': 100, 'window_scaling': 200}, training_params_dict_int={1: 0.1, 2: 0.01})
```

prashantgupta24 (approved):

PR #351, opened 2023-07-21
:sparkles::white_check_mark: Use text for text generation task

evaline-ju:
For trying to move caikit-nlp text generation modules to use the prescribed interfaces and tasks for text generation per `caikit.interfaces.nlp` in https://github.com/caikit/caikit-nlp/issues/92, tests can be moved here and `inputs` was presumably used to conform with Huggingface interfaces. However for semantically meaningful use, the modules have all used `text` as the input. With the REST server, `inputs` will be usable to align with Huggingface interfaces.

### DIFF
--- a/caikit/interfaces/nlp/tasks.py
+++ b/caikit/interfaces/nlp/tasks.py
@@ -24,7 +24,7 @@ from .data_model.text_generation import GeneratedTextResult, GeneratedTextStream
 
 
 @task(
-    unary_parameters={"inputs": str},
+    unary_parameters={"text": str},
     unary_output_type=GeneratedTextResult,
     streaming_output_type=Iterable[GeneratedTextStreamResult],
 )
--- a/tests/interfaces/nlp/test_nlp_tasks.py
+++ b/tests/interfaces/nlp/test_nlp_tasks.py
@@ -13,13 +13,13 @@
 # limitations under the License.
 """Tests for the NLP task definitions"""
 # Standard
-from typing import Dict, Type
+from typing import Dict
 
 # Third Party
 import pytest
 
 # Local
-from caikit.core import ModuleBase, TaskBase, module
+from caikit.core import ModuleBase, module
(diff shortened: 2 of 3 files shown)


alex-jw-brooks (approved):
Looks good to me, thanks!

Can you elaborate a bit on how this: `With the REST server, inputs will be usable to align with Huggingface interfaces.` will be accomplished? I admittedly have not followed the REST work at all, but is that not going to mirror the gRPC interface?

joerunde:
@alex-jw-brooks
> but is that not going to mirror the gRPC interface?

Correct, the REST service is responsible for presenting a huggingface-compatible API. One of the changes it makes is that it replaces the name of the task's required input parameter with `inputs`. We still don't know what we'll do if a task has multiple required input parameters but we're crossing that bridge when we get there!
Future changes will allow that `inputs` field to be a list for batch inference, and allow raw bytes to be posted to the inference endpoints as well, to match what huggingface accepts.

caikit/interfaces/nlp/tasks.py
```diff
@@ -24,7 +24,7 @@
 
 
 @task(
-    unary_parameters={"inputs": str},
+    unary_parameters={"text": str},
```
gabe-l-hart, line 27: Why are we making this change? Is this just the fact that it's more meaningful? This is kind of a larger question of how we're handling input arguments in the HTTP server versus gRPC. We went the route of bundling everything under `inputs` in the HTTP server to avoid needing to have all modules change, but if we're writing net-new modules, we could (should?) use `inputs` explicitly so that the gRPC objects line up too

evaline-ju:
Moving to draft pending discussion on interface alignment

evaline-ju:
Update: based on discussion we will be moving forward with `text` - Huggingface alignment i.e. with `inputs` is mainly important for the REST layer and any future Python SDK

PR #356, opened 2023-07-24
🚩 add save_with_id config

prashantgupta24:
Make `save_with_id` configurable through the global config.

fix https://github.com/caikit/caikit/issues/362

### DIFF
--- a/tests/fixtures/config/config.yml
+++ b/tests/fixtures/config/config.yml
@@ -10,7 +10,6 @@ runtime:
 
     training:
         output_dir: test/training_output
-        save_with_id: true
 
 inference_plugin:
     model_mesh:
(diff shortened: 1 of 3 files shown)


tests/fixtures/config/config.yml
```diff
@@ -10,7 +10,7 @@ runtime:
 
     training:
         output_dir: test/training_output
-        auto_load_trained_model: true
+        save_with_id: true
```
gabe-l-hart, line 13: NIT: We shouldn't need this in the test fixture since it's the same value as the default

gabe-l-hart (approved):
This is a nice change to make this more flexible. I'm curious the motivation behind adding it and I have one little NIT, but I think it's good as is

Issue #354, opened 2023-07-24
Add error reason api to model futures


joerunde:
## Is your feature request related to a problem? Please describe.

Model futures currently only return a status enum. This isn't sufficient to provide back a human-interpretable string of the reason behind the error, which our training status api currently leaves unfilled.

## Describe the solution you'd like

Either extend model_future.status to return a string reason as well, or add another abstract interface `.reason() -> str` that can return a string about why the future is in the terminal state that it's in.

## Describe alternatives you've considered

???

PR #357, opened 2023-07-24
:sparkles: Add training info DM

gkumbhat:
Supports: https://github.com/caikit/caikit/issues/354

### Changes
- Add TrainingInfo DM object
- Replace `get_status` with `get_info`
- Add error information for local trainer


caikit/core/data_model/training_info.py
```diff
@@ -0,0 +1,35 @@
...
+from typing import List
+
+# Local
+from ..toolkit.wip_decorator import Action, WipCategory, work_in_progress
+from .dataobject import DataObjectBase, dataobject
+from .package import PACKAGE_COMMON
+from .training_status import TrainingStatus
+
+
+@work_in_progress(action=Action.WARNING, category=WipCategory.BETA)
```
gabe-l-hart, line 28: I'm pretty tempted to yank all of these decorators off since technically the whole project is still in `beta` and it's pretty noisy/confusing to have them just on train related stuff

caikit/core/data_model/training_info.py
```diff
@@ -0,0 +1,35 @@
...
+@dataobject(PACKAGE_COMMON)
+class TrainingInfo(DataObjectBase):
+    errors: List[str]
```
gabe-l-hart, line 31: We should be explicit about `Annotated[..., FieldNumber(x)]` for these so that we can preserve proto compatibility if we shuffle things around

caikit/core/data_model/training_info.py
gabe-l-hart, line 1: I'd probably vote that this whole file get merged into `training_status.py`. We haven't typically required a single module per data object

caikit/core/model_management/local_model_trainer.py
```diff
@@ -132,29 +132,30 @@ def completion_time(self) -> Optional[datetime]:
...
 
             # If the worker is currently alive it's doing work
             if self._worker.is_alive():
-                return TrainingStatus.RUNNING
+                return TrainingInfo(status=TrainingStatus.RUNNING)
 
             # The worker threw outside of a cancellation process
             if self._worker.threw:
-                return TrainingStatus.ERRORED
+                return TrainingInfo(status=TrainingStatus.ERRORED)
```
gabe-l-hart, line 151: I think we should be able to get the error itself here?

gabe-l-hart (changes requested):
Overall looks good, just a couple of NITs

caikit/core/data_model/training_info.py
gkumbhat: True. I'll get that removed

caikit/core/model_management/local_model_trainer.py
gkumbhat: Yep. Will get that added and corresponding tests

caikit/core/data_model/training_info.py
joerunde: I think I added these explicitly for the initial drop of this stuff where we only had local training and none of it supported any sort of distribution. Now that we have real distributed trainers, I think it's reasonably "no longer work in progress"

caikit/core/model_management/local_model_trainer.py
```diff
@@ -173,7 +176,6 @@ def wait(self):
 
             # If running a subprocess, handle abnormal exit codes
             if self._use_subprocess:
-                self._completion_time = self._worker.get_or_throw()
```
gkumbhat, line 176: I think calling `get_or_throw` here was a bug.. @gabe-l-hart sound correct?

gabe-l-hart: Nope! This is how the completion time gets communicated from the subprocess back to the parent. This is needed for the retention policy stuff, otherwise subprocess-based trainings will live forever

gabe-l-hart: It's quite possible that this does not play nicely with errors, though

gkumbhat: hmm. the destroyable process seems to be returning error string or throw error for the object `_result` returned by `get_or_throw()` 🤔

caikit/core/data_model/training_status.py
```diff
@@ -40,3 +42,12 @@ def is_terminal(self):
             self.__class__.CANCELED,
             self.__class__.ERRORED,
         ]
+
+
+@dataobject(PACKAGE_COMMON)
+class TrainingInfo(DataObjectBase):
+    errors: Annotated[List[str], FieldNumber(1)]
+    # TODO: Add elements to conveying other useful information
```
gabe-l-hart, line 50: This is interesting. I think the best we can do there is add `info: JsonDict` to be an arbitrary blob

caikit/core/data_model/training_status.py
```diff
@@ -40,3 +42,12 @@ def is_terminal(self):
...
+    # regarding training status, such as iterations progressed
+    # evaluation so far, etc.
+    status: TrainingStatus
```
gabe-l-hart, line 53: Need a `FieldNumber` here too

caikit/core/model_management/local_model_trainer.py
```diff
@@ -173,14 +173,7 @@ def wait(self):
...
-                if self._worker.exitcode and self._worker.exitcode != os.EX_OK:
-                    if self._worker.exitcode == OOM_EXIT_CODE:
-                        raise MemoryError("Training process died with OOM error!")
-                    if not self._worker.canceled:
-                        raise RuntimeError(
-                            f"Training process died with exit code {self._worker.exitcode}"
-                        )
+                self._completion_time = datetime.now()
```
gabe-l-hart, line 176: What happened to getting this from the child? Also, even if we go with this, we need to make sure it's only set the first time. This would also allow you to remove the `if self._use_subprocess` check.

```py
self._completion_time = self._completion_time or datetime.now()
```

gabe-l-hart: Also, if we do go with this, you can remove the janky bits about returning the completion time in `_train_and_save`

caikit/core/toolkit/destroyable_process.py
```diff
@@ -175,9 +178,18 @@ def _update_result(self):
     @property
     def error(self) -> Optional[Exception]:
         self._update_result()
+
         if isinstance(self.__result, Exception):
             return self.__result
 
+        if self.exitcode and self.exitcode != os.EX_OK:
```
gabe-l-hart, line 185: I wonder if it makes more sense to put this in `self._update_result`? That way the value is stored and not recreated.

caikit/core/data_model/training_status.py
```diff
@@ -40,3 +42,12 @@ def is_terminal(self):
             self.__class__.CANCELED,
             self.__class__.ERRORED,
         ]
+
+
+@dataobject(PACKAGE_COMMON)
+class TrainingInfo(DataObjectBase):
+    errors: Annotated[List[str], FieldNumber(1)]
```
gabe-l-hart, line 49: Why is this a list? Under what circumstances do we expect there to be multiple elements here?

caikit/runtime/servicers/global_train_servicer.py
```diff
@@ -250,6 +250,9 @@ def rpc_termination_callback():
                 log.debug, "Training %s complete in: ", model_future.id
             ):
                 model_future.wait()
+                training_info = model_future.get_info()
+                if training_info.errors:
+                    raise training_info.errors[0]
```
gabe-l-hart, line 255: I commented above, but I'm not really sure why this is a list of errors rather than a single one. It seems like having a list makes this confusing for users that need to basically do exactly what you're doing here (figure out _which_ error to care about).

gabe-l-hart (changes requested):
A couple more things to tighten up with the completion time stuff. I also think we should consider making `TrainingInfo.error` a single instance rather than a list

caikit/core/data_model/training_status.py
gkumbhat: I think couple of examples I was thinking about was, in case we add a validation step where we collect all the data validation errors and return it to user.

caikit/core/model_management/local_model_trainer.py
gkumbhat: > What happened to getting this from the child?

I thought doing it this way would be mostly equivalent.

caikit/core/toolkit/destroyable_process.py
gkumbhat: @gabe-l-hart you mean storing MemoryError etc in `self.__result` ? hmm.  We can do that, but these would be static values only and the exit code etc is with the parent process, so in terms of processing, it would be equivalent.  Adding it in result, would result in 2nd time getting the error from `__result`, which might make a bit confusing, although practically, it would be same value, so may be this confusion would be unnoticeable

caikit/core/model_management/local_model_trainer.py
gkumbhat: Another problem with get_or_throw is, that function will throw error, in which cases, we won't get the completion time set.. which could be a problem?

caikit/core/data_model/training_status.py
gabe-l-hart: Cool, that makes sense!

gabe-l-hart (approved):
Ship it! Very nice

PR #359, opened 2023-07-24
:sparkles: Update text gen interfaces with TGIS compatibility

evaline-ju:
Originally the text generation interfaces in caikit were developed here in alignment with [Huggingface TGI](https://github.com/huggingface/text-generation-inference/blob/main/docs/openapi.json) but the [TGIS streaming interfaces](https://github.com/IBM/text-generation-inference/blob/main/proto/generation.proto) predate and slightly differ from TGI, with the main focus on text.

[`caikit-nlp`](https://github.com/caikit/caikit-nlp) using [caikit-tgis-backend](https://github.com/caikit/caikit-tgis-backend) has text generation module implementations that require TGIS compatibility. The stream result here has been edited for TGIS compatibility but still keeping most TGI fields, where the original TGI `token` could be one of the TGIS-compatible `tokens`.

NOTE: The current interfaces already have more "FinishReason" enums than TGI, more in line with TGIS but missing two. This PR adds the remaining two TGIS enums so there is no gap.

### DIFF
--- a/caikit/interfaces/nlp/data_model/text_generation.py
+++ b/caikit/interfaces/nlp/data_model/text_generation.py
@@ -15,7 +15,7 @@
 
 # Standard
 from enum import Enum
-from typing import Optional
+from typing import List, Optional
 
 # Third Party
 import numpy as np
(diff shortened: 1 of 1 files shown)


gabe-l-hart (approved):
Given the disconnect between TGIS and TGI and the fact that `caikit` is currently using TGIS explicitly, I think this is the right change. In the worst case, a user needs to change the key and unwrap a list of size `1` to have TGI compatibility.

Issue #332, opened 2023-07-13
Implement CancelTraining in TrainingManagementServicer


gabe-l-hart:
## Is your feature request related to a problem? Please describe.

Once a training has been launched, the user should be able to cancel it with the unique ID.

## Describe the solution you'd like

Add a new `CancelTraining` endpoint to the `TrainingManagementServicer` [here](https://github.com/caikit/caikit/blob/main/caikit/runtime/servicers/training_management_servicer.py#L38) that uses the `ModelFuture`'s `cancel` feature.

## Additional context

* Follow on to https://github.com/caikit/caikit/pull/312

PR #360, opened 2023-07-24
Add CancelTraining to training_mangement_servicer

aluu317:
Issue #332 

This PR adds:
- another rpc to `service_factory.py`, TRAINING_MANAGEMENT_SERVICE_SPEC to include a new rpc for cancelling a training. The rpc accepts a training id as input, and returns the TrainingStatus of that training as output after canceling the model future.
- implements the `CancelTraining` method in `training_management_servicer.py`.
- Tests

### DIFF
--- a/tests/runtime/servicers/test_training_management_servicer.py
+++ b/tests/runtime/servicers/test_training_management_servicer.py
@@ -187,7 +187,7 @@ def test_training_raises_when_cancel_on_incorrect_id(training_management_service
     event.set()
     model_future.wait()
 
-    # cancel the training request, check that its status is now canceled
+    # cancel a training with wrong training id
     cancel_request = TrainingInfoRequest(training_id="some_random_id").to_proto()
     with pytest.raises(CaikitRuntimeException) as context:
         training_management_servicer.CancelTraining(cancel_request, context=None)
--- a/tests/runtime/servicers/test_training_management_servicer.py
+++ b/tests/runtime/servicers/test_training_management_servicer.py
@@ -92,15 +92,14 @@ def test_training_cannot_cancel_on_completed_training(training_management_servic
     event.set()
     model_future.wait()
 
-    # cancel the training request, check that its status is still completed because it was completed
+    # cancel the training request, but this won't change its status as it was finished
     training_management_servicer.CancelTraining(request, context=None)
-
     response = training_management_servicer.GetTrainingStatus(request, context=None)
     assert response.status == TrainingStatus.COMPLETED.value
 
 
 def test_training_cancel_on_correct_id(training_management_servicer):
-    # Create a future and set it in the training manager
+    # Create a training future for first model with a wait event
     event_1 = threading.Event()
     model_future_1 = MODEL_MANAGER.train(
         SampleModule,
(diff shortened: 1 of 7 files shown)


tests/runtime/servicers/test_training_management_servicer.py
```diff
@@ -75,6 +75,72 @@ def test_training_runs(training_management_servicer, training_pool):
...
+    # send a request, check it's not errored
+    request = TrainingInfoRequest(training_id=model_future.id).to_proto()
+    response = training_management_servicer.GetTrainingStatus(request, context=None)
+    assert response.status != TrainingStatus.ERRORED.value
+
+    event.set()
+    model_future.wait()
+
+    # cancel the training request, check that its status is now canceled
+    training_management_servicer.CancelTraining(request, context=None)
```
gabe-l-hart, line 96: Interesting. I would expect that calling cancel on the service _after_ calling `model_future.wait()` would show that the model has completed. This may be a bug in the status reporting for the local trainer.

tests/runtime/servicers/test_training_management_servicer.py
```diff
@@ -75,6 +75,72 @@ def test_training_runs(training_management_servicer, training_pool):
...
+
+    event_2.set()
+    model_future_2.wait()
+
+    # cancel the first training request, check that its status is now canceled
+    # and the status of second training request is completed
+    training_management_servicer.CancelTraining(request_1, context=None)
+
+    response_1 = training_management_servicer.GetTrainingStatus(request_1, context=None)
+    assert response_1.status == TrainingStatus.CANCELED.value
```
gabe-l-hart, line 138: Same comment here

tests/runtime/test_grpc_server.py
```diff
@@ -602,6 +602,110 @@ def test_train_fake_module_ok_response_with_datastream_csv_file(
...
+    training_info_request = TrainingInfoRequest(training_id=training_id)
+    training_management_response: TrainingInfoResponse = (
+        TrainingInfoResponse.from_proto(
+            training_management_stub.GetTrainingStatus(training_info_request.to_proto())
+        )
+    )
+
+    assert training_management_response.status != TrainingStatus.ERRORED.value
+
+    # cancel the training
```
gabe-l-hart, line 636: If we update the logic to not return `CANCELED` when it's already `COMPLETED`, this test will have a race condition since there's no synchronization of the training job

gabe-l-hart (changes requested):
I think your changes are actually surfacing an interesting bug where a `ModelFuture` that has reached `COMPLETED` can be moved to `CANCELED` if the user cancels it, even if it completed successfully. Since this will be the first "public" exposure of the cancelation (in the runtime), I think we should look into fixing that before merging. I'll take a look.

tests/runtime/servicers/test_training_management_servicer.py
aluu317: I think what this is doing (I copied the `wait_event` and `model_future.wait()` from another test) is the `wait_event` is telling the model's train [function](https://github.com/caikit/caikit/blob/main/tests/fixtures/sample_lib/modules/sample_task/sample_implementation.py#L125) "wait, don't complete the train". And then the model_future.wait() is waiting for that training to complete and therefore will not reach "COMPLETED" state? So I can do CancelTraining here while it's still in state RUNNING.

Do you expect the `model_future.wait()` to make it `COMPLETED`?

gabe-l-hart:
PR to fix the incorrect `COMPLETED` -> `CANCELED` transition: https://github.com/caikit/caikit/pull/367

tests/runtime/test_grpc_server.py
aluu317: true. In the case of user trying to cancel a `COMPLETED` job, looks like it [does nothing](https://github.com/caikit/caikit/blob/main/caikit/core/toolkit/destroyable_thread.py#L203) and returns nothing, so I expect for the status of the response here to come back remained `COMPLETED`.
But you are right, there could be race condition that it is `COMPLETED` and not `CANCELED`... hmm... I'm not sure how to test that it can be canceled here.

gabe-l-hart: The problem is that right above that line, it sets `self.__destroyed = True`, so even though it doesn't _do_ a destroy operation, it still marks itself as `destroyed`. That was the reason I added the `canceled` property to distinguish between `canceled` (destroyed before completing) and `destroyed` (destroy was called at _some_ point).

For tests, I think what you really want to do is use two separate events, one that the train function sets to indicate that it's started, and one that the train function waits on before completing. Then, the test waits on the "make sure training started" event, calls cancel, and ensures that the job terminates without setting the "block training completion" one. You'll probably have to add some more args to the sample train module to make this all work

tests/runtime/test_grpc_server.py
```diff
@@ -628,6 +628,108 @@ def test_train_fake_module_ok_response_with_datastream_csv_file(
...
+            TrainingInfoResponse.from_proto(
+                training_management_stub.GetTrainingStatus(
+                    training_info_request.to_proto()
+                )
+            )
+        )
+        status = training_management_response.status
+        assert status != TrainingStatus.ERRORED.value
+        i += 1
+        assert i < 100, "Waited too long for training to complete"
```
joerunde, line 728: nit: this is copied from the `is_good_train_response` helper method, is ther a reason we can't call that?

Also while you're at it, some potential upgrades to that might be to:
1. rename that function to be a bit better
```
def wait_for_training_to_complete(...)
def wait_for_training_and_assert_successful(...)
def assert_training_successful(...)
```

2. Try to use the model future's `.wait()` function directly instead of a polling loop. For grpc server tests that would require digging down one extra layer to be able to grab the model future and interact with it, so we should probably still make a final `GetTrainingStatus` call to make sure the server is wired up properly.

aluu317: done and done!

tests/runtime/servicers/test_training_management_servicer.py
```diff
@@ -75,6 +75,69 @@ def test_training_runs(training_management_servicer, training_pool):
...
+    )
+
+    # Cancel first training
+    request_1 = TrainingInfoRequest(training_id=model_future_1.id).to_proto()
+    training_management_servicer.CancelTraining(request_1, context=None)
+
+    response_1 = training_management_servicer.GetTrainingStatus(request_1, context=None)
+    assert response_1.state == TrainingStatus.CANCELED.value
+
+    # release the blocked training that is waiting to clean up
```
gabe-l-hart, line 131: I'm still not sure this is doing what we expect. You shouldn't need to run this in a separate thread and you shouldn't need to do `event_1.set()` since the cancelation _should_ terminate the blocking `wait` in the training function.

gabe-l-hart (commented):
I think we need one more change to get the test right. I can take a look at this

gabe-l-hart (approved):
We're in business!

Issue #363, opened 2023-07-25
Support S3 pointers for data streams


joerunde:
## Additional context

Add any other context about the feature request here.

Issue #300, opened 2023-07-05
Convert DataStreamSource dynamic DM objects to static objects


prashantgupta24:
## Describe the bug

`DataStreamSource` creates a bunch of dynamically generated `DataObjects`, which are used for instantiating different flavors of`DataStreamSource`.

Ideally we want to create less of such dynamically created `DataObjects`, and more of statically typed `DataObjects`, which can live under `caikit.interfaces.common.data_model`.

Something similar has been done in this PR: https://github.com/caikit/caikit/pull/294

## Platform

Please provide details about the environment you are using, including the following:

- Interpreter version: 3.9
- Library version: 0.9.0

PR #365, opened 2023-07-25
✨ S3 streams

joerunde:
Implements #363 , https://github.com/caikit/caikit/issues/300

This PR adds an `S3Files` source for datastreams.
This is not implemented in the core of caikit yet, we don't have support for pulling files from S3. But it's left open to extension for other projects to implement as needed.

This also removes the dynamic generation of the `Files` `ListOfFiles` and `Directory` data models, in favor of static class definitions.

### DIFF
--- a/caikit/interfaces/common/data_model/stream_sources.py
+++ b/caikit/interfaces/common/data_model/stream_sources.py
@@ -57,30 +57,4 @@ class S3Files(DataObjectBase):
 
     # IAM credentials
     IAM_id: Annotated[str, FieldNumber(7)]
-    IAM_api_ky: Annotated[str, FieldNumber(8)]
-
-    # TLS info
-    CA_bundle_key: Annotated[str, FieldNumber(9)]
-
-
-# File = make_dataobject(
-#             package=package,
-#             proto_name=f"{cls_name}File",
-#             name="File",
-#             attrs={"__qualname__": f"{cls_name}.File"},
-#             annotations={"filename": str},
-#         )
-#         ListOfFiles = make_dataobject(
-#             package=package,
-#             proto_name=f"{cls_name}ListOfFiles",
-#             name="ListOfFiles",
-#             attrs={"__qualname__": f"{cls_name}.ListOfFiles"},
-#             annotations={"files": List[str]},
-#         )
-#         Directory = make_dataobject(
-#             package=package,
-#             proto_name=f"{cls_name}Directory",
-#             name="Directory",
-#             attrs={"__qualname__": f"{cls_name}.Directory"},
-#             annotations={"dirname": str, "extension": str},
-#         )
+    IAM_api_key: Annotated[str, FieldNumber(8)]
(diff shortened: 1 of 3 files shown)


gkumbhat (approved):
LGTM

caikit/interfaces/common/data_model/stream_sources.py
```diff
@@ -0,0 +1,69 @@
...
+
+    # IAM credentials
+    IAM_id: Annotated[str, FieldNumber(7)]
+    IAM_api_key: Annotated[str, FieldNumber(8)]
+
+
+@dataobject(PACKAGE_COMMON)
+class S3Files(S3Base):
+    # List of file paths relative to the bucket
+    files: Annotated[List[str], FieldNumber(1)]
```
gabe-l-hart, line 63: I'm curious why you chose to keep `FieldNumber(1)` as the one that is different across "real" implementations. Very not important, though.

caikit/runtime/service_generation/data_stream_source.py
```diff
@@ -241,13 +238,15 @@ def make_data_stream_source(data_element_type: Type) -> Type[DataBase]:
                 File.__name__: File,
                 ListOfFiles.__name__: ListOfFiles,
                 Directory.__name__: Directory,
+                S3Files.__name__: S3Files,
```
gabe-l-hart, line 241: Any reason you didn't also include `S3Path`?

gabe-l-hart (approved):
Looks great. Much nicer to have the ones that can be statically declared done that way!

caikit/interfaces/common/data_model/stream_sources.py
joerunde: I figured that `S3Base` would probably need changing / extending later, but this one probably wouldn't, so I stuck this first

caikit/runtime/service_generation/data_stream_source.py
joerunde: We thought `S3Path` would be useful for the `output_path` parameter where you only want to allow one path to be passed. But for an input stream, a user should be free to specify multiple.

I'm fine either way if we want to add both here similar to `File` and `ListOfFiles`

tests/runtime/service_generation/test_data_stream_source.py
```diff
@@ -110,10 +111,6 @@ def validate_data_stream(data_stream, length, data_item_type, data_item_length=N
 
 
 def test_make_data_stream_source_type_name():
-    assert "DataStreamSourceInt" == _make_data_stream_source_type_name(int)
-    assert "DataStreamSourceFloat" == _make_data_stream_source_type_name(float)
-    assert "DataStreamSourceStr" == _make_data_stream_source_type_name(str)
-    assert "DataStreamSourceBool" == _make_data_stream_source_type_name(bool)
```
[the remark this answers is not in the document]
joerunde: faster tests! :D

Hah, real talk though I thought these were pre-generated in interfaces/primitive_sequences but I see that those are just the `IntSequence` etc. ones. I can put these back, unless we want to statically declare them as well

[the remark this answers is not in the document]
prashantgupta24: Yeah these ones are dynamically generated in case we get an argument like `training_data: DataStream[int]` (I put one in `other_implementation.py`)

PR #366, opened 2023-07-25
Server shutdown with sync

### LINKED ISSUE #364: HTTP Server Shutdown with Local Model Dirs Sync
## Describe the bug

When running `caikit.runtime.http_server` with `runtime.lazy_load_local_models: true`, the server fails to terminate on ctrl-C due to the polling timer.

gabe-l-hart:
## Description

Closes #364

This PR updates the shutdown logic in `grpc_server` and `http_server` to correctly shut down the runtime Model Manager's poll timer for lazy model loading. While I was working on this, I discovered several other bugs that needed fixing:

1. Runtime was using `runtime.port` instead of `runtime.grpc.port`
2. The prometheus metrics were _always_ booting up in the `grpc_server`'s `main()`, regardless of `runtime.metrics.enabled`

### DIFF
--- a/caikit/config/config.yml
+++ b/caikit/config/config.yml
@@ -92,9 +92,6 @@ runtime:
         client:
             cert: ""
 
-    # Service exposure options
-    unix_socket_path: /tmp/mmesh/grpc.sock
-
     # Configuration for the grpc server
     grpc:
         enabled: true
--- a/caikit/config/config.yml
+++ b/caikit/config/config.yml
@@ -92,6 +92,9 @@ runtime:
         client:
             cert: ""
 
+    # Service exposure options
+    unix_socket_path: /tmp/mmesh/grpc.sock
+
     # Configuration for the grpc server
     grpc:
         enabled: true
@@ -104,9 +107,6 @@ runtime:
         server_shutdown_grace_period_seconds: 45
         # Listen on a unix socket (model mesh feature)
         unix_socket_path: /tmp/mmesh/grpc.sock
-    # Service exposure options
-    port: 8085
-    unix_socket_path: /tmp/mmesh/grpc.sock
 
     # Configuration for the http server
     http:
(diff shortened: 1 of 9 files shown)


caikit/config/config.yml
```diff
@@ -92,6 +92,9 @@ runtime:
         client:
             cert: ""
 
+    # Service exposure options
+    unix_socket_path: /tmp/mmesh/grpc.sock
```
joerunde, line 96: this is duplicated in `runtime.grpc.unix_socket_path`?

caikit/runtime/server_base.py
```diff
@@ -75,6 +77,10 @@ def _find_port(cls, start=8888, end=None, host="127.0.0.1"):
                 # port is open
                 return start
 
+    def _shut_down_model_manager(self):
+        """Shared utility for shutting down the model manager"""
+        ModelManager.get_instance().shut_down()
```
joerunde, line 82: this is interesting....

we're shutting down an instance of a runtime server, and in the process shutting down a singleton model manager?

Super fine in the context of a container that is running the runtime entrypoint, but this seems like it could muck with tests where a server that shuts down would kill the singleton poll. I think we have plenty of these hanging around:
```
with runtime_test_server(...) as runtime:
```

How do those not just kill the shared model manager's poll?

joerunde (approved):
lgtm with the caveat that I don't understand how the shutdown stuff is safe with the temporary runtime servers in tests, but if it works it works and I don't think we should use this PR to wrestle with the lifetimes of all the caikit core objects

caikit/runtime/server_base.py
gabe-l-hart: ug, you're probably right. I really hate that we made everything a singleton! There's no reason for these to be singletons and it just screws with unit tests badly

gabe-l-hart: I think the reason that this works is that `shut_down()` only stops the polling and the tests that actually test the polling explicitly create temporary `ModelManager` instances (for exactly this need to isolate them)

PR #373, opened 2023-07-26
:recycle: Move training info out of interfaces

gkumbhat:
1.  Move `TrainingInfo` down to `caikit.core` as part of the ModelTrainer / ModelFuture api
2. Make trainers return exception instead of returning error strings

**If applicable**:

### DIFF
--- a/caikit/core/model_management/model_trainer_base.py
+++ b/caikit/core/model_management/model_trainer_base.py
@@ -26,7 +26,7 @@ model_management:
 """
 
 # Standard
-from typing import Optional, Type
+from typing import List, Optional, Type
 import abc
 import dataclasses
 import os
@@ -41,7 +41,7 @@ from ..toolkit.reversible_hasher import ReversibleHasher
 @dataclasses.dataclass
 class TrainingInfo:
     status: TrainingStatus
-    errors: Optional[Exception] = None
+    errors: Optional[List[Exception]] = None
 
 
 class ModelTrainerBase(FactoryConstructible):
--- a/caikit/core/model_management/model_trainer_base.py
+++ b/caikit/core/model_management/model_trainer_base.py
@@ -28,13 +28,20 @@ model_management:
 # Standard
 from typing import Optional, Type
 import abc
+import dataclasses
 import os
 
 # Local
-from ..data_model import TrainingInfo
+from ..data_model import TrainingStatus
 from ..modules import ModuleBase
 from ..toolkit.factory import FactoryConstructible
 from ..toolkit.reversible_hasher import ReversibleHasher
+
+
+@dataclasses.dataclass
+class TrainingInfo:
+    status: TrainingStatus
+    errors: Optional[Exception] = None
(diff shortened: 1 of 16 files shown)


caikit/runtime/servicers/global_train_servicer.py
```diff
@@ -262,7 +262,7 @@ def rpc_termination_callback():
                     )
                     if isinstance(model_future, LocalModelTrainer.LocalModelFuture):
                         raise model_future._worker.error
-                    raise RuntimeError(training_info.errors[0])
+                    raise training_info.errors[0]
```
joerunde, line 265: can you fix the `_worker` stuff above?

joerunde: (it should no longer be needed)

gkumbhat: We are still not raising the error in `wait()`. so it will still be needed?  Or you mean the if condition at line 264?

caikit/core/model_management/model_trainer_base.py
```diff
@@ -28,15 +28,22 @@
...
+from ..data_model import TrainingStatus
 from ..modules import ModuleBase
 from ..toolkit.factory import FactoryConstructible
 from ..toolkit.reversible_hasher import ReversibleHasher
 
 
+@dataclasses.dataclass
+class TrainingInfo:
+    status: TrainingStatus
+    errors: Optional[Exception] = None
```
joerunde, line 44: @gkumbhat do we want a list or a single one here? We should choose and make the plural-ness of the noun here match.

I think the only reason we came up with to go with a list of exceptions was for the case of data validation where we may want to have a big old list of exceptions for each spot where there's a problem with the user's data.

```
@dataclasses.dataclass
class TrainingInfo:
    status: TrainingStatus
    exception: Optional[Exception] = None
    # or
    exceptions: List[Exception]  # can be empty
```

caikit/runtime/servicers/training_management_servicer.py
```diff
@@ -45,9 +45,14 @@ def GetTrainingStatus(self, request, context):  # pylint: disable=unused-argumen
                 training_id=training_info.training_id
             )
 
+            reasons = []
+            if model_future.get_info().errors:
+                reasons = [str(error) for error in model_future.get_info().errors]
```
joerunde, line 50: nit that `CodeQL` is gonna barf on this since `str(exception)` could possibly include stack trace info. See this: https://github.com/caikit/caikit/security/code-scanning/4

I think we may need something a little bit more fancy here to grab just the relevant error message off of the exception. We do a pretty good job in the inference stack of catching error cases and translating to appropriate `CaikitRuntimeException`s, maybe the training stack needs a bit more love as well?

I could be fine with merging this since I think stack trace exposure of open source software isn't extremely bad, but we should think on this

joerunde (approved):
lpgtm!

Issue #343, opened 2023-07-19
Config cleanup post-REST-merge


joerunde:
When we merged in REST support we missed a bit of cleanup:

- `runtime.port` still in base config
  - `runtime.port` is still used for the grpc server instead of `runtime.grpc.port`
- `runtime.unix_socket_path` still in base config
- `find_available_ports` made its way back in after being killed

The base runtime server should not have a private method to find open ports. We moved that to an `open_port` test fixture since it's only used for unit testing. See this PR for where we made the change: https://github.com/caikit/caikit/pull/302/files

PR #375, opened 2023-07-26
:recycle: Only find open ports on tests

evaline-ju:

`find_available_ports` cleanup (round 2) for #343 in line with the changes made in PR #302 

As described, the base runtime server should not have a private method to find open ports. This should only be done for unit tests, where an `open_port` fixture exists.

**If applicable**:


gabe-l-hart (approved):
Thank you! So glad to see this removed from the production path

PR #376, opened 2023-07-27
♻️ refactor init for http (see #389 instead)

prashantgupta24:

This PR refactors the `init` function of the `http_server` to align with `grpc_server`. The idea is to eventually have a common `caikt/runtime/__main__.py` which can start both servers with similar arguments.

Alternate implementation here:  https://github.com/caikit/caikit/pull/389

Also fixes https://github.com/caikit/caikit/issues/380

**If applicable**:

### DIFF
--- a/caikit/runtime/http_server.py
+++ b/caikit/runtime/http_server.py
@@ -557,15 +557,29 @@ class RuntimeHTTPServer(RuntimeServerBase):
 
 
 def main():
+
     caikit.core.toolkit.logging.configure()
+    log.debug("Starting up caikit.runtime.http_server")
 
     # This code will live in caikt/runtime/__main__.py eventually
-    inference_service = ServicePackageFactory().get_service_package(
+    # pylint: disable=duplicate-code
+
+    # We should always be able to stand up an inference service
+    inference_service: ServicePackage = ServicePackageFactory.get_service_package(
         ServicePackageFactory.ServiceType.INFERENCE,
     )
-    training_service = ServicePackageFactory().get_service_package(
-        ServicePackageFactory.ServiceType.TRAINING,
-    )
+
+    # But maybe not always a training service
+    try:
+        training_service: Optional[
+            ServicePackage
+        ] = ServicePackageFactory.get_service_package(
+            ServicePackageFactory.ServiceType.TRAINING,
+        )
+    except CaikitRuntimeException as e:
+        log.warning("Cannot stand up training service, disabling training: %s", e)
+        training_service = None
+
     server = RuntimeHTTPServer(
         inference_service=inference_service, training_service=training_service
     )
--- a/caikit/runtime/http_server.py
+++ b/caikit/runtime/http_server.py
@@ -106,15 +106,18 @@ class RuntimeHTTPServer(RuntimeServerBase):
     ## Interface ##
     ###############
 
-    def __init__(self, tls_config_override: Optional[aconfig.Config] = None):
+    def __init__(
+        self,
+        inference_service: ServicePackage,
+        #  # pylint: disable=unused-argument
+        training_service: Optional[ServicePackage],
+        tls_config_override: Optional[aconfig.Config] = None,
+    ):
         super().__init__(get_config().runtime.http.port, tls_config_override)
 
         self.app = FastAPI()
 
         # Set up the central predict servicer
-        inference_service = ServicePackageFactory().get_service_package(
-            ServicePackageFactory.ServiceType.INFERENCE,
-        )
         self.global_predict_servicer = GlobalPredictServicer(inference_service)
 
         # Set up the central train servicer
@@ -178,12 +181,8 @@ class RuntimeHTTPServer(RuntimeServerBase):
         # Placeholder for thread when running without blocking
         self._uvicorn_server_thread = None
 
-    def __del__(self):
-        if get_config().runtime.metering.enabled:
-            self.global_predict_servicer.stop_metering()
-
     def start(self, blocking: bool = True):
-        """Boot the gRPC server. Can be non-blocking, or block until shutdown
+        """Boot the http server. Can be non-blocking, or block until shutdown
 
         Args:
             blocking (boolean): Whether to block until shutdown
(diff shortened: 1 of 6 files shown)


caikit/runtime/http_server.py
```diff
@@ -106,15 +106,18 @@ class RuntimeHTTPServer(RuntimeServerBase):
     ## Interface ##
     ###############
 
-    def __init__(self, tls_config_override: Optional[aconfig.Config] = None):
+    def __init__(
+        self,
+        inference_service: ServicePackage,
+        #  # pylint: disable=unused-argument
```
gabe-l-hart, line 112: This syntax looks odd

gabe-l-hart: Also, if these are `Optional` should they have a default value of `None`?

caikit/runtime/http_server.py
```diff
@@ -106,15 +106,18 @@ class RuntimeHTTPServer(RuntimeServerBase):
...
+        training_service: Optional[ServicePackage],
+        tls_config_override: Optional[aconfig.Config] = None,
+    ):
         super().__init__(get_config().runtime.http.port, tls_config_override)
 
         self.app = FastAPI()
 
         # Set up the central predict servicer
-        inference_service = ServicePackageFactory().get_service_package(
```
gabe-l-hart, line 115: What are the `Optional` semantics on these args supposed to mean? I could see the answer being either "use the global singleton" or "run without this service"

caikit/runtime/http_server.py
```diff
@@ -178,12 +181,8 @@ def shutdown_wrapper(*args, **kwargs):
         # Placeholder for thread when running without blocking
         self._uvicorn_server_thread = None
 
-    def __del__(self):
```
gabe-l-hart, line 181: Did this move somewhere?

tests/runtime/conftest.py
```diff
@@ -149,6 +161,47 @@ def runtime_grpc_server(
...
+            },
+            "merge",
+        ):
+            with http_server.RuntimeHTTPServer(*args, **kwargs) as server:
+                # Give tests access to the workdir
+                server.workdir = workdir
+                yield server
+
+
+# This is honestly not required for fastAPI testing
```
gabe-l-hart, line 188: If we don't actually need this, let's not keep it in. That said, we _do_ want to be able to test the real server behavior to ensure that our server booting and binding works (particularly with (m)TLS). I know we have tests for that somewhere, so either they should use this, or we should remove this if there's a compelling reason to not use this (which I think there is since they want the server to boot/terminate in isolation).

tests/runtime/conftest.py
```diff
@@ -323,3 +376,21 @@ def _check_server_readiness(server):
...
+                f"http://localhost:{server.port}{http_server.HEALTH_ENDPOINT}"
+            )
+            assert response.status_code == 200
+            assert response.text == "OK"
+            done = True
+        except AssertionError:
+            log.debug(
+                "[HTTP server not ready]; will try to reconnect to test server in 0.1 second."
+            )
+            time.sleep(0.1)
```
gabe-l-hart, line 396: Channeling Joe, but let's make this `0.01` so that it those tenths of a second don't add up!

tests/runtime/conftest.py
gabe-l-hart: Ok, I see that you've refactored those tests to do `runtime_http_test_server`, so let's remove this one

tests/runtime/conftest.py
```diff
@@ -149,6 +161,47 @@ def runtime_grpc_server(
...
+                    "metering": {"log_dir": temp_log_dir},
+                    "training": {"output_dir": temp_save_dir},
+                    "http": {"port": open_port},
+                }
+            },
+            "merge",
+        ):
+            with http_server.RuntimeHTTPServer(*args, **kwargs) as server:
+                # Give tests access to the workdir
+                server.workdir = workdir
```
gabe-l-hart, line 184: I think we should throw `_check_http_server_readiness(server)` in here before the `yield`

gabe-l-hart: That said, we'd need to update `_check_http_server_readiness` to support TLS too

gabe-l-hart (changes requested):
I think we need to clarify the `Optional` semantics of the services in the server's init (and possibly in the grpc server too)

caikit/runtime/http_server.py
prashantgupta24: Hmm, I tried to use the same one we have for `grpc_server`?
```
    def __init__(
        self,
        inference_service: ServicePackage,
        training_service: Optional[ServicePackage],
```

prashantgupta24: But yeah the `optional` should probably be None in both places

caikit/runtime/http_server.py
prashantgupta24: I assumed they meant that we _can_ start without a `trainer_service` - which makes sense since we might not always have training

tests/runtime/conftest.py
prashantgupta24: I'm thinking I can actually add a test case that uses this to make actual REST calls to the server?

caikit/runtime/http_server.py
prashantgupta24: So, having this around was giving an issue after a successful test run in which the `flush_metrics` was trying to write to a file that was already closed. 
```
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/logging/__init__.py", line 1082, in emit
    stream.write(msg + self.terminator)
ValueError: I/O operation on closed file.
```

I figured if we want to keep this, we should have a `flag` in `global_predict_servicer.stop_metering()` that checks if we've already stopped metering before trying to `flush_metrics`?

caikit/runtime/servicers/global_predict_servicer.py
```diff
@@ -95,8 +95,10 @@ def __init__(
         inference_service: ServicePackage,
         use_abortable_threads: bool = get_config().runtime.use_abortable_threads,
     ):
+        self._started_metering = False
```
prashantgupta24, line 98: This flag is added to make sure we don't try to `stop_metering` multiple times and invoke `self.rpc_meter.flush_metrics` multiple times, as it gives `ValueError: I/O operation on closed file.` when invoked if the metering file is already closed

tests/runtime/conftest.py
prashantgupta24: This also removes the duplicated code snippet from all tests
```
server = http_server.RuntimeHTTPServer()
```

Issue #369, opened 2023-07-26
Nest training parameters in runtime API, add output path


joerunde:
## Additional context

Add any other context about the feature request here.

PR #378, opened 2023-07-27
Add output_path to train apis

joerunde:

For #369 

This PR adds `output_path: S3Path` to all train apis

This PR _does not_ nest the rest of the training parameters. That turned out to be a bit of a bear so we're following up in a separate PR.

We went with just an `S3Path` for now and not also a `str` because we already have pretty good on-disk-path support via the `runtime.training.output_dir` configuration. That allows a deployment-wide base path to be set and then the `model_name` parameter provides the directory separation within that base path on a per-training-request basis. We think some more design thought is needed if we want to start overriding disk paths directly from the training API.

This change updates the base trainer implementation to fail if an S3 path is passed to the logic that builds disk paths- we expect trainers that support S3 paths to do their own path handling.

**If applicable**:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 __pycache__
 .coverage
 .coverage.*
+durations/*
 coverage*.xml
 dist
 htmlcov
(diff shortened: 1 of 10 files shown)


gkumbhat (approved):
LGTM. 🚢

PR #379, opened 2023-07-27
✨ Introduce a common start script  (see #389 instead)

prashantgupta24:

Based on https://github.com/caikit/caikit/pull/376

From the official docs:
```The contents of __main__.py typically isn’t fenced with if __name__ == '__main__' blocks. Instead, those files are kept short, functions to execute from other modules. Those other modules can then be easily unit-tested and are properly reusable.```

**If applicable**:

### CODE REVIEW

File: caikit/runtime/__main__.py [hunk 1]
```diff
@@ -0,0 +1,97 @@
...
+from caikit.runtime.grpc_server import RuntimeGRPCServer
+from caikit.runtime.http_server import RuntimeHTTPServer
+from caikit.runtime.service_factory import ServicePackage, ServicePackageFactory
+from caikit.runtime.types.caikit_runtime_exception import CaikitRuntimeException
+import caikit
+
+log = alog.use_channel("RUNTIME-MAIN")
+
+# Make sample_lib available for import, useful for local testing
+sys.path.append("tests/fixtures")
```
prashantgupta24, line 24: tmp obviously

caikit/runtime/__main__.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+# Start serving http server
+if caikit_config.runtime.http.enabled:
+    log.debug("Starting up caikit.runtime.http_server")
+
+    http_server = RuntimeHTTPServer(
+        inference_service=inference_service, training_service=training_service
+    )
+    http_server.start(blocking=not caikit_config.runtime.grpc.enabled)
```
prashantgupta24, line 86: we want http to be blocking in case grpc is disabled

Issue #381, opened 2023-07-28
Handle model_name in trainers


joerunde:
## Is your feature request related to a problem? Please describe.

Today the global train servicer is responsible for appending the training output directory with the user-provided model_name.

This is gnarly because then the trainer implementations need to assume that a model name _may have been_ appended to a save path, and contain logic to deal with that when figuring out where they need to save. This gets harder when a trainer wants to deal with some remote storage situation like uploading to s3.

## Describe the solution you'd like

The `ModelTrainer`s should have authority over how to edit the base save path that they are given, based on the training id and model name and other configs relative to the trainer.

## Additional context

Add any other context about the feature request here.

PR #382, opened 2023-07-28
:sparkles: Add model name to Trainer API

joerunde:

For #381 

This PR adds `model_name` as an optional top-level arg to the model trainer's `.train` interface.

This allows model trainer implementations to construct their own save paths based on
- the base path given
- the training id
- the model name

The abstract base class is updated so that the save path is constructed as
```
${save_path}/${training_id}/${model_name}
```

This should not be a breaking change since the parameter is optional with a default of none

(except maybe how the .train function may have keyword collisions with actual training kwargs but... /shrug)

**If applicable**:


gabe-l-hart (approved):
Nice, and nice that it's not a breaking change

PR #388, opened 2023-08-01
Use spawn for train subprocess 384

### LINKED ISSUE #384: Support "spawn" for training subprocesses
## Describe the bug

Some libraries such as `torch` have significant issues when running in a `subprocess` using the `"fork"` method. The current `DestroyableThread` [explicitly uses "fork"](https://github.com/caikit/caikit/blob/main/caikit/core/toolkit/destroyable_process.py#L18) to avoid problems with data model object not being recreated correctly, but this can break downstream usage of libraries such as `torch`.

gabe-l-hart:

Closes #384 

This PR makes it safe to use "spawn" when performing training jobs.

**If applicable**:

### DIFF
--- a/caikit/core/model_management/local_model_trainer.py
+++ b/caikit/core/model_management/local_model_trainer.py
@@ -271,6 +271,16 @@ class LocalModelTrainer(ModelTrainerBase):
         # Always purge old futures
         self._purge_old_futures()
 
+        # Wrap any models in the kwargs for safe spawning if needed
+        if self._use_subprocess and self._subprocess_start_method != "fork":
+            wrapped_models = {
+                key: _SpawnProcessModelWrapper(val)
+                for key, val in kwargs.items()
+                if isinstance(val, ModuleBase)
+            }
+            log.debug2("Subprocess wrapped models: %s", wrapped_models.keys())
+            kwargs.update(wrapped_models)
+
         # Create the new future
         model_future = self.LocalModelFuture(
             self._instance_name,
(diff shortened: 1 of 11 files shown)


caikit/runtime/utils/servicer_util.py
```diff
@@ -260,6 +269,140 @@ def validate_data_model(
...
+            from caikit.runtime.service_factory import (  # pylint: disable=import-outside-toplevel
+                ServicePackageFactory,
+            )
+
+            ServicePackageFactory.get_service_package(
+                ServicePackageFactory.ServiceType.INFERENCE
+            )
+            ServicePackageFactory.get_service_package(
+                ServicePackageFactory.ServiceType.TRAINING
+            )
```
joerunde, line 338: I love it so much and yet hate it at the same time 🙃

caikit/runtime/utils/servicer_util.py
```diff
@@ -260,6 +269,140 @@ def validate_data_model(
...
+
+    def __getstate__(self) -> Tuple[str, Optional[bytes]]:
+        """When pickling, only send the serialized model body for non-fork. This
+        is not a general-purpose pickle solution for models, but makes them safe
+        for training jobs that need to move models between processes.
+        """
+        start_method = multiprocessing.get_start_method()
+        body = None
+        if start_method != "fork":
+            body = self._model.as_bytes()
```
joerunde, line 386: nice!

Though this also requires that models implement `.save()` which they might not

caikit/core/model_management/local_model_trainer.py
```diff
@@ -98,11 +100,13 @@ def __init__(
...
             # Set up the worker and start it
             self._use_subprocess = use_subprocess
+            self._subprocess_start_method = subprocess_start_method
             if self._use_subprocess:
                 log.debug2("Running training %s as a SUBPROCESS", self.id)
                 self._worker = DestroyableProcess(
+                    start_method=self._subprocess_start_method,
                     target=self._train_and_save,
-                    return_result=True,
+                    return_result=False,
```
joerunde, line 109: Is this flag required in the implementation of `DestroyableProcess` at all?
Currently our only usage of `DestroyableProcess` is right here in the local trainer and I didn't think we had other immediate use cases for this.

gabe-l-hart: It's not _except_ if we want `DestroyableProcess` to match `DestroyableThread` which was the goal of adding it in the first place

caikit/core/model_management/local_model_trainer.py
```diff
@@ -277,6 +295,7 @@ def train(
             save_with_id=save_with_id,
             external_training_id=external_training_id,
             use_subprocess=self._use_subprocess,
+            subprocess_start_method=self._subprocess_start_method,
```
joerunde, line 298: Do we need to be doing something different that's not just `**kwargs` here to pass along the training kwargs?

If I already had `subprocess_start_method` as a kwarg for the train method, then this would raise something like `got multiple values for argument subprocess_start_method`.
Probably not a huge deal but we're starting to put quite a few things alongside the training parameters and I'd think we should set out a good pattern in the local trainer for other implementations to reference

tests/conftest.py
```diff
@@ -49,12 +49,26 @@ def test_environment():
...
-    # Local
-    from tests.core.helpers import MockBackend
-
-    yield
+    with tempfile.TemporaryDirectory() as workdir:
+        with temp_config(
+            {
+                "runtime": {
+                    "metering": {
+                        "log_dir": os.path.join(workdir, "metering_logs"),
```
joerunde, line 57: could use a 🌶️ comment here!

caikit/core/model_management/local_model_trainer.py
joerunde: gotcha...
I'm not sure we need it to entirely match, and I wouldn't expect most things to be returnable from a subprocess anyway? /shrug

I'm fine either way, but I lean towards removing so that nobody stubs their toe in the future expecting this to return a loaded model, like we did today 😉

joerunde (approved):
lpgtm, thanks!

caikit/core/model_management/local_model_trainer.py
gabe-l-hart: Hm, totally valid. I'll compromise and make it default to False so that you need to opt in

caikit/core/model_management/local_model_trainer.py
gabe-l-hart: Very good call. Pretty easy fix to just take `args`/`kwargs` as explicit arguments and not do `*`/`**` until the underlying training is invoked

caikit/runtime/utils/servicer_util.py
gabe-l-hart: 😬

Issue #380, opened 2023-07-28
Dump services should be controlled by configs


prashantgupta24:
## Describe the bug

Right now, calling `dump_services` module tries to dump both `grpc` and `http` services regardless of whether the user has enabled both services in their global config.

Ideally, `caikit_config.runtime.http.enabled` and `caikit_config.runtime.grpc.enabled` flags should control what service should be dumped.

## Platform

Please provide details about the environment you are using, including the following:

- Interpreter version: 3.9
- Library version: 0.13.x

## Additional context

The fix should be simple, checking the configs [here](https://github.com/caikit/caikit/blob/main/caikit/runtime/dump_services.py#L74) before proceeding

Issue #225, opened 2023-06-12
Rework entrypoints and config for REST support


joerunde:
## Is your feature request related to a problem? Please describe.

We currently just have a `caikit.runtime.grpc_server::main` function that we use to boot the runtime. We'll also want the ability to run a REST server, and we'll probably also want to be able to run both side by side.

## Describe the solution you'd like

Wrapping up our runtime boot-up logic into something like `caikit.runtime.main.py` and having a solid set of config options to enable what you want sounds pretty nice.

## Describe alternatives you've considered

Simply having a rest/main entrypoint

## Additional context

The current grpc main will just block on the server and let the server catch shutdown signals and stop. If we wanna run two servers, we'd need to not block on them and add a little mechanism to wait for a shutdown signal

PR #389, opened 2023-08-01
✨ Introduce a common start script

prashantgupta24:

For: https://github.com/caikit/caikit/issues/225

This PR:
- Refactors the init function of the `http_server` to align with `grpc_server`. 
- Moves a lot of common code into `server_base.py`.
- Introduces a common `caikt/runtime/__main__.py` which can start both servers based on configurations.
- Refactors the `text-sentiment` example to use this common script.
- Also fixes https://github.com/caikit/caikit/issues/380

**If applicable**:


caikit/runtime/grpc_server.py
```diff
@@ -297,43 +271,20 @@ def __exit__(self, type_, value, traceback):
 
 
 def main(blocking: bool = True):
-    # Configure using the log level and formatter type specified in config.
-    caikit.core.toolkit.logging.configure()
-    log.debug("Starting up caikit.runtime.grpc_server")
 
+    # pylint: disable=duplicate-code
```
prashantgupta24, line 275: need to figure out how to remove this duplicated code

caikit/runtime/http_server.py
```diff
@@ -179,11 +183,11 @@ def shutdown_wrapper(*args, **kwargs):
         self._uvicorn_server_thread = None
 
     def __del__(self):
-        if get_config().runtime.metering.enabled:
+        if get_config() and get_config().runtime.metering.enabled:
```
gabe-l-hart, line 186: Why do we need the truthiness check on `if get_config()`? Are there cases you've encountered where this results in an empty dict or `None`? If we _do_ need that check, we should only call `get_config()` once since it _might_ do some work in there some day.

```suggestion
        config = get_config()
        if config and config.runtime.metering.enabled:
```

caikit/runtime/server_base.py
```diff
@@ -39,6 +47,54 @@ def __init__(self, base_port: int, tls_config_override: Optional[aconfig.Config]
         )
         log.debug4("Full caikit config: %s", get_config())
 
+        # Configure using the log level and formatter type specified in config.
+        caikit.core.toolkit.logging.configure()
+
+        # We should always be able to stand up an inference service
```
gabe-l-hart, line 53: This is where we'd need to make these optional. Probably a follow-on PR though.

caikit/runtime/__main__.py
```diff
@@ -0,0 +1,58 @@
...
+            _grpc_server.stop()
+        if _http_server:
+            _http_server.stop()
+
+    # NOTE: signal function can only be called from main thread of the main
+    # interpreter. If this function is called from a thread (like in tests)
+    # then signal handler cannot be used. Thus, we will only have real
+    # termination_handler when this is called from the __main__.
+
+    signal.signal(signal.SIGINT, interrupt)
```
gabe-l-hart, line 35: This is interesting. I was surprised to learn that signals can only be bound once. As written this will work because the individual servers call `signal.signal` in their inits and these calls come afterwards to [override the previously registered handlers](https://docs.python.org/3/library/signal.html#signal.signal). This seems a bit confusing and relies on knowledge that `signal.signal` has overwrite behavior. I think it might be cleaner to NOT call `signal.signal` in the individual server inits and instead do those individual-server registrations in the `main` functions of those modules so that the pattern is "the code defines the `interrupt` method, the `main` registers the signals.

caikit/runtime/server_base.py
```diff
@@ -39,6 +47,54 @@ def __init__(self, base_port: int, tls_config_override: Optional[aconfig.Config]
...
+        if not cls._metrics_server_started and get_config().runtime.metrics.enabled:
+            log.info(
+                "Serving prometheus metrics on port %s",
+                get_config().runtime.metrics.port,
+            )
+            with alog.ContextTimer(log.info, "Booted metrics server in "):
+                start_http_server(get_config().runtime.metrics.port)
+            cls._metrics_server_started = True
+
+    def _intercept_interrupt_signal(self) -> None:
```
gabe-l-hart, line 85: Per comment in `__main__`, I think this should probably move out of this base class into the `main` functions of the respective server modules

tests/runtime/conftest.py
```diff
@@ -138,17 +150,59 @@ def runtime_grpc_test_server(open_port, *args, **kwargs):
...
+                config_overrides = kwargs["tls_config_override"]
+                kwargs["tls_config_override"] = config_overrides["runtime"]["tls"]
+            with http_server.RuntimeHTTPServer(*args, **kwargs) as server:
+                _check_http_server_readiness(server, config_overrides)
+                # Give tests access to the workdir
+                server.workdir = workdir
+                yield server
+
+
+# This is honestly not required for fastAPI testing
```
gabe-l-hart, line 191: If it's not required and not used, I'd be in favor of removing it to avoid unused code

tests/runtime/conftest.py
```diff
@@ -321,5 +375,39 @@ def _check_server_readiness(server):
...
+                cert=cert,
+            )
+            assert response.status_code == 200
+            assert response.text == "OK"
+            done = True
+        except AssertionError:
+            log.debug(
+                "[HTTP server not ready]; will try to reconnect to test server in 0.01 second."
+            )
+            time.sleep(0.01)
```
gabe-l-hart, line 413:
```suggestion
            time.sleep(0.001)
```

gabe-l-hart (commented):
I think this is great @prashantgupta24! I've got a bunch of little NITs/thoughts, but I think this is really close and could certainly be convinced to address them in a follow-on

caikit/runtime/__main__.py
prashantgupta24: Hmm. That's how I have it right now? The individual servers call `server._intercept_interrupt_signal()` within their `main` functions, and I made it like that to avoid running it when we run the `__main__` script (which just calls their init functions).

caikit/runtime/server_base.py
prashantgupta24: I have it in the base class to just avoid duplication of code, but this is called from within `main` functions of individual servers. Does that make sense code wise?

gabe-l-hart (approved):
Somehow I totally missed that those calls were in the main functions and not in the `__init__`s. Ship it!

Issue #390, opened 2023-08-02
Add core exception type for trainers, finders, initializers


joerunde:
## Is your feature request related to a problem? Please describe.

We now have more pluggable components in caikit for model management- trainers, finders, and initializers.

These may be implemented in a way that interacts with remote systems, e.g.
- Finding models in remote storage
- Initializing models on remote workers
- Training models on distributed workers

There are plenty of failure modes (bad credentials, unavailable remotes, resources not found, invalid arguments) that need to be communicated back to the user in a nice way, whether that user is a local python program or a caller of a caikit runtime server.
However we don't want to require that implementations of core resources use dependencies from `caikit.runtime`, nor do we want code in `caikit.core` that invokes these things require dependencies like `http` or `grpc`, so we don't want to pass around things like a `grpc.RpcError` or a `grpc.Status.StatusCode` or an `http.StatusCode`.

## Describe the solution you'd like

Kinda like how we have the `DataValidationError` that module authors can use when training data has an invalid format, we think we want something like a `CaikitCoreException` that has some easily-mappable enum that we can translate to valid grpc or http errors in our servicers, and local python programs can interpret.

## Additional context

We know more indirection and mapping isn't the greatest..... but we figure we will contain this to the `model_management` stuff at the moment.

PR #393, opened 2023-08-02
Add CaikitCoreException 

aluu317:
Related to #390 

**If applicable**:

### DIFF
--- a/caikit/core/exceptions/caikit_core_exception.py
+++ b/caikit/core/exceptions/caikit_core_exception.py
@@ -22,7 +22,7 @@ from enum import Enum
 class CaikitCoreStatusCode(Enum):
     NOT_FOUND = 1
     INVALID_ARGUMENT = 2
-    CONNECTION = 3
+    CONNECTION_ERROR = 3
     UNAUTHORIZED = 4
     FORBIDDEN = 5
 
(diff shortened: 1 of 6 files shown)


caikit/core/exceptions/caikit_core_exception.py
```diff
@@ -0,0 +1,36 @@
...
+"""
+
+# Standard
+from enum import Enum
+
+
+class CaikitCoreStatusCode(Enum):
+    NOT_FOUND = 1
+    INVALID_ARGUMENT = 2
+    CONNECTION = 3
```
joerunde, line 25: I might say `CONNECTION_ERROR` or `CONNECTION_LOST` or something to indicate that the connection is the problem, what do you think?

caikit/runtime/servicers/training_management_servicer.py
```diff
@@ -80,6 +82,8 @@ def CancelTraining(self, request, context):  # pylint: disable=unused-argument
                 state=model_future.get_info().status,
                 reasons=reasons,
             ).to_proto()
+        except CaikitCoreException as err:
```
joerunde, line 85: I think the one thing missing here is handling the issue where `model_future.get_info()` can return a 404 after `model_future.cancel()`. Would we want to catch that and just return `CANCELLED`?

It does feel a bit wonky to handle here since that's a quirk of some remote trainers, and we would have to write a mock or something to behave this way since the local trainer won't. Did you wanna tackle now or in a future PR?

joerunde (approved):
lpgtm!

Follow up PR for handling the cancellation 404 issue?

PR #397, opened 2023-08-03
Non disk lazy loading

### LINKED ISSUE #394: Allow lazy-loaded local models in runtime that use non-local finders

Currently, the `lazy_load_local_models` mechanism in `caikit.runtime` relies entirely on the presence of the local model on disk. This precludes the possibility of using a custom `ModelFinder` to "find" (or dynamically create) models by name. When using models that proxy to a remote such as TGIS, this requires that there be an on-disk `config.yml` file which requires manual effort from users.

To support non-local `ModelFinders`, we need to delegate [the on-disk check](https://github.com/caikit/caikit/blob/main/caikit/runtime/model_management/model_manager.py#L103) to the finder itself rather than asserting it in the `ModelManager`.

gabe-l-hart:

Closes #394 

This PR makes it possible to write a `ModelFinder` that finds models in non-disk locations. This could be fetching a model from a remote repository or a config-only model that does not need any load-time artifacts (e.g. remote proxy to TGIS)

### DIFF
--- a/caikit/core/model_management/local_model_initializer.py
+++ b/caikit/core/model_management/local_model_initializer.py
@@ -28,7 +28,8 @@ model_management:
                       config: {}
 """
 # Standard
-from typing import Callable, Optional
+from typing import Callable, List, Optional
+import copy
 import inspect
 
 # First Party
@@ -198,6 +199,11 @@ class LocalModelInitializer(ModelInitializerBase):
 
         # Return the loaded model if it was able to load
         return loaded_model
+
+    @property
+    def backends(self) -> List[BackendBase]:
+        """Return an immutable view of the priority sequence of backends"""
+        return copy.copy(self._backends)
 
     ## Implementation Details ##################################################
 
--- a/caikit/core/model_management/local_model_initializer.py
+++ b/caikit/core/model_management/local_model_initializer.py
@@ -180,13 +180,27 @@ class LocalModelInitializer(ModelInitializerBase):
                 extra_kwargs = {}
                 if self._supports_arg(module_backend_impl.load, "load_backend"):
                     extra_kwargs["load_backend"] = load_backend
-                if self._supports_arg(module_backend_impl.load, "model_config"):
-                    extra_kwargs["model_config"] = model_config
-                loaded_model = module_backend_impl.load(
-                    model_path,
-                    **extra_kwargs,
-                    **kwargs,
-                )
+                # Try loading with the ModuleConfig directly and fall back to
+                # loading with the model_path for compatibility
+                try:
+                    loaded_model = module_backend_impl.load(
+                        model_config,
+                        **extra_kwargs,
+                        **kwargs,
+                    )
+                except TypeError as err:
+                    log.warning(
+                        "<COR98539580W>",
+                        "DEPRECATION: Loading %s failed with ModuleConfig. Using model_path for compatibility. %s",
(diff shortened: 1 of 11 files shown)


caikit/runtime/model_management/model_manager.py
```diff
@@ -197,9 +197,16 @@ def load_model(
...
                     # Estimate the model's size and update the LoadedModel
-                    model_size = self.model_sizer.get_model_size(
-                        model_id, local_model_path, model_type
-                    )
+                    try:
+                        model_size = self.model_sizer.get_model_size(
+                            model_id, local_model_path, model_type
+                        )
+                    except CaikitRuntimeException:
+                        log.debug(
```
gkumbhat, line 205: hmm.. is this the only case when `get_model_size` would fail with runtime exception?

caikit/core/model_management/local_model_initializer.py
```diff
@@ -177,8 +177,10 @@ def init(self, model_config: ModuleConfig, **kwargs) -> Optional[ModuleBase]:
                     module_backend_impl.__name__,
                 )
                 extra_kwargs = {}
-                if self._supports_load_backend_kwarg(module_backend_impl.load):
+                if self._supports_arg(module_backend_impl.load, "load_backend"):
                     extra_kwargs["load_backend"] = load_backend
+                if self._supports_arg(module_backend_impl.load, "model_config"):
```
gkumbhat, line 182: question: The idea would be that the underlying module can accept `model_config` as an argument to `load` function to find the model mapping in remote cases?

gkumbhat: I think this would also mean that the caikit-module should not be having any sort of validation or configuration options which rely on reading the actual model's config file right?

gkumbhat: Wondering how does this correlate to using HF `config.json` as finder?

gabe-l-hart: Yeah, exactly. When we did the `finder`/`initializer` split, the interface actually got a bit awkward. Ideally, a `finder` returns a `ModuleConfig` and an `initializer` takes that `ModuleConfig` and is able to load directly. If it _requires_ on-disk assets (e.g. weights), there's no way around it, but ideally the `config.yml` would not need to be read from disk. The problem is that every `def load` function relies on being passed a `model_path` which is assumed to be the parent directory of the `config.yml`. Having `model_config` as an arg allows modules that are able to load without a `config.yml` to do so whereas now, the only way to get back that coordination information is to read `config.yml` or "just know it" (i.e. look for a `config.json` and parse it).

caikit/runtime/model_management/model_manager.py
gabe-l-hart: Yeah, it wraps all of its error handling with `CaikitRuntimeException`. If there are other unhandled error cases, they're bugs, so we would want them to surface anyway.

caikit/core/model_management/local_model_initializer.py
gkumbhat: Wondering if an alternate implementation of converting the `model_config` here loadable / readable automatically by `ModuleConfig` itself make it opaque to modules and require no change in like TGIS module etc?

gkumbhat (approved):
LGTM. Thanks Gabe!

PR #399, opened 2023-08-03
More training info

### LINKED ISSUE #396: Add start / end timestamps to TrainingInfo container

The `GetTrainingStatus` training management API has timestamps defined for the submission and completion time of a training job. However the model future's `get_info()` api does not return this info and nothing is hooked up in the servicer.

- Update the `TrainingInfo` message to add this info
- Hook up that data in the `GetTrainingStatus` api

## Additional context

We should probably be using real `datetime` objects here, and it'll be up to the trainer implementation to make sure that things like strings are parsed into correct timestamps

joerunde:

Closes #396

This does 3 things:
- Adds support for proto-ing timestamps in our data model
- Hooks up the submission and completion timestamps in the GetTrainingStatus api
- Fixes a bug in the local trainer where the completion time was unset

**If applicable**:

### DIFF
--- a/caikit/core/model_management/local_model_trainer.py
+++ b/caikit/core/model_management/local_model_trainer.py
@@ -18,7 +18,7 @@ The LocalModelTrainer uses a local thread to launch and manage each training job
 # Standard
 from concurrent.futures.thread import _threads_queues
 from datetime import datetime, timedelta
-from typing import Any, Dict, Iterable, Optional, Type, Union
+from typing import Any, Dict, Iterable, List, Optional, Type, Union
 import os
 import re
 import threading
@@ -78,25 +78,18 @@ class LocalModelTrainer(ModelTrainerBase):
         ):
             super().__init__(
                 trainer_name=trainer_name,
-                training_id=str(uuid.uuid4()),
+                training_id=external_training_id or str(uuid.uuid4()),
                 save_with_id=save_with_id,
                 save_path=save_path,
                 model_name=model_name,
+                use_reversible_hash=external_training_id is None,
             )
-            # 🌶️🌶️🌶️ For the external training id override, we don't want to include the
-            # reversible hash bit on the id. Therefore, we have to re-do the id and save
-            # path stuff done in the super() init here instead
-            if external_training_id is not None:
-                # No extra hash bit
-                self._id = external_training_id
-                self._save_path = self._save_path_with_id(
-                    save_path, save_with_id, external_training_id, model_name
-                )
-
             self._module_class = module_class
 
             # Placeholder for the time when the future completed
             self._completion_time = None
+            # Set the submission time as right now. (Maybe this should be supplied instead?)
+            self._submission_time = datetime.now()
 
             # Set up the worker and start it
             self._use_subprocess = use_subprocess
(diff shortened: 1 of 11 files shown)


tests/runtime/servicers/test_training_management_servicer.py
```diff
@@ -75,6 +79,46 @@ def test_training_runs(training_management_servicer, training_pool):
...
+        DataStream.from_iterable([]),
+        wait_event=event,
+    )
+
+    # send a request, check that the submission timestamp was set
+    request = TrainingInfoRequest(training_id=model_future.id).to_proto()
+    response = training_management_servicer.GetTrainingStatus(request, context=None)
+    response_dm: TrainingStatusResponse = TrainingStatusResponse.from_proto(response)
+    assert response_dm.submission_timestamp is not None
+    assert (
```
aluu317, line 98: what is the purpose of making sure that the time of this code and the time the training was submitted is less than 50 ms?

caikit/core/model_management/local_model_trainer.py
```diff
@@ -78,25 +78,18 @@ def __init__(
...
-                self._save_path = self._save_path_with_id(
-                    save_path, save_with_id, external_training_id, model_name
-                )
-
             self._module_class = module_class
 
             # Placeholder for the time when the future completed
             self._completion_time = None
+            # Set the submission time as right now. (Maybe this should be supplied instead?)
+            self._submission_time = datetime.now()
```
aluu317, line 92: Maybe this is not applicable for local model trainer, but I'm curious how the submission time will work once we have multiple training jobs queued up that may take a long time, would they all have same submission time even though the training potentially won't start until later?

tests/core/data_model/test_timestamp.py
```diff
@@ -0,0 +1,46 @@
+# Copyright The Caikit Authors
```
aluu317, line 1: Assuming we only use timestamp internally, set by trainers, we probably don't need to test `to_json`, `from_json` and `to_dict`

aluu317 (commented):
This is looking niceeeee!!! I just have a couple questions

tests/runtime/servicers/test_training_management_servicer.py
joerunde: If I had been going with a mock trainer I would set the timestamp explicitly and check for an exact result. But since I don't know what time exactly the local trainer started, I'm just checking that it was in the very recent past.

50ms could be pretty tight on some setups, but at least on my machine most tests take on the order of 1ms

caikit/core/model_management/local_model_trainer.py
joerunde: Right, so for other remote training systems that do their own queuing the trainer implementation would presumably ask that system what time the training was submitted at, and it'd be up to that system to provide a correct value

tests/core/data_model/test_timestamp.py
joerunde: good point though, I could create a few `@dataobjects` in these tests that contain timestamps to ensure the json/dict stuff actually works

aluu317 (approved):
LGTM

Issue #329, opened 2023-07-13
REST for Train


prashantgupta24:
## Is your feature request related to a problem? Please describe.

We have a `RuntimeGRPCServer` class that implements a grpc server that servers all of our `Train` runtime apis. We need the REST version of this.

## Describe the solution you'd like

The rest server (`FastAPIServer`?) class should be able to accept the same set of service packages as the grpc server, and expose REST endpoints for `Train` endpoints.

This should hook up to the `GlobalTrainServicer` to run model training.

PR #401, opened 2023-08-04
✨ add REST for train

prashantgupta24:

Adds REST support for Training APIs.

## Global train servicer
One big architectural challenge remains that the `global_train_servicer` is heavily proto-centric, specially the `run_training_job`(which accepts a `ProtoMessageType` in its function signature for now). Since `model_train` also utilizes the same function and that’s proto-centric as well, it’s not an easy function to refactor.

Currently in order to make things work with it, I added a bunch of ifs and elses within `global_train_servicer` to accommodate `request_params` passed through `kwargs` by the `http_server`. Joe recommended a better approach would be to actually create a `ProtoMessageType` within the `http_server` and send that to the `global_train_servicer` so that there’re minimal changes to it for now.

In the next refactor, we will eventually refactor `run_training_job` to actually accept a `request_data_model` type instead, which should make it type-agnostic and should work with both `http` and `grpc`

## Global predict servicer

Changing inference calls for interacting with `Global predict servicer` prove to be much harder just because the `Predict` function requires a `ProtoMessageType` along with a grpc context that has the `model_id` baked in.

For https://github.com/caikit/caikit/issues/329

Depends on https://github.com/caikit/caikit/pull/312 for internal code refactoring.

Good links: 
-  https://docs.pydantic.dev/latest/migration/#changes-to-pydanticbasemodel ?
- https://docs.pydantic.dev/dev-v2/api/main/#pydantic.main.BaseModel.model_dump_json
- https://github.com/tiangolo/fastapi/issues/3314

Todos:
- [x] Fix `jsondict` 
- [x] Incorporate Train API changes
- [x] Fix test hanging
- [x] Tmp dir for trainings
- [x] Check inference after training

**If applicable**:

### DIFF
--- a/caikit/runtime/http_server.py
+++ b/caikit/runtime/http_server.py
@@ -298,7 +298,9 @@ class RuntimeHTTPServer(RuntimeServerBase):
                 dm_kwargs[field_name] = self.build_dm_object(field_value)
             elif isinstance(field_value, list):
                 if all(type(val) in PYDANTIC_TO_DM_MAPPING for val in field_value):
-                    dm_kwargs[field_name] = [self.build_dm_object(val) for val in field_value]
+                    dm_kwargs[field_name] = [
+                        self.build_dm_object(val) for val in field_value
+                    ]
                 else:
                     dm_kwargs[field_name] = field_value
             else:
--- a/caikit/runtime/http_server.py
+++ b/caikit/runtime/http_server.py
@@ -305,6 +305,10 @@ class RuntimeHTTPServer(RuntimeServerBase):
                 substituted_json
             )
             request_params["training_data"] = json_data_obj
+        if output_path := request_params.get("output_path", None):
+            request_params["output_path"] = PYDANTIC_REGISTRY.get(
+                type(output_path)
+            ).from_json(output_path.model_dump_json())
         return request_params
 
     def _train_add_unary_input_unary_output_handler(self, rpc: CaikitRPCBase):
--- a/caikit/runtime/http_server.py
+++ b/caikit/runtime/http_server.py
@@ -59,6 +59,7 @@ from caikit.runtime.service_generation.rpcs import (
 from caikit.runtime.servicers.global_predict_servicer import GlobalPredictServicer
 from caikit.runtime.servicers.global_train_servicer import GlobalTrainServicer
 from caikit.runtime.types.caikit_runtime_exception import CaikitRuntimeException
+import caikit
 
 ## Globals #####################################################################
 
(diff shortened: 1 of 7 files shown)


caikit/runtime/http_server.py
```diff
@@ -272,10 +278,99 @@ def _get_request_params(
...
+
+            self.build_request_params_dict(request_params)
+            try:
+                call = partial(
+                    self.global_train_servicer.run_training_job,
+                    request=request,
+                    module=rpc.clz,
+                    training_output_dir=None,  # pass None so that GTS picks up the config one # TODO: double-check?
+                    request_params=request_params,
+                    # context=context,
```
prashantgupta24, line 349: Need to figure this out. The `context` we expect in `GTS` is different than the `context` sent here.

prashantgupta24:
Not really ready for review but worth getting an initial review in !

caikit/runtime/http_server.py
prashantgupta24: Should be fixed after we do #415 ?

caikit/runtime/http_server.py
```diff
@@ -272,10 +277,100 @@ def _get_request_params(
...
+        dm_kwargs = {}
+
+        for field_name, field_value in pydantic_model:
+            # field could be a DM:
+            # pylint: disable=unidiomatic-typecheck
+            if type(field_value) in PYDANTIC_TO_DM_MAPPING:
+                dm_kwargs[field_name] = self.build_dm_object(field_value)
+            elif isinstance(field_value, list) and len(field_value) > 0:
+                if all(type(val) in PYDANTIC_TO_DM_MAPPING for val in field_value):
+                    dm_kwargs[field_name] = [self.build_dm_object(field_value[0])]
```
joerunde, line 301: Won't this just give you a list of length 1?
Don't we want
```
dm_kwargs[field_name] = [self.build_dm_object(val) for val in field_value])
```

caikit/runtime/http_server.py
```diff
@@ -290,6 +385,7 @@ async def _handler(
             loop = asyncio.get_running_loop()
 
             request_params = self._get_request_params(rpc, request)
+            self.build_request_params_dict(request_params)
```
joerunde, line 390: nit: This function says it returns a dict, but it looks like it's also modifying the input dict in-place

Quick questions:
- Why is this here on the inference endpoints, is it a bugfix?
- I see this is only called right after `self._get_request_params`, should it just be wrapped into that?

prashantgupta24: Yeah could be combined I guess!

prashantgupta24: inference endpoints also need pydantic models to be converted to our DM objects right?

tests/fixtures/sample_lib/modules/other_task/other_implementation.py
```diff
@@ -54,7 +54,7 @@ def train(
         batch_size: int = 64,
     ) -> "OtherModule":
         """Sample training method that produces a trained model"""
-        assert type(sample_input) == SampleInputType or str
+        assert type(sample_input) == (SampleInputType or str)
```
joerunde, line 57: This was definitely wrong before, but I'm not sure this is correct either?
```
>>> str == (int or str)
False
```

prashantgupta24: lol yeah 🤦

prashantgupta24: I'm going to go with
```
isinstance(sample_input, SampleInputType) or isinstance(sample_input, str)
```

caikit/runtime/http_server.py
joerunde: I suppose that actually wasn't happening before and the equivalent pydantic objects were just being passed along?
In that case duck-typing would be fine, but any code that a module author wrote that type-checked against the data model type would fail. Is that why there's now a `SampleInputType` type check in one of the test modules?

prashantgupta24: Exactly.

caikit/runtime/http_server.py
joerunde: Yup, there's some mad refactoring opportunity here.

It's interesting that this http server is actually following the pattern we had wanted for the grpc server but didn't get around to- where individual handlers are built for each `rpc`. This could all be so congruent and beautiful... if we ever have the time to do it 😉

caikit/runtime/http_server.py
joerunde: Sweet, solid upgrade 👍

tests/fixtures/sample_lib/modules/other_task/other_implementation.py
joerunde: We do have the type checker if you wanted to use that, but I don't really mind either way

joerunde (approved):
lpgtm! (I may have lost steam in the actual unit tests)

This is pretty slick now that the changes are contained to the http server

prashantgupta24:
Thanks @joerunde ! The unit tests are similar to the ones we have in the `grpc` testing for training and inferencing on the trained model.

PR #405, opened 2023-08-07
Retry load failures 404

### LINKED ISSUE #404: Retry failed local model loads
## Describe the bug

With the periodic load OR with an explicitly requested load based on an inference call, we can easily hit the case where the artifacts themselves have not yet finished writing to the shared disk. When this happens, a thread is spawned to load the model which will eventually raise for one reason or another (corrupted file, missing file, etc...). These exceptions won't tank the server, and the model future will be removed from the loaded_models dict, but any inference calls that are in flight will get 500s. Furthermore, this can happen really easily in tests!

gabe-l-hart:

Closes #404 

This PR adds retry support for loading models from `local_models_dir`. This is needed to handle races with model artifact copying since an artifact will not be atomically copied to the mounted directory. If the periodic sync or a user inference request is triggered when a model is partially uploaded, the `load` function will begin, but will fail due to corrupted resources. Prior to this PR, this situation was handled by simply removing the model from the set of loaded models and letting the lazy loading restart on the next stimulus (period or inference request). The problem with this approach is that any requests received during the beginning of the load and the eventual failure/unload would receive 500 responses due to the failed load. With this PR, those requests will continue to wait until either the load succeeds or all retries are exhausted.

### DIFF
--- a/caikit/runtime/model_management/loaded_model.py
+++ b/caikit/runtime/model_management/loaded_model.py
@@ -40,8 +40,10 @@ try:
 except TypeError:  # pragma: no cover
     CaikitModelFuture = Future
 
+CaikitModelFutureFactory = Callable[[], CaikitModelFuture]
 
-class LoadedModel:
+
+class LoadedModel:  # pylint: disable=too-many-instance-attributes
     __doc__ = __doc__
 
     class Builder:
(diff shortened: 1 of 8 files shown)


caikit/runtime/model_management/loaded_model.py
```diff
@@ -110,9 +123,27 @@ def wait(self):
...
+                        self._retries,
+                    )
+                    self._caikit_model_future = self._caikit_model_future_factory()
+                    # Try waiting again with a fresh load future. This may open
+                    # a recursive retry.
+                    try:
+                        self.wait()
+                    except CaikitRuntimeException:
+                        if self._fail_callback:
+                            self._fail_callback()
```
joerunde, line 141: won't this potentially call the fail callback `n` times? One for each retry failure in the case of many retry failures

gabe-l-hart: That seems... quite likely 🤦. Covid coding at its finest!

joerunde (approved):
looks slick!

Issue #386, opened 2023-08-01
Cancellation can result in NOT_FOUND for model deletions


evaline-ju:
## Platform

Please provide details about the environment you are using, including the following:

- Interpreter version:
- Library version:

## Sample Code

Please include a minimal sample of the code that will (if possible) reproduce the bug in isolation

## Observed behavior

What you see happening (error messages, stack traces, etc...)

PR #406, opened 2023-08-07
Handle trainers that cancel-by-delete

joerunde:

Wraps up #386 

This returns `CANCELLED` successfully when the underlying model future raises a not found error after cancellation

**If applicable**:

### DIFF
--- a/caikit/runtime/servicers/training_management_servicer.py
+++ b/caikit/runtime/servicers/training_management_servicer.py
@@ -23,7 +23,11 @@ import alog
 
 # Local
 from caikit.core import MODEL_MANAGER
-from caikit.core.exceptions.caikit_core_exception import CaikitCoreException
+from caikit.core.data_model import TrainingStatus
+from caikit.core.exceptions.caikit_core_exception import (
+    CaikitCoreException,
+    CaikitCoreStatusCode,
+)
 from caikit.interfaces.runtime.data_model import (
     TrainingInfoRequest,
     TrainingStatusResponse,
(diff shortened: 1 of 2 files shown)


aluu317 (approved):
LGTM

caikit/runtime/servicers/training_management_servicer.py
```diff
@@ -69,34 +73,44 @@ def GetTrainingStatus(self, request, context):  # pylint: disable=unused-argumen
...
 
             return TrainingStatusResponse(
-                training_id=training_info.training_id,
-                state=model_future.get_info().status,
+                training_id=model_future.id,
+                state=training_info.status,
                 reasons=reasons,
             ).to_proto()
         except CaikitCoreException as err:
+            # In the case that we get a `NOT_FOUND`, we assume that the training was canceled.
```
gabe-l-hart, line 94: Does this mean that calling `CancelTraining` with a random sequence of characters will show up as `CANCELED` and not `NOT_FOUND`? That seems maybe worse than returning `NOT_FOUND` after cancellation? Hm, that's a tricky one.

gabe-l-hart (commented):
One question, but not a show stopper

caikit/runtime/servicers/training_management_servicer.py
joerunde: It doesn't, since the initial call to get the model future would fail. This try only covers the `future.cancel()` and subsequent `future.get_info()`

PR #408, opened 2023-08-08
🐛 fix start_runtime_with_sample_lib script

prashantgupta24:

This PR fixes the `start_runtime_with_sample_lib` script that starts a server and serves the `sample_lib` within `Caikit` for users to play around with the APIs.

**If applicable**:


examples/sample_lib/start_runtime_with_sample_lib.py
```diff
@@ -0,0 +1,59 @@
...
+                        {"number": 2},
+                    ]
+                )
+            )
+
+        alog.configure(default_level="debug")
+        main()
+
+        # remove generated protos
+        shutil.rmtree("protos", ignore_errors=True)
```
gabe-l-hart, line 59: NIT: Since you need to `ctrl-C` to kill a server, I suspect that this will not actually get run

gabe-l-hart (approved):
Looks good, thanks Prashant!

examples/sample_lib/start_runtime_with_sample_lib.py
joerunde: To add a hint here: https://stackoverflow.com/questions/1112343/how-do-i-capture-sigint-in-python

Should be pretty easy to register an interrupt handler to do this instead

joerunde:
oh lol, we merged it anyway 🤷

Issue #412, opened 2023-08-09
Support multiple model management components in runtime


gabe-l-hart:
## Is your feature request related to a problem? Please describe.

Currently, `caikit.core.model_management` provides support for a key/value map of model management components (`trainer`, `finder`, and `initializer`), and those keys can be used in `caikit.train`/`caikit.load` to indicate instances to use beyond `"default"`. The services in `caikit.runtime` don't allow these arguments to be set, so `"default"` is always used. This requires that every component implementation needs to support all possible operations (one `trainer` for all trainings, one `finer` for all flavors of model, etc...).

## Describe the solution you'd like

We need the ability to configure a set of component implementations that can work together without requiring the user to explicitly specify the desired implementation label.

## Describe alternatives you've considered

There are a number of different ways we could implement this:

* Allow these kwargs through the service requests (possibly as headers or similar)
    * Advantages:
        * Pure runtime solution
        * Mirrors the "local python" experience in runtime
    * Problems:
        * There is currently no API for `load`
        * This _could_ be a security vector to expose config if the upstream application is not sufficiently careful
        * Lots of potential error cases that need to be handled at runtime

* Create implementations of each of the component abstractions that implement "multi-instance" semantics
    * Advantages:
        * Explicitly configurable at deploy time
        * No deploy-side config leaked into the API
    * Problems:
        * Configuration changes only handled at deploy time

## Additional context

Add any other context about the feature request here.

joerunde:
@gabe-l-hart to split

PR #413, opened 2023-08-09
Multi mm component 412

gabe-l-hart:

Supports #412 

This PR adds a `MultiModelFinder` that is capable of proxying to an ordered list of multiple finders in the `model_management.finders` section.

This will be needed in order to fully take advantage of the `TGISAutoFinder` in `caikit-nlp` ([PR](https://github.com/caikit/caikit-nlp/pull/123))

### DIFF
--- a/caikit/core/model_management/multi_model_finder.py
+++ b/caikit/core/model_management/multi_model_finder.py
@@ -1,0 +1,137 @@
+# Copyright The Caikit Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+The MultiModelFinder configures a set of other model finders that will be used
+in sequence to try loading models.
+
+Configuration for MultiModelFinder lives under the config as follows:
+
+model_management:
+    finders:
+        <finder name>:
+            type: MULTI
+            config:
+                # Sequence of other finder names to use in priority order
+                finder_sequence:
+                    - other_finder1
+                    - other_finder2
+"""
+# Standard
+from typing import Optional
+
+# First Party
+import aconfig
+import alog
+
+# Local
+from ...config import get_config
+from ..exceptions import error_handler
+from ..modules import ModuleConfig
+from .model_finder_base import ModelFinderBase
+
+# NOTE: Top-level import done so that global MODEL_MANAGER can be used at
+#   construction time without incurring a circular dependency
+import caikit.core
+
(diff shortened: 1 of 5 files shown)


caikit/core/model_management/multi_model_finder.py
```diff
@@ -0,0 +1,137 @@
...
+            invalid_finders,
+        )
+        error.value_check(
+            "<COR54613971E>",
+            self._instance_name not in finder_priority,
+            "Cannot include self in multi finder priority",
+        )
+        model_manager = config.model_manager or caikit.core.MODEL_MANAGER
+        log.debug2("Setting up %s with finder priority: %s", self.name, finder_priority)
+        self._finders = [model_manager.get_finder(finder) for finder in finder_priority]
```
joerunde, line 92: 🤔  but won't this fail if you try to look up a finder that has not yet been initialized?

caikit/core/model_management/multi_model_finder.py
```diff
@@ -0,0 +1,137 @@
...
+            not invalid_finders,
+            "Invalid finders given in finder_priority: {}",
+            invalid_finders,
+        )
+        error.value_check(
+            "<COR54613971E>",
+            self._instance_name not in finder_priority,
+            "Cannot include self in multi finder priority",
+        )
+        model_manager = config.model_manager or caikit.core.MODEL_MANAGER
```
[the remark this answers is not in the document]
gabe-l-hart: You got me! This is a workaround for the fact that we have global state in `MODEL_MANAGER`. This is not something that a user could actually do through env vars or `caikit.configure`. What _can_ be done is manually instantiating a `ModelFinder`. The two scenarios where this would happen would be (a) unit tests (the real reason I did this) or (b) a custom piece of code for an application that wants a non-global model manager.

caikit/core/model_management/multi_model_finder.py
gabe-l-hart: `model_manager.get_finder` does the initialization lazily

caikit/core/model_management/multi_model_finder.py
```diff
@@ -0,0 +1,137 @@
...
+
+# Local
+from ...config import get_config
+from ..exceptions import error_handler
+from ..modules import ModuleConfig
+from .model_finder_base import ModelFinderBase
+
+# NOTE: Top-level import done so that global MODEL_MANAGER can be used at
+#   construction time without incurring a circular dependency
+import caikit.core
```
joerunde, line 45: Is our goal to eventually kill the global singletons like the `caikit.core.MODEL_MANAGER`? And instead do some sort of dependency injection where it's needed?

gabe-l-hart: yes! As I've been working around these things in all of our tests, I'm starting to imagine a future where there is a `CaikitContext` that holds the "globals" as instance variables. We'd then have a special `contextmanater` that would allow a fresh context to replace the default one.

joerunde: or we pick something from here instead of rolling our own? https://wasinski.dev/comparison-of-dependency-injection-libraries-in-python/

joerunde: Yeah, I would like to rent a big cabin in the woods where we can all go crash on this for a couple days and just get it done

PR #416, opened 2023-08-11
🐛 Fix derived DM to pydantic conversion bug

### LINKED ISSUE #419: Conversion of dataobjects-with-inheritance to pydantic models is broken
## Describe the bug

```
    @dataobject
    class Base(DataObjectBase):
        foo: int
        bar: int

    @dataobject
    class Derived(Base):
        bar: str
        baz: str
```

If we create a `pydantic model` using our `._dataobject_to_pydantic(Derived)` function call, it omits `foo`

## Platform

- Interpreter version: 3.9
- Library version: 0.16.x

prashantgupta24:

Fixes an issue when converting dataobject to pydantic models when there is inheritance involved.

fix #419 

> FYI - this has a consequence in all cases where we have derived DM objects. Instantiating any derived DM object will require that we satisfy ALL fields within the base DM object as well. (applies to `S3Path`)

#### Useful links:
- https://docs.python.org/3/library/typing.html#typing.get_type_hints
- https://peps.python.org/pep-0526/#runtime-effects-of-type-annotations
    
     > The recommended way of getting annotations at runtime is by using typing.get_type_hints function; as with all dunder attributes, any undocumented use of __annotations__ is subject to breakage without warning:

**If applicable**:

### DIFF
--- a/tests/runtime/test_http_server.py
+++ b/tests/runtime/test_http_server.py
@@ -350,7 +350,7 @@ def test_health_check_ok(runtime_http_server):
         assert response.text == "OK"
 
 
-def test_pydantic_wrapping_with_enums():
+def test_pydantic_wrapping_with_enums(runtime_http_server):
     """Check that the pydantic wrapping works on our data models when they have enums"""
     # The NLP GeneratedTextStreamResult data model contains enums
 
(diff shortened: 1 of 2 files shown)


tests/runtime/test_http_server.py
```diff
@@ -350,7 +350,7 @@ def test_health_check_ok(runtime_http_server):
         assert response.text == "OK"
 
 
-def test_pydantic_wrapping_with_enums():
+def test_pydantic_wrapping_with_enums(runtime_http_server):
```
gabe-l-hart, line 353: I'm curious why this changed since `_dataobject_to_pydantic` is a `@classmethod` and doesn't need an instance of the server. This seems like it's adding extra work by booting a real server, although I believe `runtime_http_server` is session scoped, so it's probably not _actually_ extra work.

gabe-l-hart (approved):
Very nice! One question about a test change, but otherwise this is good to go

tests/runtime/test_http_server.py
prashantgupta24: Yeah I guess that makes sense. I think it was added only for convenience so that we don't have to create an instance, but unnecessary overhead for sure

PR #425, opened 2023-08-21
Module load api cleanup 417

### LINKED ISSUE #417: Module.load contract is kinda odd

We used to pass `model_path: str` to all module's `.load` functions.

This needed to change to support the ability to load modules in a situation where config exists in memory, but not on disk.
However, module implementations out there in the wild make some assumptions that `module_path` is in fact a path to a spot on disk.

Relevant commit where `module_path: str` was changed to `module_path: str | ModuleConfig`: https://github.com/caikit/caikit/pull/397/commits/e01ec976a9fbe48d34b44933ba2bf4f82d4369cb

and a discussion between @gabe-l-hart and @gkumbhat on the PR: https://github.com/caikit/caikit/pull/397#discussion_r1283695548

and an example of a place this blows up now: https://github.ibm.com/ai-foundation/watson_doc_understanding/blob/6f65145b48f8918ad6eba05c934d8ddc5f4918ca/watson_doc_understanding/blocks/ocr/iocr.py#L153

The reason that this is immediately a problem is that the `LocalModelIniitializer` is now only passing in `ModuleConfig`s, never paths. The workaround for users is to not assume that `module_path` is in fact a path on disk, and instead use a ModuleLoader, like:
```
def load(model_path):
    loader = ModuleLoader(model_path)
    model_path = loader.load_arg("model_path")
```
This is, according to @sboagibm 
> obscure, but ok

This presents some more problems, since the differences between loading from disk vs. an in-memory config aren't well encapsulated behind the `ModuleLoader`. If load implementations simply pull things from the module config that they expect to be paths on disk, the load method will still fail because nothing actually exists on disk. This is probably acceptable for now when we only expect certain specific modules to need to load ephemerally, but could cause some problems in the future.

At the end of the day, we should make it clear what users are supposed to expect as the argument to `load`, and if everything is supposed to go through the `ModuleLoader`, maybe it should be just that?

??????

- Refactor the module's `save` and `load` methods to take `ModuleSaver`s and `ModuleLoader`s
- Hide the FS behind the `ModuleLoader`, e.g. `loader.load_file("some_artifact_name")` can fetch content from disk or return it from memory

## Additional context

gabe-l-hart:

Closes #417 

This PR fixes the overly-strict assumptions made about how `load` functions have been implemented in the wild, thus fixing backwards compatibility for recent changes to pass `ModuleConfig` as the first argument to `load`. It does so explicitly to fix backwards compatibility, so did not attempt to make anything more strict w.r.t. setting expectations about how `load` should be implemented in the future

**If applicable**:

### DIFF
--- a/caikit/core/model_manager.py
+++ b/caikit/core/model_manager.py
@@ -203,7 +203,7 @@ class ModelManager:
 
     def load(
         self,
-        module_path: str,
+        module_path: Union[str, BytesIO, bytes],
         *,
         load_singleton: bool = False,
         finder: Union[str, ModelFinderBase] = "default",
--- a/caikit/core/model_manager.py
+++ b/caikit/core/model_manager.py
@@ -270,7 +270,7 @@ class ModelManager:
                 module_path, load_singleton, finder, initializer, **kwargs
             )
         try:
-            return self._load_from_dir(
+            return self._do_load(
                 module_path, load_singleton, finder, initializer, **kwargs
             )
         except FileNotFoundError:
@@ -422,9 +422,7 @@ class ModelManager:
 
     ## Implementation Details ##################################################
 
-    def _load_from_dir(
-        self, module_path, load_singleton, finder, initializer, **kwargs
-    ):
+    def _do_load(self, module_path, load_singleton, finder, initializer, **kwargs):
         """Load a model from a directory.
 
         Args:
@@ -519,7 +517,7 @@ class ModelManager:
             # to files directly, or it may unpack to a (single) directory containing the files.
             # We expect the former, but fall back to the second if we can't find the config.
             try:
-                model = self._load_from_dir(
+                model = self._do_load(
                     extract_path, load_singleton, finder, initializer, **kwargs
                 )
             # NOTE: Error handling is a little gross here, the main reason being that we
@@ -549,7 +547,7 @@ class ModelManager:
                 # Otherwise, try again. If we fail again stop, because the zip creation should only
                 # create one potential extra layer of nesting around the model directory.
                 try:
-                    model = self._load_from_dir(
+                    model = self._do_load(
                         nested_dirs[0], load_singleton, finder, initializer, **kwargs
                     )
                 except FileNotFoundError:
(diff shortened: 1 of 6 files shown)


caikit/core/model_manager.py
```diff
@@ -251,7 +251,9 @@ def load(
         load_path = get_config().load_path
         if load_path is not None and isinstance(module_path, str):
             if not os.path.exists(module_path):
-                module_path = os.path.join(load_path, module_path)
+                full_module_path = os.path.join(load_path, module_path)
+                if os.path.exists(full_module_path):
+                    module_path = full_module_path
```
joerunde, line 256: nit: The `self._load_from_dir` after this has a misleading name now since it invokes finders + initializers that may not require that a path to a directory on disk

tests/core/test_model_manager.py
```diff
@@ -616,6 +618,60 @@ def test_find_without_on_disk_model(good_model_path, reset_globals):
...
+    with tempfile.TemporaryDirectory() as workdir:
+        with temp_config(
+            {
+                "load_path": workdir,
+                "model_management": {
+                    "finders": {
+                        "default": {
+                            "type": TestFinder.name,
+                            "config": {
+                                "fallback_to_local": False,
```
gkumbhat, line 634: `fallback_to_local` does't seem to get activated since https://github.com/caikit/caikit/pull/425/files#diff-b6b4baaaf129f8633d11eaae2a9b6ae490f04e65276cc1d541cafe822422e31fR90 will be false?

joerunde, line 634: Why is this `fallback_to_local` config required?

It looks like that is used only if the model path actually exists, but for this test the path should not exist

```
        if os.path.exists(model_path):
            if self._fallback_to_local:
...
```

gabe-l-hart: Yep! Totally missed that. Good catch

caikit/core/model_manager.py
```diff
@@ -420,9 +422,7 @@ def get_initializer(
 
     ## Implementation Details ##################################################
 
-    def _load_from_dir(
-        self, module_path, load_singleton, finder, initializer, **kwargs
-    ):
+    def _do_load(self, module_path, load_singleton, finder, initializer, **kwargs):
```
joerunde, line 425: love me a `_do` method

joerunde (approved):
nice!

PR #430, opened 2023-08-22
LazyLoadEphemeralModels

### LINKED ISSUE #429: Allow ephemeral models to lazy load in runtime

The current implementation of lazy loading in `caikit.runtime.model_management.model_manager` is centered around `local_models_dir`. This makes sense for physical models, but with the presence of the `ModelFinder` abstraction, we also want to support "auto-finders" that are able to infer a model's configuration without needing the model on disk physically.

An "auto-finder" may be configured to receive the `model_path` and find a model based on it in any arbitrary way. Given this, an auto-finder is free to play tricks with the on-disk path such as `os.path.basename` to strip off path prefixes, but this also interferes with the lazy-load sync's unloading logic which assumes that a model whose path begins with `local_models_dir` which is not present on disk has been deleted and therefore should be unloaded.

To solve this, we want the runtime `ModelManager` to explicitly permit lazy loading of non-disk models by using the `model_id` as the `model_path` for models where `os.path.join(local_models_dir, model_id)` does not exist.

We could solve this by requiring auto-finders to be smart enough to handle this themselves, but this places several limitations on what finders can do and decentralizes the logic for this kind of ephemeral model.

## Additional context

This is currently limiting the use of `TGISAutoFinder` ([here](https://github.com/caikit/caikit-nlp/blob/main/caikit_nlp/model_management/tgis_auto_finder.py)) with `runtime`

gabe-l-hart:

Closes #429 

**If applicable**:

### DIFF
--- a/caikit/runtime/model_management/model_manager.py
+++ b/caikit/runtime/model_management/model_manager.py
@@ -381,6 +381,12 @@ class ModelManager:  # pylint: disable=too-many-instance-attributes
             log.debug2(
                 "Lazy loading local model %s from %s", model_id, local_model_path
             )
+            # If the model is not present on disk, attempt to lazy load it
+            # anyway. This allows auto-finders that can infer the model's config
+            # to load based on the ID
+            if not os.path.exists(local_model_path):
+                log.debug2("Attempting to load ephemeral model %s", model_id)
+                local_model_path = model_id
             self.load_model(
                 model_id=model_id,
                 local_model_path=local_model_path,
@@ -392,7 +398,7 @@ class ModelManager:  # pylint: disable=too-many-instance-attributes
 
         # If still not loaded, there's nothing to find, so raise NOT_FOUND
         if not model_loaded:
-            msg = "Model '%s' not loaded" % model_id
+            msg = f"Model '{model_id}' not loaded"
             log.debug(
                 {"log_code": "<RUN61105243D>", "message": msg, "model_id": model_id}
             )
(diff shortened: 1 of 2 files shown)


caikit/runtime/model_management/model_manager.py
```diff
@@ -381,6 +381,12 @@ def retrieve_model(self, model_id: str) -> ModuleBase:
             log.debug2(
                 "Lazy loading local model %s from %s", model_id, local_model_path
             )
+            # If the model is not present on disk, attempt to lazy load it
+            # anyway. This allows auto-finders that can infer the model's config
+            # to load based on the ID
+            if not os.path.exists(local_model_path):
```
gkumbhat, line 387: super nit: For clarity perspective, should we add in the above comment that in case the `local_model_path` doesn't exist, we will be using `model_id` as a backup to let underlying module impl make use of that as required.

gkumbhat: Also wondering if making using `model_id` as `local_model_path` can open up some security concern?

gabe-l-hart: good call on the comment

gabe-l-hart: >Also wondering if making using model_id as local_model_path can open up some security concern?

Hmm, that's interesting. I suppose it's _possible_ that an attacker could probe the filesystem layout by putting absolute paths in as `model_id` and looking at what kind of errors could come. I don't _think_ our errors would be informative enough on the user-side to be able to tell the difference between "that doesn't exist on disk" and "that's not a valid path to a model but it does exist." Even with that information, I don't think there's any way this could result in code execution without control over the `finder/initializer` code, so I don't think this is a big concern. It's also not at all new with this PR since `os.path.join("/foo", "/bar/baz")` results in just `"/bar/baz"` since it's absolute already.

gkumbhat (approved):
LGTM. Thanks Gabe

PR #432, opened 2023-08-23
fix: change in function args order to _extract_pred_set

Ssukriti:
The proposed change is **original watson_core order** of function args and also what is specified in docstrings. changing order in caikit can cause weird behavior in downstreams unless we all change order of arguments, which is a huge change for nLP and unnecessary
 
blocking NLP rebase. There were no unit tests for these functions from the start, will be a larger effort to add for all these functions in module base. 

**If applicable**:

### DIFF
--- a/caikit/core/modules/base.py
+++ b/caikit/core/modules/base.py
@@ -677,7 +677,7 @@ class ModuleBase(metaclass=_ModuleBaseMeta):
             NotImplementedError("This is not available in this module."),
         )
 
-    def _extract_pred_set(self, dataset, *args, preprocess_func=None, **kwargs):
+    def _extract_pred_set(self, dataset, preprocess_func=None, *args, **kwargs):
         """Method for extracting pred set from dataset. Implemented in subclass.
 
         Args:
--- a/caikit/core/modules/base.py
+++ b/caikit/core/modules/base.py
@@ -677,7 +677,7 @@ class ModuleBase(metaclass=_ModuleBaseMeta):
             NotImplementedError("This is not available in this module."),
         )
 
-    def _extract_pred_set(self, dataset, preprocess_func=None, *args, **kwargs):
+    def _extract_pred_set(self, dataset, preprocess_func=None, **kwargs):
         """Method for extracting pred set from dataset. Implemented in subclass.
 
         Args:
@@ -687,7 +687,7 @@ class ModuleBase(metaclass=_ModuleBaseMeta):
                 steps that need to be taken to run the model on the input text.
                 This helper function ultimately leads to the input to this
                 module and may involve executing other modules.
-            *args, **kwargs (dict): Optional keyword arguments for prediction set extraction.
+            **kwargs (dict): Optional keyword arguments for prediction set extraction.
         Returns:
             list: List of labels in the format of the module_type that is being
                 called.


gabe-l-hart (approved):
Wild! I have no idea how this got changed. My only thought is that by putting `preprocess_func` after `*args` it was intended to require it be passed as a keyword arg.

gabe-l-hart (changes requested):
Ah ha! https://github.com/caikit/caikit/actions/runs/5953707459/job/16155052455?pr=432#step:6:24

This was changed because `pylint` doesn't like it. This brings up another question, though: Why is this exact function ever being called? I would always expect this to be overridden by a child if it's going to be used since this one simply raises `NotImplementedError`.

Ssukriti:
It is defined by derived class, but from the team's testing on caikit branch they are seeing an error -> `we need to specify "preprocess_func=preprocess_func", otherwise it will throw out the error that multiple preprocess_func is assigned .`

for example with function overriden heer https://github.ibm.com/ai-foundation/watson_nlp/blob/caikit_rebase/watson_nlp/blocks/classification/base.py#L72 

the corresponding notebook https://github.ibm.com/ai-foundation/watson_nlp_demos/blob/main/demos/watson_nlp_lib/Modules/classification/svm_classification_demo.ipynb , with caikit migration now gives error that it found multiple definitions of preprocess

let me think about what to do here then if pylint is complaining

gabe-l-hart:
Makes sense. I'm open to anything on this (including `# pylint: disable=...`)

joerunde:
> Makes sense. I'm open to anything on this (including # pylint: disable=...)

👍 on doing anything simple now to unblock work

> There were no unit tests for these functions from the start, will be a larger effort to add for all these functions in module base.

Right, all of these evaluation-related methods grew organically out of our internal product needs and then happened to land here. Now we have a big portion of our `caikit.module` code that's a big untested black box- the only user is a set of proprietary tools within IBM, and no documentation exists here for module developers or data scientists to know how to use it. So while these evaluation methods are great, in order to be a useful thing in caikit they need to not be a black box, which means I think a minimum of 1. open-sourcing the tools we use to run these quality evaluations, 2. committing to writing good user guides for using those tools + overriding these methods, and 3. covering this all with testing.

I think adding unit tests would be great, but I wouldn't want y'all to spend a bunch of time doing that here if we're not going to commit to supporting model evaluation in caikit. WDYT @Ssukriti @gabe-l-hart ?

Ssukriti:
I think we need to have a longer discussion about supporting quality evaluation in particular, as we (internal IBMers) have found it complicated to adopt in internal libraries (also because of lack of documentation anywhere about it :) )- I think watson NLP is only one using this from old times. If we do want to continue to support it, I agree we have lot of documenting to do. Can we open a backlog issue for it ? 
Let us not block this PR on better support for evaluation though, as this PR is needed for ongoing efforts

gabe-l-hart (approved):

gabe-l-hart:
Good discussion @joerunde @Ssukriti. Definitely agree that in the long run we need to either treat evaluation as a first-class part of the library or remove it back to an internal IBM "add on layer." For now, I'll merge with this fix.

Ssukriti:
thank you! I know you all recently made a caikit release, but may I please request a release next week as well for the few fixes that went in?

PR #434, opened 2023-08-24
♻️  split training parameters

joerunde:

POC for #369 

This updates the training RPC to manage an inner and outer request message, and it seems to work pretty slick.
cc @gabe-l-hart, @gkumbhat and @prashantgupta24 I think this'll work if we wanted to rip the bandaid on doing this nesting so we can have the pydantic apis convert these data models directly on the http side as discussed.

Open problem now is getting the caikit user access to the inner data model type so they can construct one of these messages directly. But running via compiled protobuf or REST that won't be an issue.

> Update - the `client_module` has been updated to incorporate ALL DM objects, so that unit tests as well as users can fetch the inner Proto message from it.

## Additional context

Pending pydantic investigation: https://github.com/pydantic/pydantic/discussions/7239 - thinking what to do about `forward-compatibility` - right now adding `extra_forbid` on all `pydantic models` has the unindented consequence that we won't be able to pass in anything that's not in the pydantic model, which means if the clients update before the server does, they won't work.

One quick workaround was to only attach `extra_forbid` to DMs that are a `Union`, but then if we first come across a data model used in a non-union context, and then later it is used within a union, then we would only create the pydantic data model without the `extra_forbid` bit, which is incorrect.

Tracking forward-compatibility issue here: https://github.com/caikit/caikit/issues/450

**If applicable**:
- [] this PR has been tested for backwards compatibility

### DIFF
--- a/caikit/runtime/http_server.py
+++ b/caikit/runtime/http_server.py
@@ -602,7 +602,9 @@ class RuntimeHTTPServer(RuntimeServerBase):
 
     @classmethod
     # pylint: disable=too-many-return-statements
-    def _get_pydantic_type(cls, field_type: type) -> type:
+    def _get_pydantic_type(
+        cls, field_type: type, is_extra_forbid: bool = False
+    ) -> type:
         """Recursive helper to get a valid pydantic type for every field type"""
         # pylint: disable=too-many-return-statements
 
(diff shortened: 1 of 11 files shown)


tests/runtime/test_http_server.py
```diff
@@ -475,7 +475,7 @@ class FooTest(DataObjectBase):
     foo = FooTest(bars=[BarTest(1)])
     assert foo.bars[0].baz == 1
 
-
+@pytest.mark.skip()
```
prashantgupta24, line 478: remove this

caikit/runtime/http_server.py
```diff
@@ -311,35 +311,13 @@ def _get_request_params(
         # convert pydantic objects to our DM objects
         for param_name, param_value in request_params.items():
             if issubclass(type(param_value), pydantic.BaseModel):
-                request_params[param_name] = self.build_dm_object(param_value)
+                request_params[param_name] = self._build_dm_object(param_value)
         return request_params
 
-    def build_dm_object(self, pydantic_model: pydantic.BaseModel) -> DataBase:
```
prashantgupta24, line 317: No code change, just converted to `classmethod` and renamed it to `_build_dm_object`

joerunde: Nice!

It's looking like it's hight time for some restructuring here- this one file has a whole lotta stuff about `pydantic` and `http handlers` intermixed.

Thoughts on extracting out the pydantic logic to a `caikit/core/data_model/pydantic_view.py` or similar spot? And slapping a few unit tests in there for:
- pydantic model creation from DM class
- pydantic instance -> dm instance translation
- dm instance -> pydantic instance translation

caikit/runtime/service_factory.py
```diff
@@ -141,18 +142,17 @@ def get_service_package(cls, service_type: ServiceType) -> ServicePackage:
...
             "ClientMessages",
             "Package with service message class implementations",
         )
 
-        for dm_class in request_data_models:
-            # We need the message class that data model serializes to
-            setattr(client_module, dm_class.__name__, type(dm_class().to_proto()))
+        for proto_class in _AUTO_GEN_PROTO_CLASSES:
+            # We need all the DM objects in the client_module for ease of use
+            setattr(client_module, proto_class.DESCRIPTOR.name, proto_class)
```
joerunde, line 155: After hacking a bit with @sboagibm today, I think we'll end up wanting something even more robust here.

The `messages` module can only be used if you happen to know the name of the thing you're requesting, which isn't super useful in many contexts. We wanted to write a generic initializer that would spin up a shim module to another caikit runtime, but then we had to write things like:
```
request_class = getattr(service_package.messages, f"{task_class}Request")
```

Food for thought: what if we had something like this here? And returned full data models instead of just the proto messages?
```
def get_inference_request(module_class: Type[ModuleBase]) -> Type[DataBase]:

def get_training_request(module_class: Type[ModuleBase]) -> Type[DataBase]:

def get_training_parameters(module_class: Type[ModuleBase]) -> Type[DataBase]:
```

joerunde: (food for _future_ thought, to be clear)

tests/runtime/test_http_server.py
```diff
@@ -240,6 +251,235 @@ def test_services_disabled(open_port, enabled_services):
             # )
 
 
+## Functional Tests #######################################################################
+
+
+def test_build_dm_object_simple(runtime_http_server):
```
joerunde, line 257: love all the new tests! (and definitely want them separated form the http server ones later 😄 )

joerunde (commented):
Hah, I can't approve my own PR lol

Thanks for slogging through all the test updates with the API change, really appreciate it @prashantgupta24 !

prashantgupta24 (approved):

PR #438, opened 2023-08-24
Don't overwrite the __repr__ with dataclass and fix HTTP output format

### LINKED ISSUE #437: DataObjectBase's __repr__ does not work properly
## Describe the bug

When we build new class which is inherited from DataObjectBase, the new class could not inherit the correct DataObjectBase's __repr__ function.

## Platform

- Interpreter version:
- Library version:

## Sample Code

@dataobject(package="caikit_data_model.nlp")
class AggregatedSentiment(DataObjectBase):

## Expected behavior

The output of the new class should be the same as when we use watson_core, like
image

## Observed behavior
image

gabe-l-hart:

Closes #437 

This PR also stumbled on a fuuuun bug: We were never checking the output format of server-side streaming with SSE. Turns out it was just calling `__repr__` on each yielded data object which was returning as `"MyDataObject(foo=1)"` rather than the `json` version `{"foo": 1}`. This made the HTTP output stream virtually useless! Updating the base `__repr__` fixes this, so this PR also fixes the test for server side checking to make sure the output is `json` parsable.

### DIFF
--- a/caikit/core/data_model/dataobject.py
+++ b/caikit/core/data_model/dataobject.py
@@ -200,7 +200,7 @@ def dataobject(*args, **kwargs) -> Callable[[_DataObjectBaseT], _DataObjectBaseT
             if _has_dataclass_init(cls):
                 log.debug3("Resetting default dataclass init")
                 delattr(cls, "__init__")
-            cls = dataclasses.dataclass(cls)
+            cls = dataclasses.dataclass(repr=False)(cls)
             setattr(cls, _USER_DEFINED_DEFAULTS, user_defined_defaults)
 
         descriptor = _dataobject_to_proto(dataclass_=cls, **kwargs)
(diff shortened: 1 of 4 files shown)


prashantgupta24 (approved):

caikit/core/data_model/dataobject.py
```diff
@@ -200,7 +200,7 @@ def decorator(cls: _DataObjectBaseT) -> _DataObjectBaseT:
             if _has_dataclass_init(cls):
                 log.debug3("Resetting default dataclass init")
                 delattr(cls, "__init__")
-            cls = dataclasses.dataclass(cls)
+            cls = dataclasses.dataclass(repr=False)(cls)
```
gkumbhat, line 203: thats a neat little trick

gkumbhat (approved):
LGTM

PR #439, opened 2023-08-24
🐛  ServerBootLogMessage: Use the right config path in the log message!

gabe-l-hart:

When we added the `runtime.grpc` level of indentation in config for the `grpc/http` split, we missed this log message!


gkumbhat (approved):
LGTM.

PR #441, opened 2023-08-28
Oneof fix

aluu317:

The previous code relies on the existence of `_sequence` in the field name for union of list types, or splitting on underscores `_` to get the oneof name (how silly!). 

This PR:
- removes the check for `_sequence` because users could name their oneof field names custom names without the word `sequence` and without `_`.
- lots of tests to cover the cases of default union of list names when users don't provide a field name (ex: "foo_int_sequence") as well as when they provide custom names (ex: "myfooseq")

**If applicable**:

### DIFF
--- a/caikit/core/data_model/base.py
+++ b/caikit/core/data_model/base.py
@@ -670,11 +670,7 @@ class DataBase(metaclass=_DataBaseMetaClass):
             if field in cls._fields_primitive or field in cls._fields_enum:
                 # special case for oneofs
                 if field not in cls._fields_to_oneof or proto.HasField(field):                    
-                    if proto.HasField(field): # this is a union type
-                        oneof = cls._fields_to_oneof.get(field)
-                        kwargs[oneof] = proto_attr
-                    else:
-                        kwargs[field] = proto_attr
+                    kwargs[field] = proto_attr
             elif (
                 field in cls._fields_primitive_repeated
                 or field in cls._fields_enum_repeated
--- a/caikit/core/data_model/base.py
+++ b/caikit/core/data_model/base.py
@@ -902,7 +902,11 @@ class DataBase(metaclass=_DataBaseMetaClass):
         to_dict = {}
         for field in fields_to_dict:
             dict_value = self._field_to_dict_element(field)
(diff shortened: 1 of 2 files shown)


caikit/core/data_model/base.py
```diff
@@ -666,12 +669,8 @@ def from_proto(cls, proto):
                 )
 
             if field in cls._fields_primitive or field in cls._fields_enum:
-                if field in cls._fields_to_oneof:
-                    if proto.HasField(field):
-                        # "foo_bar_int" has original field name "foo_bar"
-                        original_field_name = "_".join(field.split("_")[:-1])
-                        kwargs[original_field_name] = proto_attr
-                else:
+                # special case for oneofs
```
aluu317, line 672: This reverts the changes introduced by https://github.com/caikit/caikit/pull/350/files. We don't need this change

tests/core/data_model/test_dataobject.py
```diff
@@ -575,6 +634,101 @@ class Foo(DataObjectBase):
     assert Foo.from_json(json_repr_foo) == foo1
 
 
+def test_dataobject_list_oneof_round_trips_with_specified_names():
+    @dataobject
+    class Foo(DataObjectBase):
+        foo: Union[
+            Annotated[List[int], FieldNumber(10), OneofField("myfoointseq")],
+            Annotated[List[str], FieldNumber(20), OneofField("myfoostrseq")],
```
joerunde, line 642: quick question: What's different between this test and the one above that has
```
    class Foo(DataObjectBase):
        foo: Union[
            Annotated[int, FieldNumber(10), OneofField("myfooint")],
            Annotated[float, FieldNumber(20), OneofField("myfoofloat")],
        ]
```

I can't quickly tell if the only difference here is the names or if there's something else different going on

joerunde (approved):
Tests look nice to me, just had one question

tests/core/data_model/test_dataobject.py
aluu317: This one is union of lists, the above is union of int and float

PR #456, opened 2023-09-06
♻️ put pydantic stuff into its own module

### LINKED ISSUE #451: Refactor pydantic stuff out of http_server
## Description

As a Caikit maintainer, I want to refactor `http_server` based on https://github.com/caikit/caikit/pull/434's review comments.

## Acceptance Criteria

- [ ] Unit tests cover new/changed code
- [ ] Examples build against new/changed code
- [ ] READMEs are updated
- [ ] Type of [semantic version](https://semver.org/) change is identified

prashantgupta24:

Refactors `http_server` and pulls out all `pydantic` specific stuff into its own class, `pydantic_wrapper`. They both now live under `runtime/http_server` and not `core/data_model` as suggested previously because `pydantic_wrapper` needs access to `interfaces`, and `core` should not be importing anything from `interfaces` directly.

Unit tests changed accordingly.

fix #451 

**If applicable**:


caikit/runtime/wrappers/pydantic_wrapper.py
```diff
@@ -0,0 +1,162 @@
...
+This module holds the Pydantic wrapping required by the REST server,
+capable of converting to and from Pydantic models to our DataObjects.
+"""
+# Standard
+from typing import Dict, List, Type, Union, get_args, get_type_hints
+import enum
+
+# Third Party
+import numpy as np
+import pydantic
```
gabe-l-hart, line 24: The only thing that _might_ be problematic about this is that depending on how we manage these dependencies (proactive vs reactive), `pydantic` _might_ end up becoming a required dependency when we intend to keep it in the `runtime-http` set. This makes me think that it might make sense to have a `runtime/http_server` module and put the current `http_server.py` plus this file in there so that it's clear that's the only module where `pydantic` and `fastapi` are required.

caikit/runtime/wrappers/pydantic_wrapper.py
```diff
@@ -0,0 +1,162 @@
...
+            **{
+                name: None
+                for name, _ in get_type_hints(
+                    dm_class,
+                    localns=localns,
+                ).items()
+            },
+        },
+    )
+    PYDANTIC_TO_DM_MAPPING[dm_class] = pydantic_model
```
gabe-l-hart, line 115: NIT: Now that this is in a free function and _might_ conceivably be used somewhere besides the `http_server`, this kind of global state mutation worries me a bit. If we keep this in a generic module like `runtime.wrappers`, I think we should at least put a `threading.Lock` around access to the dict. If we move this into a `runtime.http_server` module, then I think it's fine to leave unlocked since we _know_ it won't be used from a multithreaded context.

gabe-l-hart (commented):
This is a really nice cleanup Prashant! My one suggestion is that we refactor a bit to make it clear that this logic is still explicitly coupled with the `http_server`.

caikit/runtime/wrappers/pydantic_wrapper.py
prashantgupta24: Sounds good!

caikit/runtime/__init__.py
```diff
@@ -11,3 +11,6 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
+# Local
+from .http_server import RuntimeHTTPServer
```
gabe-l-hart, line 16: I think we actually _explicitly_ want to not do this. Having this import proactively will cause all of the HTTP server dependencies to become required for _any_ usage of `caikit.runtime` (including `caikit.runtime.grpc_server`)

caikit/runtime/http_server/__init__.py
```diff
@@ -0,0 +1,17 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Local
+from .http_server import HEALTH_ENDPOINT, RuntimeHTTPServer
+from .pydantic_wrapper import dataobject_to_pydantic, pydantic_to_dataobject
```
gabe-l-hart, line 17: I think we also want to add a `main` clause to this module. I'm not _entirely_ sure the right way to do that, but we want to make sure that `python -m caikit.runtime.http_server` still works. I think it would either be:

* Add a `__main__.py` file that invokes `http_server.http_server.main`
* Add an `if __name__ == "__main__":` clause in this file.

gabe-l-hart: Also, this hasn't changed in this PR, but I think we need to update the import logic in `caikit/runtime/__main__.py` as well. Currently it [proactively imports the HTTP server](https://github.com/caikit/caikit/blob/main/caikit/runtime/__main__.py#L10C40-L10C57) which would mean that you can't use that entrypoint to run just the `grpc_server` if you didn't have the `runtime-http` extras set installed.

gabe-l-hart (changes requested):
A couple more changes to keep the `runtime-http` deps truly optional

caikit/runtime/__init__.py
prashantgupta24: ah that makes sense

caikit/runtime/http_server/__init__.py
prashantgupta24: I added a `__main__.py` file for the first point. 

For the second point, I am planning to open a separate issue and have a subsequent PR!

gabe-l-hart: Beautiful!

gabe-l-hart (approved):
Ship it. Really nice improvement!

PR #459, opened 2023-09-07
Convert to json before returning fastapi response

HonakerM:
This PR updates the `http_server` runtime to directly call a data object's `to_json` function before returning the fastapi Response. This allows users to customize a models json representation instead of defaulting to fastapi's [jsonable_encoder](https://fastapi.tiangolo.com/tutorial/encoder/?h=jsonable_enc#using-the-jsonable_encoder)

This is my first PR to caikit so please let me know if there is anything I should do differently!

**If applicable**:

### DIFF
--- a/caikit/core/data_model/base.py
+++ b/caikit/core/data_model/base.py
@@ -611,6 +611,10 @@ class DataBase(metaclass=_DataBaseMetaClass):
         if field_descriptor.type == field_descriptor.TYPE_BOOL:
             return isinstance(val, bool)
 
+        # Proto doesn't allow non utf-8 bytes fields; however, python does.
+        if field_descriptor.type == field_descriptor.TYPE_BYTES:
+            return isinstance(val, bytes)
+
         # If it's a primitive, use protobuf type checkers
         checker = proto_type_checkers.GetTypeChecker(field_descriptor)
         try:
@@ -933,7 +937,7 @@ class DataBase(metaclass=_DataBaseMetaClass):
             - datetime.datetime
             """
             if isinstance(obj, bytes):
-                return base64.encodebytes(obj).decode("utf-8")
+                return base64.b64encode(obj).decode("utf-8")
             if isinstance(obj, datetime.datetime):
                 # Use the timestamp's proto-serialized format to get the proper json serializer
                 return timestamp.datetime_to_proto(obj).ToJsonString()
--- a/caikit/runtime/http_server/pydantic_wrapper.py
+++ b/caikit/runtime/http_server/pydantic_wrapper.py
@@ -17,13 +17,13 @@ capable of converting to and from Pydantic models to our DataObjects.
 """
 # Standard
 from typing import Dict, List, Type, Union, get_args, get_type_hints
-import enum
 import base64
+import enum
 
 # Third Party
+from pydantic.functional_validators import BeforeValidator
 import numpy as np
 import pydantic
-from pydantic.functional_validators import BeforeValidator
 
 # First Party
 from py_to_proto.dataclass_to_proto import (  # Imported here for 3.8 compat
@@ -166,7 +166,7 @@ def _get_pydantic_type(field_type: type) -> type:
     raise TypeError(f"Cannot get pydantic type for type [{field_type}]")
 
 
-def _from_base64(data: Union[bytes, str])->bytes:
+def _from_base64(data: Union[bytes, str]) -> bytes:
     if isinstance(data, str):
         return base64.b64decode(data.encode("utf-8"))
     return data
--- a/caikit/runtime/http_server/pydantic_wrapper.py
+++ b/caikit/runtime/http_server/pydantic_wrapper.py
@@ -18,10 +18,12 @@ capable of converting to and from Pydantic models to our DataObjects.
 # Standard
 from typing import Dict, List, Type, Union, get_args, get_type_hints
 import enum
+import base64
 
 # Third Party
 import numpy as np
 import pydantic
+from pydantic.functional_validators import BeforeValidator
 
 # First Party
 from py_to_proto.dataclass_to_proto import (  # Imported here for 3.8 compat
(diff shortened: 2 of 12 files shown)


gabe-l-hart (changes requested):
This is great, thanks Casey! Two things:

1. This will need a rebase after Prashant's recent change to move things around (#456)
2. Let's throw in a unit test with your example of non-utf8 bytes

tests/fixtures/sample_lib/modules/file_processing/bounding_box_module.py
```diff
@@ -0,0 +1,31 @@
...
+# Local
+from ...data_model.sample import FileTask, FileDataType
+from caikit.core.modules import ModuleLoader
+import caikit.core
+
+
+@caikit.core.module(
+    "750cad4d-c3b8-4327-b52e-e772f0d6f311", "BoundingBoxModule", "0.0.1", FileTask
+)
+class BoundingBoxModule(caikit.core.ModuleBase):
```
HonakerM, line 15: This is my first time creating a caikit module so please let me know if I missed anything or didn't follow best practices

caikit/core/data_model/base.py
```diff
@@ -933,7 +937,7 @@ def _default_serialization_overrides(obj):
             - datetime.datetime
             """
             if isinstance(obj, bytes):
-                return base64.encodebytes(obj).decode("utf-8")
+                return base64.b64encode(obj).decode("utf-8")
```
gabe-l-hart, line 940: Interesting. What exactly does this change do? I've never fully known the difference between these two methods

caikit/runtime/http_server/pydantic_wrapper.py
```diff
@@ -160,3 +164,9 @@ def _get_pydantic_type(field_type: type) -> type:
         ]
 
     raise TypeError(f"Cannot get pydantic type for type [{field_type}]")
+
+
+def _from_base64(data: Union[bytes, str])->bytes:
```
gabe-l-hart, line 169: Super NIT:

```suggestion
def _from_base64(data: Union[bytes, str]) -> bytes:
```

tests/fixtures/sample_lib/modules/file_processing/bounding_box_module.py
```diff
@@ -0,0 +1,29 @@
+"""
```
gabe-l-hart, line 1: NIT: Copyright header

tests/fixtures/sample_lib/modules/file_processing/bounding_box_module.py
```diff
@@ -0,0 +1,29 @@
...
+A sample module for sample things!
+"""
+# Standard
```
gabe-l-hart, line 4: NIT: No need for `# Standard` if there are no standard imports

gabe-l-hart (changes requested):
A couple of tiny NITs, but this looks really good. Thanks for digging through it all!

caikit/core/data_model/base.py
HonakerM: Initially I thought `encodebytes` was just the legacy interface to `b64encode` but it turns out that [encodebytes](https://docs.python.org/3/library/base64.html#base64.encodebytes) adds new line characters every 76 bytes which messed with tests. I'm not sure it actually matters though as parsers/validators should still handle it

tests/fixtures/sample_lib/modules/file_processing/__init__.py
```diff
@@ -0,0 +1,2 @@
+# Local
```
gabe-l-hart, line 1: NIT: Copyright header here too (and anywhere else that's new)! We really need to add a CI check for this

tests/fixtures/sample_lib/modules/file_processing/bounding_box_module.py
gabe-l-hart: This looks great! A very nice simple dummy module implementation

gabe-l-hart (changes requested):
One more pass to add copyright headers, then it's good to go

tests/fixtures/sample_lib/modules/file_processing/__init__.py
HonakerM: Added! Do only `.py` files need it or should I also add a copyright header  to [config.yaml](https://github.com/caikit/caikit/blob/bb73b8a0bd260d18d7dff7d2f154ab8a81f049c5/tests/fixtures/models/box/config.yml)?

prashantgupta24: @HonakerM You will need to add the copyright to all files, including `yml` files!

tests/runtime/conftest.py
```diff
@@ -251,6 +251,23 @@ def sample_task_model_id(good_model_path) -> str:
...
+    # model load test already tests with archive - just using a model path here
+    model_manager.load_model(
+        model_id,
+        local_model_path=box_model_path,
+        model_type=Fixtures.get_good_model_type(),  # eventually we'd like to be determining the type from the model itself...
+    )
+    yield model_id
+
+    # teardown
+    model_manager.unload_model(model_id)
```
prashantgupta24, line 268: Worth having a central function in `conftest` that does all this ? I see a lot of repeated code. Can be tackled in a subsequent PR!

gabe-l-hart: I think this is in `runtime/conftest` because it's using the runtime's `ModelManager`, right?

gabe-l-hart (approved):
Looks great

tests/runtime/conftest.py
prashantgupta24: Ah I meant a function within `runtime/conftest` itself to avoid repeated code, we've used the same logic to create all `*model_id`

Issue #460, opened 2023-09-07
🐛 Data model objects with bytes don't round trip to/from json


gabe-l-hart:
## Describe the bug

A data model instance, when loaded from its `json` representation should match the original object. If the object contains a `bytes` field, this does not happen due to the bytes being `base64` serialized in [to_json](https://github.com/caikit/caikit/blob/main/caikit/core/data_model/base.py#L935) but not in [from_json](https://github.com/caikit/caikit/blob/main/caikit/core/data_model/base.py#L741).

## Sample Code

```py
from caikit.core.data_model import dataobject, DataObjectBase

@dataobject
class Foo(DataObjectBase):
    foo: bytes

f = Foo("asdf")
f2 = Foo.from_json(f.to_json())
# This fails!
assert f.foo == f2.foo
```

## Expected behavior

`f.foo == f2.foo` should be `True`

prashantgupta24:
@gabe-l-hart Based on recent discussions, is this bug irrelevant? If the intention is to `reject attempts to set fields with the wrong type`, then ideally
```
f = Foo("asdf")
```
is incorrect
```
f = Foo(b"asdf")
```
and using this works as expected?

prashantgupta24:
And ideally we should open a separate issue in which we explicitly `reject attempts to set fields with the wrong type`?

PR #461, opened 2023-09-08
🐛 fix Data model objects with Bytes bug

prashantgupta24:

This PR adds a couple of test cases to make sure that #460 (round tripping a DM object with bytes) works fine if we explicitly declare a `bytes` object.

There was quite a bit of back-and-forth regarding this PR. I will try to summarize the discussion.

When Gabe created the issue #460, he tried to create a DM object having a `bytes` field with a `string`, and that was failing. In the original implementation to fix this, I manually checked if the object type was a `string` and the field type was `bytes`, and I explicitly converted the `string` to `bytes`, which made the failing test work.

However, Gabe pointed out that if we go down this route, this type of conversion will have to be done holistically, which presents a lot of complexity in terms of figuring out how to convert various types to another.

So Gabe came up with 3 options:

1. Type cast (what we tried doing) - Really hard to do holistically.
2. Pedantically reject (if it's not a bytes you can't use it for TYPE_BYTES)
3. Be "pythony" and let whatever happens happen - the thing with python in this case is that it's pretty standard practice to use a type hint to tell users "this should be bytes" but then not actually bother checking at runtime. Since the thing we're building explicitly cares about typing, that's **not really a great user experience**, especially with "similar" types (e.g. `int` being assigned to a `float` field).

Hence we decided to reject the 1st and 3rd idea, and instead go with `#2` - tracked here: https://github.com/caikit/caikit/issues/464.

**If applicable**:

### DIFF
--- a/caikit/core/data_model/base.py
+++ b/caikit/core/data_model/base.py
@@ -477,20 +477,7 @@ class DataBase(metaclass=_DataBaseMetaClass):
 
         # If attempting to set one of the named fields or a oneof, instead set
         # the private version of the attribute.
-        if name in cls.fields:
-            field_descriptor = cls._proto_class.DESCRIPTOR.fields_by_name[name]
-            # for bytes, convert to bytes if not already
-            if field_descriptor.type == field_descriptor.TYPE_BYTES:
-                error.value_check(
-                    "<COR13072712E>",
-                    isinstance(val, (bytes, str)),
-                    "{} must be string or bytes",
-                    val,
-                )
-                if isinstance(val, str):
-                    val = val.encode("utf-8")
-            super().__setattr__(f"_{name}", val)
-        elif name in cls._fields_oneofs_map:
+        if name in cls.fields or name in cls._fields_oneofs_map:
             super().__setattr__(f"_{name}", val)
         else:
             super().__setattr__(name, val)
--- a/caikit/core/data_model/base.py
+++ b/caikit/core/data_model/base.py
@@ -18,7 +18,7 @@
 # Standard
 from dataclasses import dataclass
 from enum import Enum
-from typing import Any, Dict, List, Optional, Type, Union, get_type_hints
+from typing import Any, Dict, List, Optional, Type, Union
 import base64
 import datetime
 import json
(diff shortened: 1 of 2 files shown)


caikit/core/data_model/base.py
```diff
@@ -477,7 +477,20 @@ def __setattr__(self, name, val):
...
+            if field_descriptor.type == field_descriptor.TYPE_BYTES:
+                error.value_check(
+                    "<COR13072712E>",
+                    isinstance(val, (bytes, str)),
+                    "{} must be string or bytes",
+                    val,
+                )
+                if isinstance(val, str):
+                    val = val.encode("utf-8")
+            super().__setattr__(f"_{name}", val)
```
prashantgupta24, line 492: I'm curious why we aren't checking other field types while doing the `__setattr__` 🤔

gabe-l-hart: Hmm. This is an interesting. It seems like the logical place to introduce type casting, but that also feels like it could be a pretty big can of worms to open with the potential for a major performance hit to run through all of the `if isinstance(...)` checks. The alternative would be to simply reject attempts to set fields with the wrong type. I think I might already have something like this written somewhere...

caikit/core/data_model/base.py
```diff
@@ -477,7 +477,20 @@ def __setattr__(self, name, val):
...
+            field_descriptor = cls._proto_class.DESCRIPTOR.fields_by_name[name]
+            # for bytes, convert to bytes if not already
+            if field_descriptor.type == field_descriptor.TYPE_BYTES:
+                error.value_check(
+                    "<COR13072712E>",
+                    isinstance(val, (bytes, str)),
+                    "{} must be string or bytes",
+                    val,
+                )
+                if isinstance(val, str):
```
prashantgupta24, line 490: I guess technically we can convert other types to bytes as well

gabe-l-hart: Right, this is my concern with going this route. You could easily cast an `int` to `bytes` too

caikit/core/data_model/base.py
prashantgupta24: So in that case, we would reject `f2 = Foo("asdf")` for `foo: bytes`?

prashantgupta24: Did you mean the `checker = proto_type_checkers.GetTypeChecker(field_descriptor)` code for checking the type?

gabe-l-hart: Yes! That's what I was trying to remember

prashantgupta24: This idea starts complaining about things like this now:
```
@dataobject
    class Foo(DataObjectBase):
        foo: str
        bar: int

    inst = Foo()
```
I'm guessing introducing type-checking within `setattr` is going to break a bunch of stuff

gabe-l-hart (approved):
Thanks for bearing with all the back-and-forth Prashant!

Issue #251, opened 2023-06-19
Swap all server tests to use datamodel classes directly


prashantgupta24:
## Describe the bug

Right now, we are building the grpc request directly, but we can use datamodel classes directly instead. Additional bonus - all the server tests would cover the `.to_protoing` of request messages as well

## Platform

Please provide details about the environment you are using, including the following:

- Interpreter version: 3.9
- Library version: 0.8.0

## Additional context

```
train_request_class = DataBase.get_class_for_name("OtherTaskOtherModuleTrainRequest")
train_request = train_request_class(
    model_name = "Other module Training",
    training_data = training_data,
    simple_list=["hello", "world"],
    list_value_str_sequence=train_request_class.StrSequence(
        values=["str", "sequence"]
    ),
    sample_input_str="sample",
    batch_size=batch_size
).to_proto()
```

see `tests/runtime/test_grpc_server.py` for both ways of doing the same thing.

PR #468, opened 2023-09-13
test: update tests to use datamodel classes

anhuong:

closes: #251 

**If applicable**:

### DIFF
--- a/tests/runtime/utils/test_servicer_util.py
+++ b/tests/runtime/utils/test_servicer_util.py
@@ -336,6 +336,9 @@ def test_global_predict_build_caikit_library_request_dict_works_for_set_primitiv
     assert request_dict["list_type"] == ["1", "2", "3"]
 
 
+@pytest.mark.skip(
+    "Skipping until bug fixes for unset Union fields - https://github.com/caikit/caikit/issues/471"
+)
 def test_global_train_build_caikit_library_request_dict_strips_empty_list_from_request(
     sample_train_service,
 ):
@@ -358,9 +361,7 @@ def test_global_train_build_caikit_library_request_dict_strips_empty_list_from_r
 
     # model_name is not expected to be passed through
     # since using pythonic data model, keeps params that have default value and removes ones that are empty
-    # TODO: create issue for union_list that should be removed
     expected_arguments = {
-        "union_list",
         "sleep_time",
         "oom_exit",
         "sleep_increment",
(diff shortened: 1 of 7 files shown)


tests/runtime/utils/test_servicer_util.py
```diff
@@ -110,7 +111,9 @@ def test_service_util_validate_caikit_library_class_method_exists_does_raise():
 
 
 # ---------------- Tests for build_proto_response  --------------------------
-def test_servicer_util_build_proto_response_raises_on_garbage_response_type():
+def test_servicer_util_build_proto_response_raises_on_garbage_response_type(
+    sample_inference_service,
```
gabe-l-hart, line 115: Is this so that it's not relying on side effects?

tests/runtime/utils/test_servicer_util.py
```diff
@@ -124,23 +127,19 @@ def test_servicer_util_is_protobuf_primitive_returns_true_for_primitive_types(
...
-            "int_type"
-        ]
+        predict_class().to_proto().DESCRIPTOR.fields_by_name["int_type"]
     )
 
 
-def test_servicer_util_is_protobuf_primitive_returns_false_for_custom_types(
-    sample_inference_service,
-):
+def test_servicer_util_is_protobuf_primitive_returns_false_for_custom_types():
```
gabe-l-hart, line 136: Do we no longer need `sample_inference_service` top ensure that `SampleTaskRequest` is present?

gabe-l-hart (approved):
Changes all look good! I have a couple of questions about some of the changes to fixtures in some of the tests, but nothing blocking

tests/runtime/utils/test_servicer_util.py
anhuong: A good question...I think I accidentally added this in here when meaning to just add it to the below method

tests/runtime/utils/test_servicer_util.py
anhuong: I'm not sure why for `test_servicer_util_is_protobuf_primitive_returns_true_for_primitive_types()` it is needed else I get error ` ValueError: No data model class match for SampleTaskRequest` as you mention but I don't get that error for this test....not sure why though since they're so similar.... 🤔

anhuong:
I can't merge this PR but @gabe-l-hart @prashantgupta24 it should be good to go

PR #469, opened 2023-09-13
Fix duplicate logcode

gkumbhat:
### Changes
- Fix duplicate logcode. As seen here: https://github.com/search?q=repo%3Acaikit%2Fcaikit%20%3CRUN490439039W%3E&type=code


prashantgupta24 (approved):
Good catch, thanks!

PR #472, opened 2023-09-14
Find regressed modules for backwards compatibility check

aluu317:
We need a check to compare a set of modules supported in a prior service generation to this current generation.
Raise an error if at least a module becomes unsupported, hence breaking backwards compatibility.

This PR:
- adds a function `assert_compatible()` in create_service.py to assert that no module was priorly supported is no longer.
- adds a flag in `get_service_package()` to write out a `modules.json` file. Should this just be an always behavior/ not a flag?
- assumes that we have a way to give `assert_compatible()` a prev module list to do the comparison with. This PR is a first round of putting it in the config.yml as a file path, but can be defined in a better way for libraries to input this.
- 

**If applicable**:


gabe-l-hart (commented):
A couple of thoughts/comments around the error handling, but otherwise this looks great!

caikit/runtime/service_factory.py
```diff
@@ -130,9 +136,27 @@ def get_service_package(cls, service_type: ServiceType) -> ServicePackage:
...
+        )
+        if backwards_compat_conf and backwards_compat_conf.enabled:
+            previous_included_modules = set()
+            prev_modules_path = backwards_compat_conf.prev_modules_path
+            if prev_modules_path:
+                with open(prev_modules_path, "r", encoding="utf-8") as f:
+                    previous_modules = json.load(f)
+                    previous_included_task_map = previous_modules["included_modules"]
+                    for task_module in previous_included_task_map.values():
+                        previous_included_modules.update(task_module.values())
```
joerunde, line 154: 🤔  Do we want to check by `guid` instead of by name? (`task_module.keys()` instead here)

Checking by name doesn't allow module authors to refactor their module naming or packaging. I'd argue that both of those are `python` API breaking changes, but I don't think they're `runtime` API breaking changes since the runtime apis don't actually care about the name of the module (for inference, at least) nor where it is packaged within the library.

tests/runtime/test_service_factory.py
```diff
@@ -213,6 +216,91 @@ def test_get_and_filter_modules_respects_included_task_types_and_excluded_module
...
+                )
+
+
+def test_assert_compatible_does_not_raise_if_supported_modules_continue_to_be_supported():
+    with tempfile.TemporaryDirectory() as workdir:
+        prev_module_file = os.path.join(workdir, "prev_modules.json")
+
+        with open(prev_module_file, "w", encoding="utf-8") as file:
+            json_content = {
+                "excluded_modules": {},
```
joerunde, line 277: 🤔🤔🤔🤔🤔

Do we need excluded modules in this file at all?
It looks like we don't need it for the check in this PR, and I'm not sure if there's any other check we would need it for. IIRC we had used this previously when service generation was implicit, but now that it's explicit the service generation will fail if it can't generate a handler for a module that lists a task

caikit/runtime/service_factory.py
```diff
@@ -130,9 +136,27 @@ def get_service_package(cls, service_type: ServiceType) -> ServicePackage:
...
         clean_modules = ServicePackageFactory._get_and_filter_modules(
-            caikit_config, caikit_config.runtime.library
+            caikit_config, caikit_config.runtime.library, write_modules_file
         )
 
+        # Assert for backwards compatibility, if enabled
+        backwards_compat_conf = (
+            caikit_config.runtime.service_generation.backwards_compatibility
+        )
+        if backwards_compat_conf and backwards_compat_conf.enabled:
```
joerunde, line 146: Can we stick this backwards compatibility checking logic into a private method to keep this public one as concise as possible?

tests/runtime/test_service_factory.py
```diff
@@ -213,6 +216,91 @@ def test_get_and_filter_modules_respects_included_task_types_and_excluded_module
...
+                    "backwards_compatibility": {
+                        "enabled": False,
+                    }
+                },
+            }
+        },
+        "merge",
+    ):
+        ServicePackageFactory.get_service_package(
+            ServicePackageFactory.ServiceType.INFERENCE
```
joerunde, line 233: nit: it looks like this code is implemented to be used for both `TRAINING` and `INFERENCE` so we probably want to either:
- Add tests for training, or
- Only implement it for inference

examples/sample_lib/start_runtime_with_sample_lib.py
```diff
@@ -61,3 +61,4 @@
     finally:
         # remove generated protos
         shutil.rmtree("protos", ignore_errors=True)
+        os.remove("modules.json")
```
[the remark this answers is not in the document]
gabe-l-hart: 🤦 I think I missed that this was in the `sample_lib` stuff. Carry on

caikit/runtime/service_factory.py
gabe-l-hart: Addressed to use `MODULE_ID`

tests/runtime/test_service_factory.py
gabe-l-hart: Implemented to only work against `INFERENCE`

gabe-l-hart (commented):
Thanks for addressing all the comments @aluu317 and sorry for the slow review. It looks like there are a number of conflicts now. If you can get this rebased to resolve this, I think it's good to go!

gabe-l-hart (approved):
Looks good!

Issue #470, opened 2023-09-13
Make DMs work with compiled protobufs


prashantgupta24:
## Describe the bug

Currently, we support the use-case in which users have either:
-  Full access to the library as well as the DM and they use Caikit to generate the grpc stub
- No access to library or the DM, in which case they use the compiled protos.

But a third scenario is becoming common, in which:
- Users don't have access to the library, however, they have access to the publicly available DM.

This use-case presents some challenges, most notably conflicts between proto files using the same descriptor pool.

## Platform

Please provide details about the environment you are using, including the following:

- Interpreter version: 3.9
- Library version:

PR #473, opened 2023-09-14
✨ add a client for sample_lib

prashantgupta24:

This PR:
1. Adds a `client.py` to the `sample_lib` example package, so that users can interact with a python client to a running caikit runtime server.
2. Adds a `client_proto.py` to the `sample_lib` example package, that demonstrates how to use compiled `pb2s` along with the `DataModel` for a library (fix https://github.com/caikit/caikit/issues/470)
3. Updates documentation
 
This PR also tried to add a test for the `client.py` above, but until we figure out how to parallelize the tests, it's not worth adding it in because it takes a long time to set it up.

**If applicable**:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,3 +35,6 @@ _build/
 
 
 /caikit/_version.py
+# Compiled pb2s
+*_pb2.py
+*_pb2_grpc.py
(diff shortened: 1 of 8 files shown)


joerunde (approved):
noice!

PR #475, opened 2023-09-15
✨ add a better way to get request DMs

prashantgupta24:

for https://github.com/caikit/caikit/issues/476.

Essentially, we should now be able to do this:
```
caikit.runtime.get_train_request(sample_module.modules.SampleModule)(
    model_name=model_id,
    parameters=get_train_params(sample_lib.modules.SampleModule)(
        training_data=training_data
    ),
)
```

There are still some tests left that will need to be changed from `DataBase.get_class...` to this

**If applicable**:

### DIFF
--- a/caikit/runtime/__init__.py
+++ b/caikit/runtime/__init__.py
@@ -13,4 +13,4 @@
 # limitations under the License.
 
 # Local
-from .service_factory import get_request, get_train_params, get_train_request
+from .service_factory import get_inference_request, get_train_params, get_train_request
--- a/caikit/runtime/__init__.py
+++ b/caikit/runtime/__init__.py
@@ -11,3 +11,6 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
+# Local
+from .service_factory import get_request, get_train_params, get_train_request
--- a/caikit/runtime/service_factory.py
+++ b/caikit/runtime/service_factory.py
@@ -15,7 +15,7 @@
 # Standard
 from enum import Enum
 from types import ModuleType
-from typing import Callable, Dict, Set, Type
+from typing import Callable, Dict, Set, Type, Union
 import dataclasses
 
 # Third Party
@@ -33,6 +33,7 @@ from caikit import get_config
 from caikit.core import LocalBackend, ModuleBase, registries
 from caikit.core.data_model.base import DataBase
 from caikit.core.data_model.dataobject import _AUTO_GEN_PROTO_CLASSES
+from caikit.core.task import TaskBase
 from caikit.interfaces.runtime.data_model import (
     TrainingInfoRequest,
     TrainingStatusResponse,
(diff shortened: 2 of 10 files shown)


caikit/runtime/service_factory.py
```diff
@@ -266,3 +267,38 @@ def _get_and_filter_modules(
...
+
+
+def get_request(
+    module_class: Type[ModuleBase],
+    input_streaming: bool = False,
+    output_streaming: bool = False,
+) -> Type[DataBase]:
+    """Helper function to return the request DataModel for the Module Class"""
+    if input_streaming and output_streaming:
+        request_class_name = f"BidiStreaming{module_class.TASK_CLASS.__name__}Request"
```
joerunde, line 279: @prashantgupta24 since this only requires the task, what do you think about supplying the task as the parameter instead of the module?

I'm thinking ahead a little bit here: we've had requests for multi-modal modules that support multiple tasks so we might also have a future where there are multiple inference requests for a single module

caikit/runtime/service_factory.py
```diff
@@ -266,3 +267,38 @@ def _get_and_filter_modules(
             excluded_modules,
         )
         return clean_modules
+
+
+def get_request(
```
joerunde, line 272: Should we be explicit here that this is `get_inference_request`?

caikit/runtime/__init__.py
```diff
@@ -11,3 +11,6 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
+# Local
+from .service_factory import get_request, get_train_params, get_train_request
```
gabe-l-hart, line 16: Small note: I know that in `caikit.runtime` we've NOT done proactive imports. I think this is a fine thing to do since it surfaces these functions at a module level, but we do want to be a little careful about what (if any) import-time code gets run by performing these proactive imports.

caikit/runtime/service_factory.py
```diff
@@ -266,3 +267,38 @@ def _get_and_filter_modules(
...
+        request_class_name = f"{module_class.TASK_CLASS.__name__}Request"
+    return DataBase.get_class_for_name(request_class_name)
+
+
+def get_train_request(module_class: Type[ModuleBase]) -> Type[DataBase]:
+    """Helper function to return the train request DataModel for the Module Class"""
+    request_class_name = (
+        f"{module_class.TASK_CLASS.__name__}{module_class.__name__}TrainRequest"
+    )
+    print(request_class_name)
```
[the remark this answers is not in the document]
gabe-l-hart: Yeah, this needs to be removed

gabe-l-hart (changes requested):
I think this looks great. Two change requests following Joe's review:

* `get_request` -> `get_infernce_request`
* No `print`s

Also, I think Joe's suggestion of also supporting taking the `task` directly would be useful too.

caikit/runtime/service_factory.py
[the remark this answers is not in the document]
prashantgupta24: :O this is embarrasing!

gabe-l-hart (approved):
Looks good, pending any additional changes to also take `task` classes in the functions

prashantgupta24:
Thanks @gabe-l-hart and @joerunde for the review! I will make the changes. I still need to add and change a bunch of test cases before I can get this in

caikit/runtime/service_factory.py
```diff
@@ -266,3 +268,43 @@ def _get_and_filter_modules(
             excluded_modules,
         )
         return clean_modules
+
+
+def get_inference_request(
+    class_name: Type[Union[ModuleBase, TaskBase]],
```
prashantgupta24, line 274: @gabe-l-hart @joerunde does this name make sense? I have a feeling a `name` in the parameter suggests a `string` type, but I couldn't think of something that includes both `module` and a `task`

joerunde: hmmm, `honest_and_complete` naming spec might be `task_or_module_class`

PR #479, opened 2023-09-18
Mt save path 478

### LINKED ISSUE #478: MT Service Prefer Global Training Output

In the [ModelTrainServicerImpl](https://github.com/caikit/caikit/blob/main/caikit/runtime/servicers/model_train_servicer.py#L107) the `request.training_output_dir` will _always_ point to a local `emptyDir` mount that MT "owns." This works well when the output will be uploaded to S3, but if no S3 output path is given, there's no way to get MT to set this value to a pre-mounted PVC and then not purge the content after the successful training. We need a way to work around this behavior from MT so that user-defined PVCs can be used as the save target for trained models.

In order to support saving to a non-S3 PVC, `caikit` should prefer the globally configured `runtime.training.output_dir` value over `request.training_output_dir` when fielding an MT `Process` request.

* Save the model to both `runtime.training.output_dir` AND `request.training_output_dir`
* Update MT to better support this usecase

## Additional context

We plan to remove all of this MT functionality from the public `caikit` architecture in short order and do not want to add additional config surface to support this.

gabe-l-hart:
**NOTE**: This is an API breaking change as it will change the behavior of training jobs to result in models failing to be updated unless corresponding changes are made in the server's config to set `runtime.training.output_dir` to `""`.

Closes #478 

This PR adjusts the behavior of the Model Train servicer so that if `runtime.training.output_dir` is configured in the global config, the model will be saved there rather than in `request.training_output_dir`. This is needed to support the usecase of saving models to mounted PVCs rather than uploading them to S3 when S3 is not available.

**If applicable**:

### DIFF
--- a/tests/runtime/servicers/test_model_train_servicer_impl.py
+++ b/tests/runtime/servicers/test_model_train_servicer_impl.py
@@ -185,38 +185,66 @@ def test_model_train_validation_error_raises(sample_model_train_servicer, output
 
 #####################################################################
 # Normal tests
-def test_model_train_sample_widget(sample_model_train_servicer, output_dir):
+@pytest.mark.parametrize("output_in_config", (True, False))
+def test_model_train_sample_widget(
+    sample_model_train_servicer,
+    output_dir,
+    output_in_config,
+):
+    """This test tests end-to-end training. It includes verifying that the model
+    is saved in the right place. The place where the model is saved comes from:
+
+    1. get_config().runtime.training.output_dir
+    2. request.training_output_dir
+
+    IN THAT ORDER!
+    """
     training_id = str(uuid.uuid4())
 
-    training_output_dir = os.path.join(output_dir, training_id)
-    model_train_request = process_pb2.ProcessRequest(
-        trainingID=training_id,
-        customTrainingID=str(uuid.uuid4()),
-        request_dict={
-            "train_module": "00110203-0405-0607-0809-0a0b02dd0e0f",
-            "training_params": json.dumps(
-                {
-                    "model_name": "abc",
-                    "parameters": {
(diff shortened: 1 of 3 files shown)


joerunde (approved):
Sweet!

I'm fine with the API breaking change since it only affects users with access to the super-proprietary-and-not-open-sourced `model-train` component. 🤫

tests/runtime/servicers/test_model_train_servicer_impl.py
```diff
@@ -185,38 +185,66 @@ def test_model_train_validation_error_raises(sample_model_train_servicer, output
 
 #####################################################################
 # Normal tests
-def test_model_train_sample_widget(sample_model_train_servicer, output_dir):
+@pytest.mark.parametrize("output_in_config", (True, False))
```
joerunde, line 188: very slick tests 👍

tests/runtime/servicers/test_model_train_servicer_impl.py
```diff
@@ -185,38 +185,66 @@ def test_model_train_validation_error_raises(sample_model_train_servicer, output
...
+    output_dir,
+    output_in_config,
+):
+    """This test tests end-to-end training. It includes verifying that the model
+    is saved in the right place. The place where the model is saved comes from:
+
+    1. get_config().runtime.training.output_dir
+    2. request.training_output_dir
+
+    IN THAT ORDER!
```
joerunde, line 200: all the more reason that the `process_pb2` handler should move out of caikit/caikit!

gabe-l-hart: 💯

PR #482, opened 2023-09-19
:sparkles: load files without extensions

### LINKED ISSUE #480: Support explicit extension types in DataStreamSource

In the [File](https://github.com/caikit/caikit/blob/main/caikit/interfaces/common/data_model/stream_sources.py#L29) stream source message, the only field that is available is `filename`. In [DataStreamSource](https://github.com/caikit/caikit/blob/main/caikit/runtime/service_generation/data_stream_source.py#L171) the logic of how to parse the file is determined based on splitting the file's extension off.

Some clients will not have control over the file names themselves when calling `caikit`'s APIs, so the request is to allow an `extension` field (possibly `type` or `file_type`?) that will supersede the extension splitting logic, allowing users to explicitly tell `caikit` how to load the file.

* Caikit could fall back to "try them all" if it fails to parse an extension from the field

joerunde:

Closes #480 with the alternative implementation of "try everything"

This is naive as, but it works

**If applicable**:

### DIFF
--- a/caikit/runtime/service_generation/data_stream_source.py
+++ b/caikit/runtime/service_generation/data_stream_source.py
@@ -167,8 +167,11 @@ class DataStreamSourceBase(DataStream):
         """Create a data stream object by deducing file extension
         and reading the file accordingly"""
 
-        full_fname = cls._get_resolved_source_path(fname)
         _, extension = os.path.splitext(fname)
+        if not extension:
+            return cls._load_from_file_without_extension(fname)
+
+        full_fname = cls._get_resolved_source_path(fname)
         log.debug3("Pulling data stream from %s file [%s]", extension, full_fname)
 
         if not fname or not os.path.isfile(full_fname):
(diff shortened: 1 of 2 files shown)


caikit/runtime/service_generation/data_stream_source.py
```diff
@@ -187,6 +190,36 @@ def _create_data_stream_from_file(cls, fname: str) -> DataStream:
...
+                stream.peek()
+                return stream
+            except Exception as e:  # pylint: disable=broad-exception-caught
+                # Catch any exception: it's hard to know which all could be thrown by any of the
+                # formatters
+                log.debug3(
+                    "Failed to load file %s using data stream factory method %s: %s",
+                    full_fname,
+                    factory_method,
+                    e,
```
gabe-l-hart, line 216:
```suggestion
                    e,
                    exc_info=True,
```

gabe-l-hart (approved):
One tiny suggestion, but otherwise looks great!

Issue #477, opened 2023-09-18
HTTP Server mTLS Crash


gabe-l-hart:
## Describe the bug

```
(*) Traceback (most recent call last):
(*) File "/opt/miniconda/lib/python3.9/runpy.py", line 197, in _run_module_as_main
(*) return _run_code(code, main_globals, None,
(*) File "/opt/miniconda/lib/python3.9/runpy.py", line 87, in _run_code
(*) exec(code, run_globals)
(*) File "/opt/miniconda/lib/python3.9/site-packages/caikit/runtime/__main__.py", line 92, in <module>
(*) main()
(*) File "/opt/miniconda/lib/python3.9/site-packages/caikit/runtime/__main__.py", line 88, in main
(*) _http_server.start(blocking=True)  # make http always blocking
(*) File "/opt/miniconda/lib/python3.9/site-packages/caikit/runtime/http_server/http_server.py", line 199, in start
(*) self.server.run()
(*) File "/opt/miniconda/lib/python3.9/site-packages/uvicorn/server.py", line 61, in run
(*) return asyncio.run(self.serve(sockets=sockets))
(*) File "/opt/miniconda/lib/python3.9/asyncio/runners.py", line 44, in run
(*) return loop.run_until_complete(main)
(*) File "uvloop/loop.pyx", line 1517, in uvloop.loop.Loop.run_until_complete
(*) File "/opt/miniconda/lib/python3.9/site-packages/uvicorn/server.py", line 68, in serve
(*) config.load()
(*) File "/opt/miniconda/lib/python3.9/site-packages/uvicorn/config.py", line 430, in load
(*) self.ssl: Optional[ssl.SSLContext] = create_ssl_context(
(*) File "/opt/miniconda/lib/python3.9/site-packages/uvicorn/config.py", line 124, in create_ssl_context
(*) ctx.load_verify_locations(ca_certs)
(*) OSError: [Errno 36] File name too long
```

gabe-l-hart:
This is almost certainly an issue with when the client CA cert is passed by value rather than via a path to a string. Passing the tls CA by string is valid for the `grpc` server, so we should support this for HTTP as well.

PR #484, opened 2023-09-22
HTTP With mTLS Inline

gabe-l-hart:
The uvicorn server uses python's SslContext which requires tls files be on disk and not passed by value.

Closes: https://github.com/caikit/caikit/issues/477

This PR updates the logic for booting the `uvicorn` server so that if the configured TLS values are passed as inline values, they will be temporarily written to a file so that they are loaded correctly. This is needed because [uvicorn uses `ssl.SslContext`](https://github.com/encode/uvicorn/blob/master/uvicorn/config.py#L110) to manage is TLS connection and that loads values from disk only.

**If applicable**:

### DIFF
--- a/caikit/runtime/http_server/http_server.py
+++ b/caikit/runtime/http_server/http_server.py
@@ -17,12 +17,16 @@ The server is responsible for binding caikit workloads to a consistent REST/SSE
 API based on the task definitions available at boot.
 """
 # Standard
+from contextlib import contextmanager
+from dataclasses import dataclass
 from functools import partial
 from typing import Any, Dict, Iterable, Optional, Type, get_args
 import asyncio
 import json
+import os
 import re
 import ssl
+import tempfile
 import threading
 import time
 
@@ -94,6 +98,14 @@ OPTIONAL_INPUTS_KEY = "parameters"
 # Endpoint to use for health checks
 HEALTH_ENDPOINT = "/health"
 
+# Small dataclass for consolidating TLS files
+@dataclass
+class _TlsFiles:
+    server_cert: Optional[str] = None
+    server_key: Optional[str] = None
+    client_cert: Optional[str] = None
+
+
 ## RuntimeHTTPServer ###########################################################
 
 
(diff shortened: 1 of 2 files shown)


caikit/runtime/http_server/http_server.py
```diff
@@ -532,6 +551,50 @@ def _health_check() -> str:
...
+                for fname, content in non_file_opts.items():
+                    temp_fname = os.path.join(tls_dir, fname)
+                    updated_files[fname] = temp_fname
+                    with open(temp_fname, "w", encoding="utf-8") as handle:
+                        handle.write(content)
+                yield _TlsFiles(**updated_files)
+        except OSError as err:
+            log.error(
+                "<RUN80977064E>",
+                "Cannot create temporary TLS files. Either pass config as file paths or run with write permissions.",
```
joerunde, line 593: oof :(
Right, it's probably not a safe assumption that a container on openshift will have write access to `/tmp`...

There's not any chance that uvicorn kwargs like `tls_kwargs["ssl_keyfile"]` can be passed as open io streams, is there?

gabe-l-hart: Right, this is not great for exactly that reason (thus my error handling). The code in `uvicorn` is here: https://github.com/encode/uvicorn/blob/master/uvicorn/config.py#L121

It's all just raw `ssl` library, so we can poke around there, but I'm not optimistic.

joerunde: Yeah, just tried to pass it an `io.StringIO` and it barfs all the same:
```
TypeError: certfile should be a valid filesystem path
```

joerunde (approved):
🌶️🌶️ 🔒 !

Slick security upgrades

caikit/runtime/http_server/http_server.py
gabe-l-hart: Found it! https://github.com/python/cpython/blob/main/Modules/_ssl.c#L3884C61-L3884C61

Issue #433, opened 2023-08-23
Can't dump only http services


tharapalanivel:
## Describe the bug

`dump_services` doesn't create `out_dir` when only REST is enabled and fails to save generated interfaces.

```
  File ".../caikit/runtime/dump_services.py", line 78, in <module>
    dump_http_services(out_dir)
  File ".../caikit/runtime/dump_services.py", line 63, in dump_http_services
    with open(
FileNotFoundError: [Errno 2] No such file or directory: '<out_dir>/openapi.json'
```

PR #489, opened 2023-09-26
fix: dump http services, add tests

anhuong:
closes: #433 
Fixes bug where dumping http services failed if the output directory didn't already exist.

**If applicable**:

### DIFF
--- a/tests/runtime/test_dump_services.py
+++ b/tests/runtime/test_dump_services.py
@@ -21,8 +21,6 @@ import alog
 
 # Local
 from caikit.runtime.dump_services import dump_grpc_services, dump_http_services
-import caikit.interfaces.common
-import sample_lib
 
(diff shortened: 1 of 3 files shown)


tests/runtime/test_dump_services.py
```diff
@@ -0,0 +1,72 @@
...
+import shutil
+import tempfile
+
+# First Party
+import alog
+
+# Local
+from caikit.runtime.dump_services import dump_grpc_services, dump_http_services
+import caikit.interfaces.common
+import sample_lib
```
evaline-ju, line 25: Initially I thought this was an extra import but then remembered we do need library information - as discussed offline could we try to use the config for sample_lib info within the tests?

evaline-ju (approved):
LGTM!

PR #490, opened 2023-09-26
✨ Multipart/form-data support for Data Stream Sources

### LINKED ISSUE #483: Support form-data files

Users want to dump form-data to files and supply them as data stream sources in the training apis.

These files look like:
```
----------------------------855490206437000809074191
Content-Disposition: form-data; name=""; filename="my_training_data.json"
Content-Type: application/json

[
    {
        "input": "this is an example.",
        "output": "1"
    },
    {
        "input": "this is another example.",
        "output": "2"
    }
]

----------------------------855490206437000809074191--
```

We could support reading these file types in as `DataStreamSource`s, and using the appropriate parser based on the `Content-Type`.

For http APIs, it would be cool if we could supply real form data in POSTs. But for grpc, and for cases where users' systems have already been posted some form-data that they've serialized to file, it would be nice for caikit to support this file type natively instead of making users create new files with just the content. That might be infeasible anyway for extremely large payloads. 

## Additional context

joerunde:

Closes #483 

This adds multipart/form-data support for data streams.

This is importantly _not_ the same as supporting multipart/form-data uploads on any runtime endpoints. This only adds support for generating data streams from files that are already in multipart/form-data format.

This allows these files to be supplied in the runtime APIs as `File`, `ListOfFiles`, or `S3Files` references.

This PR allllllllllllso finally refactors the data stream tests to be in `pytest` format. We did this because it was getting very messy supporting the core/data_stream and runtime/data_stream_source tests, each with their own fixtures and setup. 
They now share the same set of fixtures from the top-level `conftest.py`

:gabe-happy:

**If applicable**:


caikit/core/data_model/streams/data_stream.py
```diff
@@ -38,13 +39,14 @@
 from ...augmentors import AugmentorBase
 from ...exceptions import error_handler
 from ...toolkit import fileio
+from ..json_dict import JsonDictValue
+from .multipart_decoder import is_multipart_file, stream_multipart_file
```
gabe-l-hart, line 43: NIT: I _think_ there's some dark corner of `toolkit` that has a bunch of `is_xxxxx` functions. Should `is_multipart_form` live there?

caikit/core/data_model/streams/data_stream.py
```diff
@@ -344,10 +360,23 @@ def from_header_csv(cls, filename: str, *args, **kwargs) -> "DataStream[Dict]":
     def _from_header_csv_generator(cls, filename, *csv_args, **csv_kwargs):
         # open the csv file (closure around `filename`)
         with open(filename, mode="r", encoding="utf8") as fh:
+            yield from cls._from_header_csv_buffer_generator(
```
gabe-l-hart, line 363: woooooah, I've never seen `yield from` before!

caikit/core/data_model/streams/data_stream.py
```diff
@@ -567,6 +597,62 @@ def _from_jsonl_collection_generator(cls, dirname):
         for data_item in DataStream.chain(data_stream_list).flatten():
             yield data_item
 
+    @classmethod
+    def from_multipart_file(cls, filename: str) -> "DataStream[JsonDictValue]":
+        """Loads up a DataStream from a multipart file
+
+            The data items returned in the data stream are determined by the
```
gabe-l-hart, line 604: NIT: Funny indentation? I guess this is what the other docstrings in this file look like which is weird

caikit/core/data_model/streams/data_stream.py
```diff
@@ -567,6 +597,62 @@ def _from_jsonl_collection_generator(cls, dirname):
...
+        stream_list = []
+        for part in stream_multipart_file(filename):
+            content_type = part.content_type
+            if "json" in content_type:
+                stream_list.append(
+                    cls(cls._from_json_array_buffer_generator, part.fp, part.filename)
+                )
+            elif "csv" in content_type:
+
+                class UtfEncodeIOWrapper(io.IOBase):
```
gabe-l-hart, line 626: I think we should pull this out. Having it here will redeclare the class on every step of the iteration which is pretty wasteful

caikit/core/data_model/streams/multipart_decoder.py
```diff
@@ -0,0 +1,120 @@
...
+    content_type: str
+    filename: str
+    fp: typing.IO[bytes]
+
+
+def is_multipart_file(file) -> bool:
+    """Returns true if the file appears to contain a multi-part form data request"""
+    with open(file, "r", encoding="utf-8") as fp:
+        # Read a small bit of the file
+        head: str = fp.read(50)
```
gabe-l-hart, line 76: Is this safe if the file is less than 50 bytes?

caikit/core/data_model/streams/multipart_decoder.py
```diff
@@ -0,0 +1,120 @@
...
+
+    # The beginning of the file content should start with "--"
+    return head.lstrip().startswith("--")
```
gabe-l-hart, line 79: Yuck! What about

```yaml
---
foo: 1
bar: 2
---
foo: 3
bar: 4
```

We currently don't support multi-yaml, but I wouldn't be _shocked_ if that comes down the pike

gabe-l-hart: Reference: https://github.ibm.com/watson-vision/ServerConfig/blob/d97cab43378c9031a4cce6d86ce7c9016ff2fa7a/src/trans_inst_image_base.cpp#L494

caikit/core/data_model/streams/multipart_decoder.py
```diff
@@ -0,0 +1,120 @@
...
+def _get_multipart_boundary(file) -> str:
+    """Returns the multipart boundary string by looking for it in the first line of the file with
+    content"""
+    with open(file, "r", encoding="utf-8") as fp:
+        line = ""
+        while not line:
+            line = fp.readline().lstrip()
+    error.value_check(
+        "",
+        line.startswith("--"),
```
gabe-l-hart, line 117: I think this is ok IF we can have a better validator to ensure that the file is in fact a multipart form

pyproject.toml
```diff
@@ -29,6 +29,7 @@ dependencies = [
     "semver>=2.13.0,<4.0",
     "six>=1.16.0,<2.0.0",
     "tqdm>=4.59.0,<5.0.0",
+    "werkzeug>=2.3.7,<3.0.0"
```
gabe-l-hart, line 32: 😞 but ok

tests/core/data_model/streams/test_data_stream.py
```diff
@@ -17,16 +17,31 @@
...
+###############################
+# Helper functions and fixtures
+###############################
+
+
+def validate_data_stream(data_stream, length, data_item_type, data_item_length=None):
+    # Verify lengths and types
+    assert isinstance(data_stream, DataStream)
+    assert len(data_stream) == length
+    assert sum(1 for _ in data_stream) == length
```
gabe-l-hart, line 38: I've never understood why we did this instead of `len(list(data_stream)) == length`

tests/core/data_model/streams/test_data_stream.py
```diff
@@ -72,571 +169,530 @@ def test_data_stream_from_jsonl_is_pickleable(tmp_path):
     assert pre_pickle_vals == post_pickle_vals
 
 
-class TestDataStream(TestCaseBase):
-    def setUp(self):
-        self.list_data_stream = core_dm.DataStream.from_iterable(
-            ["hello", "world", "!"]
+def test_bad_json_stream(tmp_path):
```
gabe-l-hart, line 172: We'll probably want this for a bad multipart too

tests/core/data_model/streams/test_data_stream.py
```diff
@@ -72,571 +169,530 @@ def test_data_stream_from_jsonl_is_pickleable(tmp_path):
...
+        DataStream.from_multipart_file(multipart_file)
 
-        def generator_func():
-            generator_func.it = 0
-            for i in range(10):
-                yield i
-                generator_func.it += 1
 
-        counter_stream = core_dm.DataStream(generator_func)
+###########################
```
gabe-l-hart, line 227: I'm just going to assume everything below here got better...

gabe-l-hart (changes requested):
A couple of little NITs and a request to move that stream adapter class to a place where it's only declared once.

The only other thing is thinking about whether we want a better `is_multipart_form` detector. According to [the spec](https://www.w3.org/TR/html401/interact/forms.html#h-17.13.4.2) I would expect the _actual_ first line of the file to look like `Content-Type: multipart/form-data; boundary=AaB03x` which you could then use to look for the boundary after the `--`. It looks like from your sample files, we may not expect this to be true though?

caikit/core/data_model/streams/data_stream.py
evaline-ju: There was a `isa` file of functions in a legacy version of this codebase but I don't see it (or those functions) existent in this codebase anymore.

caikit/core/data_model/streams/data_stream.py
evaline-ju: yeah I went for consistency here but could clean this up to make the PR a little bigger 😄

caikit/core/data_model/streams/multipart_decoder.py
evaline-ju: It'll just read the whole file if <50 bytes

caikit/core/data_model/streams/data_stream.py
joerunde: Neither had we!

tests/core/data_model/streams/test_data_stream.py
joerunde: We found out the hard way- it's to test re-entrance. The initial `len` call has to iterate over the entire stream, then the `sum` will fail if the stream isn't re-entrant

pyproject.toml
joerunde: yeah.... it was the lightest-weight (I think this only pulls in `MarkupSafe`) and most user-friendly thing I found while looking through some options

caikit/core/data_model/streams/multipart_decoder.py
joerunde: Yeah....
Started an internal convo with some users that are supplying files without the boundary header. We should probably also check for the initial content-type header though as well

tests/core/data_model/streams/test_data_stream.py
joerunde: Ah- but yeah we probably could do that with a `len(list(stream))` ? But that might hold the whole stream in memory and some nefarious person could put a 50GB stream in the unit tests 😉

caikit/core/data_model/streams/multipart_decoder.py
joerunde: Went with a `readline()` loop to make it a bit more robust

caikit/core/data_model/streams/multipart_decoder.py
joerunde: yeah, made that a bit more robust and removed the check here

tests/core/data_model/streams/test_data_stream.py
joerunde: added!

caikit/core/data_model/streams/multipart_decoder.py
```diff
@@ -0,0 +1,156 @@
...
+    fp: typing.IO[bytes]
+
+
+def is_multipart_file(file) -> bool:
+    """Returns true if the file appears to contain a multi-part form data request"""
+    log.debug3("Determining if %s is a multipart file", file)
+    first_line = _get_first_nonempty_line(file)
+
+    # Either: The beginning of the file starts with a boundary string (must start with --)
+    if first_line.startswith("--"):
```
gabe-l-hart, line 85: Ug. I would really want this to be more rigorous, but I just found this little chestnut in [the official RFC](https://www.w3.org/Protocols/rfc1341/7_2_Multipart.html):

```
The simplest boundary possible is something like "---", with a closing boundary of "-----".
```

tests/core/data_model/streams/test_data_stream.py
gabe-l-hart: Fair enough!

gabe-l-hart (approved):
Looks great! Ship it

PR #491, opened 2023-09-27
✅ try pytest xdist for parallelizing tests

prashantgupta24:

Seems to run the tests locally way faster!

Can't seem to find the right number to use

**If applicable**:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -67,6 +67,7 @@ dev-test = [
     "tls_test_tools>=0.1.1",
     "wheel>=0.38.4",
     "caikit[interfaces-vision]",
+    "pytest-xdist>=3.3.1,<4"
 ]
 
 dev-docs = [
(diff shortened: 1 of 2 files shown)


joerunde:
> Can't seem to find the right number to use

https://pytest-xdist.readthedocs.io/en/latest/distribution.html#:~:text=With%20%2Dn%20auto%20%2C%20pytest%2D,back%20to%20%2Dn%20auto%20behavior.

`-n auto` seems to be the trick. But as @gabe-l-hart mentioned this is only gonna work if we quit using global state.
I think the biggest problem there is the config- every test uses `with temp_config` and that modifies the global caikit config. There are some other things like the module registries and loaded model caches kicking around that could trip us up as well

The two opposite routes we talked about were either:
- Creating a thread-local `CaikitContext` that holds all state
- Using python `mocks` for global state during tests

Neither seemed like the best idea since using mocks for everything is pretty hard, and a thread local caikit context doesn't seem to have a whole lot of use outside of running unit tests. I think we had come up with something in the middle? But I forget... :(

prashantgupta24:
For some reason, using `auto` fails the tests on my local machine :/

gabe-l-hart:
@prashantgupta24 This PR seems to be stale. Closing now and can reopen down the line if needed

PR #497, opened 2023-09-28
:wrench: Add 9 as another OOM indicative exit code

gkumbhat:
### Changes
- Change `OOM_EXIT_CODE` to a list (and change variable name accordingly)
- Add 9, -9 to OOM exit code list (generally unsigned exit codes are expected but -9 has appeared before)

### DIFF
--- a/tests/runtime/servicers/test_global_train_servicer_impl.py
+++ b/tests/runtime/servicers/test_global_train_servicer_impl.py
@@ -11,8 +11,8 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+"""General global train servicer tests - run for both subprocess and local"""
 # Standard
-from contextlib import contextmanager
 from tempfile import TemporaryDirectory
 from unittest.mock import patch
 import multiprocessing
@@ -43,7 +43,7 @@ from sample_lib.data_model.sample import (
     SampleTrainingType,
 )
 from sample_lib.modules import CompositeModule, OtherModule, SampleModule
-from tests.conftest import random_test_id, reset_model_manager, temp_config
+from tests.conftest import random_test_id, set_use_subprocess
 from tests.fixtures import Fixtures
 from tests.runtime.conftest import register_trained_model
 import caikit.core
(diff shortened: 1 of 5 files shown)


tests/runtime/servicers/test_global_train_servicer_impl.py
```diff
@@ -403,6 +403,39 @@ def test_global_train_returns_exit_code_with_oom(
...
+        parameters=train_request_params_class(
+            batch_size=42,
+            training_data=training_data,
+            oom_exit=True,
+            oom_exit_code=9,
+        ),
+    ).to_proto()
+
+    # Enable sub-processing for test
+    with set_use_subprocess(True):
```
joerunde, line 427: This test and the one above it are already running within the autouse fixture `set_train_location` that runs the tests twice, once with `set_use_subprocess(True)` and once with `set_use_subprocess(False)`.

We should either find a way to disable that fixture for these two tests, or move these out to another test file that's specific to sub-process training only

tests/runtime/servicers/test_global_train_servicer_impl.py
```diff
@@ -403,6 +403,39 @@ def test_global_train_returns_exit_code_with_oom(
...
+    )
+    train_class = get_train_request(SampleModule)
+    train_request_params_class = get_train_params(SampleModule)
+    train_request = train_class(
+        model_name=random_test_id(),
+        parameters=train_request_params_class(
+            batch_size=42,
+            training_data=training_data,
+            oom_exit=True,
+            oom_exit_code=9,
```
joerunde, line 422: Also, I think we can parameterize the test above with the different exit codes instead of copy/paste.

```
@pytest.mark.parametrize("oom_code", [137, 9])
def test_whatever(oom_code):
    ...
```

joerunde (changes requested):
A couple little test improvements would be 🌶️

tests/runtime/servicers/test_global_train_servicer_impl.py
evaline-ju: sure - that was to separate out the default case of 137 but can be combined

joerunde (approved):
slick! Thanks for the test cleanup

PR #506, opened 2023-10-03
Add Support for Multi-Part Input and Binary File Output in HTTP Server

HonakerM:
This PR adds support for multi-part input and binary file output in the HTTP server. This allows users to submit requests via json or multi-part/form-data. The form-data is structured so that each key is the subpath where the value should live. For example the following form-data:
```
----
inputs.name
"test"
---
inputs.other
"test2"
---
parameters
'{"json":"field"}'
```
Would be parsed into the following structure:
```
{
   "inputs":{
       "name":"test",
       "other":"test2"
    },
    "parameters":{
         "json":"field"
    }
}
```
The Binary file output is controlled via two new optional methods on `DataObjectBase`: `to_file` and `from_file`. If these functions are implemented then the HTTP server will automatically return a bytes file using `to_file`. Currently this functionally is not configurable at runtime.

The parsing of form-data into pydantic is a little complex so please let me know if you have any questions or concerns. 

Another thing I'd like to note is that fastapi doesn't natively support multiple input content-types so I had to add some manual input parsing and openapi.json generation. 

**If applicable**:

### DIFF
--- a/caikit/interfaces/common/data_model/stream_sources.py
+++ b/caikit/interfaces/common/data_model/stream_sources.py
@@ -26,12 +26,12 @@ from caikit.core.data_model import PACKAGE_COMMON, DataObjectBase, dataobject
 
 
 @dataobject(PACKAGE_COMMON)
-class FileStream(DataObjectBase):
+class FileReference(DataObjectBase):
     filename: Annotated[str, FieldNumber(1)]
 
 
 @dataobject(PACKAGE_COMMON)
-class ListOfFileStreams(DataObjectBase):
+class ListOfFileReferences(DataObjectBase):
     files: Annotated[List[str], FieldNumber(1)]
 
 
--- a/caikit/interfaces/common/data_model/stream_sources.py
+++ b/caikit/interfaces/common/data_model/stream_sources.py
@@ -27,7 +27,7 @@ from caikit.core.data_model import PACKAGE_COMMON, DataObjectBase, dataobject
 
 @dataobject(PACKAGE_COMMON)
 class File(DataObjectBase):
-    filename: Annotated[str, FieldNumber(1)]
+    filename: Annotated[Optional[str], FieldNumber(1)]
     data: Annotated[Optional[bytes], FieldNumber(2)]
     type: Annotated[Optional[str], FieldNumber(3)]
 
--- a/caikit/interfaces/common/data_model/stream_sources.py
+++ b/caikit/interfaces/common/data_model/stream_sources.py
@@ -16,7 +16,7 @@ This file contains interfaces required to generate DataStreamSource[T] classes
 """
 
 # Standard
-from typing import List, Optional
+from typing import List
 
 # First Party
 from py_to_proto.dataclass_to_proto import Annotated, FieldNumber
@@ -26,14 +26,12 @@ from caikit.core.data_model import PACKAGE_COMMON, DataObjectBase, dataobject
 
 
 @dataobject(PACKAGE_COMMON)
-class File(DataObjectBase):
-    filename: Annotated[Optional[str], FieldNumber(1)]
-    data: Annotated[Optional[bytes], FieldNumber(2)]
-    type: Annotated[Optional[str], FieldNumber(3)]
+class FileStream(DataObjectBase):
+    filename: Annotated[str, FieldNumber(1)]
 
 
 @dataobject(PACKAGE_COMMON)
-class ListOfFiles(DataObjectBase):
+class ListOfFileStreams(DataObjectBase):
     files: Annotated[List[str], FieldNumber(1)]
 
 
--- a/caikit/interfaces/common/data_model/stream_sources.py
+++ b/caikit/interfaces/common/data_model/stream_sources.py
@@ -16,7 +16,7 @@ This file contains interfaces required to generate DataStreamSource[T] classes
 """
 
 # Standard
-from typing import List
+from typing import List, Optional
 
 # First Party
 from py_to_proto.dataclass_to_proto import Annotated, FieldNumber
@@ -28,6 +28,8 @@ from caikit.core.data_model import PACKAGE_COMMON, DataObjectBase, dataobject
 @dataobject(PACKAGE_COMMON)
 class File(DataObjectBase):
     filename: Annotated[str, FieldNumber(1)]
+    data: Annotated[Optional[bytes], FieldNumber(2)]
+    type: Annotated[Optional[str], FieldNumber(3)]
 
 
 @dataobject(PACKAGE_COMMON)
--- a/caikit/runtime/http_server/http_server.py
+++ b/caikit/runtime/http_server/http_server.py
@@ -32,7 +32,7 @@ import threading
 import time
 
 # Third Party
-from fastapi import FastAPI, Request, Response, status
+from fastapi import FastAPI, HTTPException, Request, Response, status
 from fastapi.encoders import jsonable_encoder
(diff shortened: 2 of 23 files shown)


caikit/runtime/http_server/http_server.py
```diff
@@ -546,6 +590,64 @@ def _get_response_dataobject(rpc: CaikitRPCBase) -> Type[DataBase]:
...
+        pydantic_model: Union[pydantic.BaseModel, Type[pydantic.BaseModel]]
+    ):
+        """Helper to generate the openapi schema for a given request"""
+        raw_schema = pydantic_model.model_json_schema()
+        parsed_schema = flatten_json_schema(raw_schema)
+
+        multipart_schema = convert_json_schema_to_multipart(parsed_schema)
+
+        return {
+            "requestBody": {
```
gabe-l-hart, line 622: Cool! This is a _way_ better solution than the brutal hack I'd come up with a while ago

caikit/runtime/http_server/pydantic_wrapper.py
```diff
@@ -170,3 +181,182 @@ def _from_base64(data: Union[bytes, str]) -> bytes:
...
+
+async def pydantic_from_request(
+    pydantic_model: Type[pydantic.BaseModel], request: Request
+):
+    """Function to convert a fastapi request into a given pydantic model. This
+    function parses the requests Content-Type and then correctly decodes the data.
+    The currently supported Content-Types are `application/json`
+    and `multipart/form-data`"""
+    content_type = request.headers.get("Content-Type")
+    log.debug4("Detected request using %s type", content_type)
```
gabe-l-hart, line 194: NIT: We probably want this on a more visible level since I suspect this could be very useful and will only log once per request.

```suggestion
    log.debug("Detected request using %s type", content_type)
```

caikit/runtime/http_server/pydantic_wrapper.py
```diff
@@ -170,3 +181,182 @@ def _from_base64(data: Union[bytes, str]) -> bytes:
...
+
+    # If content type is json use pydantic to parse
+    if content_type == "application/json":
+        raw_content = await request.body()
+        try:
+            return pydantic_model.model_validate_json(raw_content)
+        except pydantic.ValidationError as err:
+            raise RequestValidationError(  # pylint: disable=raise-missing-from
```
gabe-l-hart, line 202: Any reason not to do `from err`?

gabe-l-hart (commented):
Partial review comments. Dumping for now and will pick up when I can.

caikit/interfaces/common/data_model/stream_sources.py
```diff
@@ -27,7 +27,9 @@
 
 @dataobject(PACKAGE_COMMON)
 class File(DataObjectBase):
-    filename: Annotated[str, FieldNumber(1)]
+    filename: Annotated[Optional[str], FieldNumber(1)]
+    data: Annotated[Optional[bytes], FieldNumber(2)]
```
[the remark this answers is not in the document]
HonakerM: > Alternately, we can/should use a different object.

I was thinking this as well when I started working on this. Like a generic `File` class in `caikit.interfaces.common` but I didn't want to move this class so I just added fields to it. If this class is only used in StreamSource maybe we could rename this class to `FileStream`?

tests/fixtures/sample_lib/data_model/sample.py
```diff
@@ -19,6 +22,22 @@ class SampleInputType(DataObjectBase):
     name: str
 
 
+@dataobject(package="caikit_data_model.sample_lib")
+class SampleListInputType(DataObjectBase):
+    """A sample `domain primitive` input type for this library.
```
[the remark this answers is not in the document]
HonakerM: Okay! I wasn't sure so I just copied it from the out SampleInputType 😅.  I'll update both classes descriptor

caikit/runtime/http_server/pydantic_wrapper.py
HonakerM: Good call! I wasn't sure what the level scheme was for caikit so I just did the lowest I found.

caikit/runtime/http_server/pydantic_wrapper.py
HonakerM: I wasn't sure how fastapi's exception response would handle chained errors. I'll test it and add the chain if nothing breaks

HonakerM:
One issue I just noticed with this PR is that the order of the multi-part fields matter. For example if you do the following:
```
inputs: {some:"data"}
inputs.file: raw_data
```
The code correctly parses it to `{inputs:{file: raw_data, some: "data"}` however, if you reverse the keys e.g.
```
inputs.file: raw_data
inputs: {some:"data"}
```
The `inputs.file` data gets overridden by `inputs`. This can be solved by merging the dicts instead of overriding them which I should have done sometime this week

caikit/interfaces/common/data_model/stream_sources.py
```diff
@@ -26,12 +26,12 @@
 
 
 @dataobject(PACKAGE_COMMON)
-class File(DataObjectBase):
+class FileStream(DataObjectBase):
```
joerunde, line 29: iiinteresting. I know this is only used by `DataStreamSource`s but the name `FileStream` seems a bit confusing in a vacuum here. Since the goal here is to reference a file that a runtime service can open, should this be named `FileReference`? `LocalFile`? `MountedFile`? `FileMount`?

Since the name of this class ends up as the field name in the api as well, I think it'd be good to make sure that whatever we pick here conveys the meaning well enough to somebody who's just looking at the api spec and trying to make things work

caikit/runtime/service_generation/data_stream_source.py
```diff
@@ -116,14 +116,14 @@ def to_data_stream(self) -> DataStream:
             return DataStream.from_iterable(self.jsondata.data)
 
         # If jsonfile, attempt to read the file and pull in the data from there
-        if set_field == "file":
-            return self._create_data_stream_from_file(fname=self.file.filename)
+        if set_field == "filestream":
```
joerunde, line 119: While we're in this class, do you think it would be possible to implement at least `from_file` so that for data stream parameters, users can send in like:
```
inputs.my_data_stream_arg: raw_data
``

caikit/interfaces/common/data_model/stream_sources.py
HonakerM: I also didn't like `FileStream` all that well. Since this class just contains a string pointing to a file I think `FileReference` makes the most sense. Does that work?

caikit/runtime/service_generation/data_stream_source.py
HonakerM: @joerunde I'm not quite sure what you mean. Like update `_create_data_stream_from_file` to support both filename and raw bytes?

HonakerM:
@joerunde I started to work on adding support for raw File data but it started to get alot more complicated  than I expected. This was also affected by the recent switch to `DataStreamSourcePlugin` which I'm still coming up to speed on.

I opened a new issue to track that work: #529. Could this PR be merged in the meantime? We're still using a forked version which I'd like to avoid.

joerunde (approved):
did the ol' double-skim after @gabe-l-hart 's review. Looks pretty good to me!

Note to others: we went ahead and updated the `listoffiles` field name to `list_of_files`. We didn't expect our internal users to be using that field yet and it seemed pretty awkward to have the name smashed together.

PR #507, opened 2023-10-03
:sparkles: set timers to daemon

joerunde:

Just looking around at timers for other reasons and I noticed that the lazy load poll timer is not set as a daemon. I think we'd want this set so that we don't have a child thread here interfering with shutdown.

**If applicable**:


gabe-l-hart (approved):
Good catch!

PR #508, opened 2023-10-03
:goal_net: Return error for non-array JSON where array is expected

### LINKED ISSUE #500: Reject non array .json files for data streams
## Describe the bug

If a user uploads a `.json` file for a data stream parameter in a runtime api, it is expected to be an array of the data stream elements. If a user instead supplies a file with contents like:
```
{
  "input": "foo",
  "output": "bar"
}
```
caikit will happily yield an empty stream. This gives no indication to the user about what they did wrong.

## Additional context

Make sure `.jsonl` is still supported correctly

evaline-ju:

Closes #500 - previously users would just receive an empty stream for non-array JSON provided in JSON files. This PR provides users an error message when an array is expected

Includes small duplicate log code update

**If applicable**:

### DIFF
--- a/caikit/core/data_model/streams/data_stream.py
+++ b/caikit/core/data_model/streams/data_stream.py
@@ -254,9 +254,16 @@ class DataStream(Generic[T]):
             )
         # For each {} object of the array
         try:
+            item_idx = None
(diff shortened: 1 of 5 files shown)


gabe-l-hart (approved):
Thanks!

Issue #501, opened 2023-09-29
Caikit fails to propagate exception with ForkingPickler error


evaline-ju:
## Describe the bug

When an exception is thrown by `caikit_nlp` inside a runtime, the exception fails to propagate. Example:

Runtime response when a bad file format (e.g. multipart form with an extension) is given: 
```
{
  "training_id": "86956cdb-e557-47d7-9436-042452e7d253",
  "state": "ERRORED",
  "submission_timestamp": "2023-09-29T17:32:13.260Z",
  "completion_timestamp": "2023-09-29T17:32:28.784Z",
  "reasons": [
    "Exception raised during training. This may be a problem with your input: __init__() missing 1 required positional argument: 'message'"
  ]
}
```

The proper stack trace from the internal training:
```
Traceback (most recent call last):
  File \"/opt/miniconda/lib/python3.9/site-packages/caikit/core/toolkit/destroyable_process.py\", line 163, in run
    super().run()
  File \"/opt/miniconda/lib/python3.9/multiprocessing/process.py\", line 108, in run
    self._target(*self._args, **self._kwargs)
  File \"/opt/miniconda/lib/python3.9/site-packages/caikit/core/toolkit/destroyable_process.py\", line 209, in _target_wrapper
    result = target(*(args or []), **(kwargs or {}))
  File \"/opt/miniconda/lib/python3.9/site-packages/caikit/core/model_management/local_model_trainer.py\", line 222, in _train_and_save
    trained_model = self._module_class.train(*args, **kwargs)
  File \"/opt/miniconda/lib/python3.9/site-packages/caikit_nlp/modules/text_generation/peft_prompt_tuning.py\", line 376, in train
    train_dataloader, val_dataloader = cls.create_dataloaders_from_stream(
  File \"/opt/miniconda/lib/python3.9/site-packages/caikit_nlp/modules/text_generation/peft_prompt_tuning.py\", line 744, in create_dataloaders_from_stream
    train_dataloader = cls._get_data_loaders_from_stream(
  File \"/opt/miniconda/lib/python3.9/site-packages/caikit_nlp/modules/text_generation/peft_prompt_tuning.py\", line 948, in _get_data_loaders_from_stream
    wrapped_stream = SimpleIterableStreamWrapper(mapped_stream, shuffle=shuffle)
  File \"/opt/miniconda/lib/python3.9/site-packages/caikit_nlp/toolkit/data_stream_wrapper.py\", line 46, in __init__
[...]

PR #509, opened 2023-10-03
🐛 Propagate sub-process errors properly

prashantgupta24:

fix https://github.com/caikit/caikit/issues/501

There were 2 issues identified:
1. The tests in `test_model_train_servicer_impl.py` were not using `sub-process` correctly  since the `get_trainer` function always fetches a trainer that was initialized before our custom config (with `sub_process` set to `True`) comes into play. Hence we’re never actually checking a training in a `sub_process`. I verified this by running a failing test and seeing that we always hit `destroyable_thread.py` in the stacktrace instead of `destroyable_process` in both scenarios.
2. `CaikitRuntimeException` was not pickleable, which is a requirement for any object to be passed through sub-processes using `Pipe()`. 

**If applicable**:

### DIFF
--- a/tests/runtime/servicers/test_model_train_servicer_impl.py
+++ b/tests/runtime/servicers/test_model_train_servicer_impl.py
@@ -29,7 +29,7 @@ from caikit.core.data_model import TrainingStatus
 from caikit.runtime.protobufs import process_pb2
 from caikit.runtime.servicers.model_train_servicer import ModelTrainServicerImpl
 from caikit.runtime.types.caikit_runtime_exception import CaikitRuntimeException
-from tests.conftest import temp_config
+from tests.conftest import set_use_subprocess
 from tests.fixtures import Fixtures
 import sample_lib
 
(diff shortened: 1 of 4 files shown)


tests/runtime/servicers/test_model_train_servicer_impl.py
```diff
@@ -58,11 +58,11 @@ def clear_messages_from_servicer(servicer):
 
 
 @pytest.fixture(autouse=True, params=[True, False])
-def set_train_location(request):
+def set_train_location(request, reset_model_manager):
```
prashantgupta24, line 61: Even though we had this section here to parameterize `use_subprocess`, it doesn’t actually work since the `get_trainer` function always fetches a trainer that was initialized before this config comes into play. Hence we’re never actually checking a training in a `sub_process`. I verified this by running a failing test and seeing that we always hit `destroyable_thread.py` in the stacktrace instead of `destroyable_process` in both scenarios.

To fix this, we need to add `reset_model_manager` to make sure all trainers use the config that we provide.

joerunde: @prashantgupta24 can you add that context directly to the docstring for this fixture?

joerunde (approved):
Looks good!

Would still be super nice to get that context into the docstring for the test fixture

Issue #246, opened 2023-06-16
Support multi-task modules


gabe-l-hart:
## Is your feature request related to a problem? Please describe.

With the introduction of [Streaming Tasks](https://github.com/caikit/caikit/pull/208), we've opened the door to having multiple `tasks` that are implemented by the same model under the hood, but different input/output data wrangling. This may also come up in the case where base text generation models can be used to accomplish other tasks (e.g. `classification`, `summarization`) using different pre/post processing.

In the current implementation, every loaded `model` supports at most one `task` which is defined in the `@module` decorator on the `module` for the `model`. This means that in a given instance of `caikit.runtime`, in order to use a given logical `model` for multiple tasks, it must be re-loaded for each task and given a different in-memory `model_id`. This is potentially very inefficient and also confusing.

## Describe the solution you'd like

The cleanest solution, in my mind, is to allow the `@module` decorator to optionally take a mapping from `task` to function name to allow the inference service binding to run functions besides `run` as the inference function for multiple different tasks.

## Describe alternatives you've considered

We could eliminate the efficiency problem problem by using a "resource" module that holds the in-memory base model and loads as a singleton and having multiple "workflow" modules that take that by reference and use it. This would still have separate `model_id` values for the different `tasks`, but would avoid the inefficiency of loading the real model multiple times.

## Additional context

* A decision on whether to implement this will have an impact on how the REST API is organized.
* If we adopt this, we'd need to expand the RPC deduction logic to handle multiple `tasks` for a given `module`

gabe-l-hart:
Slack conversation on the topic: https://ibm-research.slack.com/archives/C0536GM5F2Q/p1694251032011869

matthewpwilson:
+1 for this requirement. My use-case is for a multi-modal model that supports inference with multiple data types.

I wonder if `@taskmethod` could be sufficient to indicate a module supports multiple tasks? The task parameter on `@module` could then be made option.  For a multi-modal model that supports text -> image or image -> text It could look something like this:

```
@task(
    required_parameters={"text_input": str},
    output_type=ImageOutput,
)
class TextToImage(TaskBase):
    pass

@task(
    required_parameters={"image": ImageInput},
    output_type=str,
)
class ImageToText(TaskBase):
    pass

@module(
    "00110203-0405-0607-0809-0a0b02ee0e0a",
    "TextImageModule",
    "0.0.1",
)
class TextImageModule(MultiTaskModuleBase)

@TextToImage.taskmethod()
[...]

PR #511, opened 2023-10-05
Suport modules with multiple tasks

matthewpwilson:
refs #246 

Today Caikit allows only a single task to be associated with a module. However, some ML models may support multiple tasks. For example, a multi-modal model may support inference tasks for each data modality. The only way to achieve this currently is to create two modules, and a model that refers to each of them. This means the model weights may be loaded twice, wasting memory. 

This PR implements multi-task support as follows:

- A new optional `tasks` parameter on `@module` allows you to specify a list of tasks. This takes precedence over `task`.
- Inference methods for each task are annotated with `@TaskClass.taskmethod()`.
- As a consequence of this, some of the task validation is relaxed as it's now valid to have multiple tasks specified in the class hierarchy of your module. 
- Services are generated for each task.
- At inference time, the global predict servicer looks up the correct service based on the task.
 
There are a few other possible considerations I haven't addressed, happy to investigate if we think they need further thought:
- It's now possible to have a module without a generic `run()` implementation.  Should ModuleBase (or maybe a new `MultiTaskModuleBase`) provide a default implementation of `run()` that attempts to delegate to one of the `taskmethod`s baesd on the parameters? Should we require a multi-task module still implement a `run()`?
- What does training mean for multi-task modules? 
- I've introduced `module.TASK_CLASSES` which contains all the tasks implemented by the module. I've kept `TASK_CLASS` for compatibility, but is this necessary? 
- Any other impacts I've missed?

**If applicable**:

### DIFF
--- a/caikit/runtime/http_server/http_server.py
+++ b/caikit/runtime/http_server/http_server.py
@@ -346,7 +346,7 @@ class RuntimeHTTPServer(RuntimeServerBase):
                 log.error("<RUN51881106E>", err, exc_info=True)
             return Response(content=json.dumps(error_content), status_code=error_code)
 
-    def _add_unary_input_unary_output_handler(self, rpc: CaikitRPCBase):
+    def _add_unary_input_unary_output_handler(self, rpc: TaskPredictRPC):
         """Add a unary:unary request handler for this RPC signature"""
         pydantic_request = dataobject_to_pydantic(self._get_request_dataobject(rpc))
         pydantic_response = dataobject_to_pydantic(self._get_response_dataobject(rpc))
@@ -373,7 +373,7 @@ class RuntimeHTTPServer(RuntimeServerBase):
                     model_id=model_id,
                     request_name=rpc.request.name,
                     inference_func_name=model.get_inference_signature(
-                        output_streaming=False, input_streaming=False
+                        output_streaming=False, input_streaming=False, task=rpc.task
                     ).method_name,
                     **request_params,
                 )
--- a/caikit/runtime/service_factory.py
+++ b/caikit/runtime/service_factory.py
@@ -222,21 +222,18 @@ class ServicePackageFactory:
         for ck_module in modules:
             # Only create for modules from defined included and exclusion list
 
-            if not ck_module.TASK_CLASS:
+            if not ck_module.tasks:
                 log.debug(
-                    "Skipping module %s with no task",
+                    "Skipping module %s with no tasks",
                     ck_module,
                 )
                 continue
-
-            if (
-                excluded_task_types
-                and ck_module.TASK_CLASS.__name__ in excluded_task_types
-            ):
+            task = next(iter(ck_module.tasks))
+            if excluded_task_types and task.__name__ in excluded_task_types:
                 log.debug(
                     "Skipping module %s with excluded task %s",
                     ck_module,
-                    ck_module.TASK_CLASS.__name__,
+                    task.__name__,
                 )
                 continue
 
@@ -257,8 +254,7 @@ class ServicePackageFactory:
             # if inclusion is specified, use that
             else:
                 if (included_modules and ck_module.MODULE_ID in included_modules) or (
-                    included_task_types
-                    and ck_module.TASK_CLASS.__name__ in included_task_types
+                    included_task_types and task.__name__ in included_task_types
                 ):
                     clean_modules.add(ck_module)
 
@@ -285,7 +281,7 @@ def get_inference_request(
         TaskBase,
     )
     task_class = (
-        task_or_module_class.TASK_CLASS
+        next(iter(task_or_module_class.tasks))
         if issubclass(task_or_module_class, ModuleBase)
         else task_or_module_class
     )
@@ -311,9 +307,8 @@ def get_train_request(module_class: Type[ModuleBase]) -> Type[DataBase]:
         module_class,
         ModuleBase,
     )
-    request_class_name = (
-        f"{module_class.TASK_CLASS.__name__}{module_class.__name__}TrainRequest"
-    )
+    first_task = next(iter(module_class.tasks))
+    request_class_name = f"{first_task.__name__}{module_class.__name__}TrainRequest"
     log.debug("Request class name %s for module %s.", request_class_name, module_class)
     return DataBase.get_class_for_name(request_class_name)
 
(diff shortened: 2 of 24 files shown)


caikit/runtime/http_server/http_server.py
```diff
@@ -368,12 +368,12 @@ async def _handler(
                 )
 
                 # TODO: use `async_wrap_*`?
-                call = partial(
+                call = partial(q
```
joerunde, line 371: q!!!!

matthewpwilson: 🤦 fixed!

joerunde:
Thanks @matthewpwilson! This looks super great and is exactly where we want to take this moving forward.

I'll address some of your other considerations here:

> It's now possible to have a module without a generic run() implementation ...

I actually thought this was already the case, since you can define a module with a single `@Task.taskmethod` that's not named run. I could be mistaken about that, but regardless I don't think that's a huge deal and I don't think we need to require that modules override a `run()` method

> What does training mean for multi-task modules?

We discussed this with many of the caikit maintainers, we think we'd probably leave it up to the module authors to decide if a single `train` call always trains a model that can do all tasks, or if it may train a model that can only do some subset of the tasks that the module supports. This has some implications of how the trained models can be used, we figured the lightest-weight way to deal with this right now is to add a generic `TaskUnsupportedException` that a module's inference methods may raise if the loaded model cannot perform that task. A future improvement might be to add a `model.supported_tasks() -> Set[Task]` api that each module can optionally implement so that a caikit runtime can accurately report which loaded models can be used for which task apis, but that certainly doesn't have to happen here.

> I've introduced module.TASK_CLASSES which contains all the tasks implemented by the module. I've kept TASK_CLASS for compatibility, but is this necessary?

Right, it probably isn't. It'd be great if we can refactor this down to a list or set of tasks held internally on the module class.
There are also quite a few places where I see `module.TASK_CLASS`  referenced outside of decorator.py, so this would be a good forcing function to stick a proper `@property` on modules to expose a set of task classes (and maybe inference / training signatures too?). Would you mind doing that bit of refactoring and cleanup on this PR?

caikit/runtime/service_generation/create_service.py
```diff
@@ -115,15 +115,15 @@ def _group_modules_by_task(
...
+        for task_class in ck_module.TASK_CLASSES:
+            ck_module_task_name = task_class.__name__
             if ck_module_task_name is not None:
                 for (
                     input_streaming,
                     output_streaming,
                     signature,
-                ) in ck_module._INFERENCE_SIGNATURES:
-                    task_groups.setdefault(ck_module.TASK_CLASS, {}).setdefault(
+                ) in ck_module._TASK_INFERENCE_SIGNATURES[task_class]:
```
joerunde, line 125: oof, yeah this was always a kludge here to begin with. I'd much rather be accessing public properties of the module class when we're outside of the `caikit.core.modules` package.

joerunde: I believe in python you can set class-level `@property`s by sticking them on the meta-class. So something like this in `caikit/core/modules/meta.py` should work:

```
class _ModuleBaseMeta(abc.ABCMeta):
    ...

    @property
    def tasks(cls) -> Set[TaskBase]:
        return cls._TASK_CLASSES
```

joerunde (changes requested):
Looks fantastic!

If you can get that refactor in to just hold `TASK_CLASSES` and cleanup the usage outside the core/modules package this should be good to merge

caikit/runtime/service_generation/create_service.py
matthewpwilson: I've implemented this property and also `ModuleBase.get_inference_signatures` to replace this use of `_TASK_INFERENCE_SIGNATURES`.

matthewpwilson:
> If you can get that refactor in to just hold `TASK_CLASSES` and cleanup the usage outside the core/modules package this should be good to merge

Getting rid of references to `TASK_CLASS` threw up a few other places that needed to consdier multiple tasks in a module.  In particular I had to refactor the handling of included/excluded tasks for service generation quite a bit, please let me know what you think.

matthewpwilson:
A new test was introduced that required changes due to the removal of TASK_CLASS, I've fixed it.

caikit/runtime/service_factory.py
```diff
@@ -311,9 +286,8 @@ def get_train_request(module_class: Type[ModuleBase]) -> Type[DataBase]:
         module_class,
         ModuleBase,
     )
-    request_class_name = (
-        f"{module_class.TASK_CLASS.__name__}{module_class.__name__}TrainRequest"
-    )
+    first_task = next(iter(module_class.tasks))
+    request_class_name = f"{first_task.__name__}{module_class.__name__}TrainRequest"
```
joerunde, line 290: Ah, I think I skimmed past this on first review and now I understand your question in the PR body better. I don't have any great ideas on how to name the training rpc now that a module may have multiple tasks. We'll probably need to address that soon-ish, but this is fine to go with for now.

This also highlights the extra coupling between this logic and the rpc implementation- any change in the naming scheme over in `rpc.py` has to be reflected here.
There's probably a better way to get at the reference to the `ModuleClassTrainRPC` reference and return the request class from that, but that doesn't need to be fixed in this PR.

caikit/runtime/service_generation/rpcs.py
```diff
@@ -153,7 +153,7 @@ def module_class_to_rpc_name(module_class: Type[ModuleBase]) -> str:
         request RPC function
         """
         return snake_to_upper_camel(
-            f"{module_class.TASK_CLASS.__name__}_{module_class.__name__}_Train"
+            f"{next(iter(module_class.tasks)).__name__}_{module_class.__name__}_Train"
```
joerunde, line 156: 🌶️

caikit/runtime/servicers/global_predict_servicer.py
```diff
@@ -349,7 +350,7 @@ def _verify_model_task(self, model: ModuleBase):
         """Raise if the model is not supported for the task"""
         rpc_set: Set[TaskPredictRPC] = set(self._inference_service.caikit_rpcs.values())
         module_rpc: TaskPredictRPC = next(
-            (rpc for rpc in rpc_set if model.TASK_CLASS == rpc.task),
+            (rpc for rpc in rpc_set if rpc.task in model.__class__.tasks),
```
joerunde, line 353: very pythonic 👍

joerunde (approved):
Ship it!

We'll follow up with some design discussion about the training API but we don't want to hold this change up on it

markstur:
WOO HOO!  :tada:   Thanks!

PR #514, opened 2023-10-09
Add rpc error handler

gkumbhat:
closes: https://github.com/caikit/caikit/issues/515

### Changes
- Add support for handling grpc errors thrown by underlying backends.

### DIFF
--- a/tests/fixtures/sample_lib/modules/sample_task/list_implementation.py
+++ b/tests/fixtures/sample_lib/modules/sample_task/list_implementation.py
@@ -34,10 +34,7 @@ class ListModule(caikit.core.ModuleBase):
         return cls(config["train"]["batch_size"], config["train"]["learning_rate"])
 
     def run(
-        self,
-        sample_input: SampleInputType,
-        throw: bool = False,
-        error_obj: Optional[Exception] = None,
+        self, sample_input: SampleInputType, throw: bool = False
     ) -> SampleOutputType:
         """
         Args:
--- a/tests/fixtures/sample_lib/modules/sample_task/list_implementation.py
+++ b/tests/fixtures/sample_lib/modules/sample_task/list_implementation.py
@@ -34,7 +34,10 @@ class ListModule(caikit.core.ModuleBase):
         return cls(config["train"]["batch_size"], config["train"]["learning_rate"])
 
     def run(
-        self, sample_input: SampleInputType, throw: bool = False
+        self,
+        sample_input: SampleInputType,
+        throw: bool = False,
+        error_obj: Optional[Exception] = None,
     ) -> SampleOutputType:
         """
         Args:
--- a/tests/runtime/servicers/test_global_predict_servicer_impl.py
+++ b/tests/runtime/servicers/test_global_predict_servicer_impl.py
@@ -68,6 +68,31 @@ def test_calling_predict_should_raise_if_module_raises(
     assert "Unhandled exception during prediction" in context.value.message
 
 
+def test_predict_raises_with_grpc_errors(
+    sample_inference_service,
+    sample_predict_servicer,
+    sample_task_model_id,
+    sample_task_unary_rpc,
+):
+    with pytest.raises(CaikitRuntimeException) as context:
+        # SampleModules will raise a RuntimeError if the throw flag is set
+        predict_class = get_inference_request(SampleModule.TASK_CLASS)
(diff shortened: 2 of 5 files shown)


tests/runtime/servicers/test_global_predict_servicer_impl.py
```diff
@@ -68,6 +68,31 @@ def test_calling_predict_should_raise_if_module_raises(
...
+def test_predict_raises_with_grpc_errors(
+    sample_inference_service,
+    sample_predict_servicer,
+    sample_task_model_id,
+    sample_task_unary_rpc,
+):
+    with pytest.raises(CaikitRuntimeException) as context:
+        # SampleModules will raise a RuntimeError if the throw flag is set
+        predict_class = get_inference_request(SampleModule.TASK_CLASS)
+        print(SampleModule.TASK_CLASS)
```
evaline-ju, line 80:
```suggestion
```

tests/fixtures/sample_lib/modules/sample_task/list_implementation.py
```diff
@@ -34,7 +34,10 @@ def load(cls, model_path, **kwargs):
         return cls(config["train"]["batch_size"], config["train"]["learning_rate"])
 
     def run(
-        self, sample_input: SampleInputType, throw: bool = False
+        self,
+        sample_input: SampleInputType,
+        throw: bool = False,
+        error_obj: Optional[Exception] = None,
```
evaline-ju, line 40: unused? but if intended to allow for future use maybe it'd be helpful to keep consistent as either exception or str between `sample_implementation` and this?

joerunde, line 40: Is this used in any tests?

tests/runtime/servicers/test_global_predict_servicer_impl.py
```diff
@@ -68,6 +68,31 @@ def test_calling_predict_should_raise_if_module_raises(
...
+
+        request = predict_class(
+            sample_input=HAPPY_PATH_INPUT_DM,
+            throw=True,
+            error="GRPC_RESOURCE_EXHAUSTED",
```
joerunde, line 85: There are a lot of things I don't like about mocks but... sometimes I wish we had just used mocks for these 🙃 

It would make it a bit easier to say like
```
with patch(SampleModule, "run") as mock_run:
    mock_run.side_effect = grpc.RpcError(grpc.StatusCode.UNIMPLEMENTED, "some error message")
```
and be able to test out any exception type and content

Not super necessary to change but that might make our tests a little easier to read as well

tests/fixtures/sample_lib/modules/sample_task/list_implementation.py
gkumbhat: accidentally pushed! Removing

tests/runtime/servicers/test_global_predict_servicer_impl.py
joerunde: That said- if we have the eventual goal of making tests parallel I don't actually know if patching like that is thread safe. If it's not, then maybe we should keep as is 🤔

tests/fixtures/sample_lib/modules/sample_task/list_implementation.py
gkumbhat: Accidentally pushed! Removed

tests/runtime/servicers/test_global_predict_servicer_impl.py
gkumbhat: hmm, good question. Not sure how patching works with parallel tests and specially when there is threading etc that can happen as a result of some of these tests 🤔

gkumbhat: internet seems to suggest its not: https://stackoverflow.com/questions/26876438/does-pythons-unittest-mock-patch-mutate-global-state#:~:text=11-,mock.,assignment%20statement%20at%20the%20end.

joerunde: > Typically, it's used to modify attributes of modules, which are global state. When used this way, it is not thread safe.

Yeah, sounds like we'd need to patch a specific model instance for the test, not the full module :/

joerunde (approved):
ship it!

Issue #517, opened 2023-10-10
Allow third-party implementations for data sources


joerunde:
## Is your feature request related to a problem? Please describe.

We've made some assumptions in caikit that for places where data locations are described in our APIs, like `DataStreamSource` parameters or the `output_path` top-level training api parameter, either files will simply exist on disk or an s3 blob is "good enough" to locate the data.

However, for security-minded applications, this requires that a consuming application be responsible for securely fetching user data and providing it to caikit. This:
- may be onerous in the context of a distributed application where services that need data are always responsible for fetching it themselves
- may not meet security requirements if the user data has to land on disk for caikit to read

## Describe the solution you'd like

Similar to how caikit allows third-party implementations of `trainers` `finders` and `initializers`, we think an abstraction for data sources may be necessary. This would allow users of caikit to plug in their own data fetching implementation to meet their own security requirements.

## Additional context

This may affect the generated service APIs?????

gabe-l-hart:
This has been completed with the addition of Output Target in #546

PR #516, opened 2023-10-09
:sparkles: Allow for extra fields in data sources

evaline-ju:
**What this PR does / why we need it**:
Closes #513 
- Users want to be able to provide training data files with extra fields that might not be strictly necessary for the training process without erroring
- Re-orders the methods tried for parsing data source files without extensions to allow multiparts to be tried first, otherwise they may be incorrectly attempted to be parsed by other methods - note: this does not address "missing" data model fields being represented as null since [protobuf is used to parse json](https://github.com/caikit/caikit/blob/b5c12d7bd93db40b6f53e9cef6e90be8113fe538/caikit/core/data_model/base.py#L765-L767)
- Removes some unneeded test fixture parameters

**Special notes for your reviewer**:

**If applicable**:
- [ ] this PR contains documentation
- [x] this PR contains unit tests
- [ ] this PR has been tested for backwards compatibility

PR #520, opened 2023-10-11
Pluggable data input

gabe-l-hart:

Addresses #517 

This PR refactors `DataStreamSource` to us a pluggable/extensible factory pattern to allow for external library extensions. Concretely, this adds the following:

* New config section `data_streams.source_plugins` to define the set of source plugins
    * The keys in this map are arbitrary labels that are used for logging, but nothing else. Do we like this?
* New base class (`DataStreamSourcePlugin`) that can be extended for new plugins
* New factory for stream source plugins

The ADR also contains details about plans to handle pluggable output with `OutputTarget`s, though those are not implemented in this PR

**If applicable**:

### DIFF
--- a/caikit/runtime/service_generation/data_stream_source.py
+++ b/caikit/runtime/service_generation/data_stream_source.py
@@ -15,7 +15,7 @@
 # Standard
 from functools import partial
 from glob import glob
-from typing import Any, List, Optional, Type, Union, Callable
+from typing import Any, List, Optional, Type, Union, Callable, Dict
 import abc
 import os
 import sys
@@ -270,11 +270,21 @@ class DirectoryDataStreamSourcePlugin(FilePluginBase):
 
 
 class JsonDataStreamSourcePlugin(DataStreamSourcePlugin):
-    """This plugin has instantiation logic: it needs the stream's element type so that it can
+    """This plugin is for inline data, elements are provided in a list.
+
+    This plugin has instantiation logic: it needs the stream's element type so that it can
     generate a data model for List[element_type]"""
 
+    # class-level cache required to avoid creating duplicate data model classes
+    stream_source_type_cache: Dict[Type[DataBase], Type[DataBase]] = {}
+
     def __init__(self, element_type: Type[DataBase]):
-        """🌶🌶🌶️: This will probably explode if instantiated twice for the same type"""
+        """Creates a new data model to wrap the element type in a list.
+        This new data model is the message type for the stream"""
+        if element_type in JsonDataStreamSourcePlugin.stream_source_type_cache:
+            self.stream_message_type = JsonDataStreamSourcePlugin.stream_source_type_cache[element_type]
+            return
+
         package = get_runtime_service_package()
         cls_name = self._make_data_stream_source_type_name(element_type)
 
@@ -287,6 +297,7 @@ class JsonDataStreamSourcePlugin(DataStreamSourcePlugin):
         )
 
         self.stream_message_type = JsonData
+        JsonDataStreamSourcePlugin.stream_source_type_cache[element_type] = self.stream_message_type
 
     def get_stream_message_type(self) -> Type[DataBase]:
         return self.stream_message_type
--- a/caikit/runtime/service_generation/data_stream_source.py
+++ b/caikit/runtime/service_generation/data_stream_source.py
@@ -340,7 +340,7 @@ class S3FilesDataStreamSourcePlugin(DataStreamSourcePlugin):
 
     def to_data_stream(self, *_, **__) -> DataStream:
         error(
-            "<COR80419785E>",
+            "<RUN80419785E>",
             NotImplementedError(
                 "S3Files are not implemented as stream sources in this runtime."
             ),
@@ -381,7 +381,7 @@ class DataStreamPluginFactory(ImportableFactory):
                 if field_numbers.count(plugin.get_field_number()) > 1
             ]
             error.value_check(
-                "<COR69189361E>",
+                "<RUN69189361E>",
                 not duplicate_field_number_names,
                 "Duplicate plugin field numbers found for plugins: {}",
                 duplicate_field_number_names,
@@ -447,14 +447,14 @@ class DataStreamSourceBase(DataStream):
         for field_name in self.get_proto_class().DESCRIPTOR.fields_by_name:
             if getattr(self, field_name) is not None:
                 error.value_check(
-                    "<COR80421785E>",
+                    "<RUN80421785E>",
                     set_field is None,
                     "Found DataStreamSource with multiple sources set: {} and {}",
                     set_field,
                     field_name,
                 )
                 error.value_check(
-                    "<COR80420785E>",
+                    "<RUN80420785E>",
                     field_name in self.name_to_plugin_map,
                     "no data stream plugin found for field: {}",
                     field_name,
@@ -504,7 +504,7 @@ def make_data_stream_source(
             if all_field_names.count(field_name) > 1
         }
         error.value_check(
-            "<COR66854455E>",
+            "<RUN66854455E>",
             not duplicates,
             "Duplicate plugin field names found for type {}: {}",
             data_element_type,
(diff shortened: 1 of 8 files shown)


caikit/runtime/service_generation/data_stream_source.py
```diff
@@ -52,124 +57,74 @@
...
 
-        # If a S3 pointer is given, raise not implemented
-        if set_field == "s3files":
-            error(
-                "<COR80419785E>",
-                NotImplementedError(
-                    "S3Files are not implemented as stream sources in this runtime."
-                ),
-            )
+    def get_field_name(self, element_type: type) -> str:
```
gabe-l-hart, line 96: I was really torn on this one. One the one hand, I would prefer if this were a class property, or at the very least an instance property and NOT dependent on the `element_type`, but on the other hand, the default implementation needs to know the name of the stream message which _is_ dependent on the `element_type`.

caikit/runtime/service_generation/data_stream_source.py
```diff
@@ -52,124 +57,74 @@
...
+            # be needed by the data objects, we ignore unknown fields
+            return element_type.from_json(raw_element, ignore_unknown_fields=True)
+        return raw_element
+
+    @staticmethod
+    def _to_element_partial(element_type: type) -> Callable:
+        return partial(DataStreamSourcePlugin._to_element_type, element_type)
+
+
+class FilePluginBase(DataStreamSourcePlugin):
```
gabe-l-hart, line 115: My expectation is that this will likely be very useful for derived implementations as well, especially those that are fetching files from external sources like S3

caikit/runtime/service_generation/data_stream_source.py
```diff
@@ -255,53 +202,354 @@ def _get_resolved_source_path(input_path: str) -> str:
...
+        oneof members.
+        """
+        new_inst = self.__class__.from_binary_buffer(pickle_bytes)
+        setattr(self, new_inst.which_oneof("data_stream"), new_inst.data_stream)
+        self.generator_func = self._generator
+        self.generator_args = tuple()
+        self.generator_kwargs = {}
+
+    @property
+    def name_to_plugin_map(self):
```
gabe-l-hart, line 434: I think we might be able to do this the "official" way: https://stackoverflow.com/a/64647117

caikit/runtime/service_generation/data_stream_source.py
```diff
@@ -255,53 +202,354 @@ def _get_resolved_source_path(input_path: str) -> str:
...
-            package=package,
-            proto_name=f"{cls_name}JsonData",
-            name="JsonData",
-            attrs={"__qualname__": f"{cls_name}.JsonData"},
-            annotations={"data": List[data_element_type]},
+        # Get the required plugins
+        plugins = plugin_factory.get_plugins(plugins_config)
+
+        # Make sure there are no field name duplicates
+        plug_to_name = {
```
gabe-l-hart, line 498: This work to check for duplicates isn't free, but it should all be boot-time work, so I'm ok not putting this behind an enablement flag to turn of for performance

tests/runtime/service_generation/test_data_stream_source.py
```diff
@@ -309,6 +299,7 @@ def test_make_data_stream_source_from_multipart_formdata_file(
     sample_multipart_json,
     sample_multipart_csv,
     sample_multipart_json_with_content_header,
+    sample_train_service,
```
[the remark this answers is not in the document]
joerunde: ah- yeah so if you run any of these tests in isolation, nothing will have created sample_lib's training APIs, so the `DataStreamSourceSampleTrainingType` class will not exist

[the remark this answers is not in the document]
joerunde: we should probably just stick an @autouse fixture at the top of this file

tests/runtime/service_generation/test_data_stream_plugins.py
```diff
@@ -0,0 +1,238 @@
...
+            },
+        },
+        override_env_vars=False,
+    )
+    plugin_factory.get_plugins(cfg)
+
+    class LocalFoobar:
+        pass
+
+    make_data_stream_source(LocalFoobar, plugin_factory, cfg)
```
joerunde, line 238: todo: maybe let's toss in some asserts about the class generated here?

tests/runtime/service_generation/test_data_stream_source.py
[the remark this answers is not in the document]
evaline-ju: ah I probably removed some of these in a previous PR #516 - autouse would be more helpful for the implications then

caikit/runtime/service_generation/data_stream_source.py
joerunde: 🤔  but the result of this should never include any details about the element type?

i.e. it should just be `file`, `jsondata`, `directory`...

Maybe worth some more huddle time and thinking

gabe-l-hart: I _could_ imagine a circumstance where you would want to have the element type inform this, right? Not sure it's a good idea, but I could imagine it.

caikit/runtime/service_generation/data_stream_source.py
```diff
@@ -255,53 +202,354 @@ def _get_resolved_source_path(input_path: str) -> str:
...
+            NotImplementedError(
+                "S3Files are not implemented as stream sources in this runtime."
+            ),
+        )
+
+    def get_field_number(self) -> int:
+        return 5
+
+
+## DataStreamSourceRegistry ####################################################
```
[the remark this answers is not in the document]
gabe-l-hart: Good catch!

docs/adrs/022-pluggable-runtime-io.md
```diff
@@ -0,0 +1,65 @@
...
+```
+
+**Example Config**
+```yaml
+data_streams:
+    source_plugins:
+        inline:
+            type: JsonData
+        custom:
+            type: MINE
```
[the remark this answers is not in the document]
gabe-l-hart: Yes, the `import_class` is tested with `ImportableFactory` already (it's used for `finder`/`initialiazer`/`trainer` factories as well). It's an interesting question about how duplicate `name` registrations are handled though.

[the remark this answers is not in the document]
gabe-l-hart:
```py
    def register(self, constructible: Type[FactoryConstructible]):
        """Register the given constructible"""
        current = self._registered_types.get(constructible.name)
        error.value_check(
            "<COR27473932E>",
            current is None or current is constructible,
            "Conflicting registration of {}",
            constructible.name,
        )
```

Looks like it's [already handled](https://github.com/caikit/caikit/blob/main/caikit/core/toolkit/factory.py#L79) in the base `Factory`

caikit/runtime/service_generation/data_stream_source.py
```diff
@@ -255,53 +204,354 @@ def _get_resolved_source_path(input_path: str) -> str:
...
+    """Unimplemented!"""
+
+    name = "S3Files"
+
+    def get_stream_message_type(self, *_, **__) -> Type[DataBase]:
+        return S3Files
+
+    def to_data_stream(self, *_, **__) -> DataStream:
+        error(
+            "<COR80419785E>",
```
evaline-ju, line 343:
```suggestion
            "<RUN80419785E>",
```

caikit/runtime/service_generation/data_stream_source.py
```diff
@@ -255,53 +204,354 @@ def _get_resolved_source_path(input_path: str) -> str:
...
+
+            # Make sure field numbers are unique
+            field_numbers = [plugin.get_field_number() for plugin in self._plugins]
+            duplicate_field_number_names = [
+                plugin.name
+                for plugin in self._plugins
+                if field_numbers.count(plugin.get_field_number()) > 1
+            ]
+            error.value_check(
+                "<COR69189361E>",
```
evaline-ju, line 384:
```suggestion
                "<RUN69189361E>",
```
etc.

evaline-ju (commented):
nit while looking at output targets

tests/runtime/service_generation/test_data_stream_source.py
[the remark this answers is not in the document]
evaline-ju: added an update to autouse a fixture in this file for this

PR #522, opened 2023-10-12
fix CORS

itaijj:

**If applicable**:

### DIFF
--- a/caikit/runtime/http_server/http_server.py
+++ b/caikit/runtime/http_server/http_server.py
@@ -34,8 +34,8 @@ import time
 from fastapi import FastAPI, Request, Response, status
 from fastapi.encoders import jsonable_encoder
 from fastapi.exceptions import ResponseValidationError
-from fastapi.responses import JSONResponse, PlainTextResponse
 from fastapi.middleware.cors import CORSMiddleware
+from fastapi.responses import JSONResponse, PlainTextResponse
 from grpc import StatusCode
 from sse_starlette import EventSourceResponse, ServerSentEvent
 import pydantic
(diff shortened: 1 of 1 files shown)


caikit/runtime/http_server/http_server.py
```diff
@@ -120,7 +121,13 @@ def __init__(self, tls_config_override: Optional[aconfig.Config] = None):
...
 
         self.app = FastAPI()
-
+        self.app.add_middleware(
+            CORSMiddleware,
+            allow_origins=["*"],
+            allow_credentials=True,
+            allow_methods=["*"],
+            allow_headers=["*"],
+        )
```
joerunde, line 130: Thanks @itaijj!

I think the two things that need to happen here are:
- This whole block should be behind an `if` statement that checks a config item to see if the user wants cors middleware at all
- The four values in here should be pulled from configuration instead of using `*` and `True`

Take a look at the `caikit/config/config.yml` file, that's where the default values for the configuration returned by `caikit.get_config()` are stored. I think we'd probably want something that looks like this:

```
runtime:
  http:
    cors_middleware:
      enabled: False
      allow_origins: "*"
      allow_credentials: True
      allow_methods: "*"
      allow_headers: "*"
```

and then use the configuration values here in the code.


It would also be great to get two test cases added- one where a cross-origin request fails and one where we turn this feature on and it passes. That said I don't know enough about CORS to know if that's even possible to do in a unit test, so I wouldn't block this PR on not having tests if that's the case.

joerunde (changes requested):
Nice!

Two things need to happen to get this merged:
- Your commits need to be signed off for DCO to pass (see https://cert-manager.io/docs/contributing/sign-off/)
- The cors middleware needs to be configurable

gabe-l-hart:
@itaijj Is this PR still needed on your end? I see that @joerunde added suggestions for how to make this configurable which we'll need implemented before merging. It also looks like we need to sort out the DCO failure still.

gabe-l-hart:
@itaijj This PR looks to have gone stale. I'm going to close it now, but we can revisit Joe's suggestions if you end up needing it in the future

itaijj:
I found a solution by inheritance from RuntimeHTTPServer, it also solves other issues with doing extra configs to fastAPI

gabe-l-hart:
Great! Glad that solution is a working path for you

PR #530, opened 2023-10-17
Remove modules.json from tests

aluu317:

Currently, test_dump_services.py calls `dump_grpc_services` and this function always prints out a `modules.json` file. This PR makes this an argument, and don't print out a `modules.json` file in unit tests, only in `start_runtime_with_sample_lib.py` do we want this file printed out

**If applicable**:

### DIFF
--- a/caikit/runtime/dump_services.py
+++ b/caikit/runtime/dump_services.py
@@ -29,14 +29,14 @@ import caikit
 log = alog.use_channel("RUNTIME-DUMP-SVC")
 
 
-def dump_grpc_services(output_dir: str):
+def dump_grpc_services(output_dir: str, write_modules_file):
     """Utility for rendering the all generated interfaces to proto files"""
     inf_enabled = get_config().runtime.service_generation.enable_inference
     train_enabled = get_config().runtime.service_generation.enable_training
 
     if inf_enabled:
         inf_svc = ServicePackageFactory.get_service_package(
-            ServicePackageFactory.ServiceType.INFERENCE, write_modules_file=True
+            ServicePackageFactory.ServiceType.INFERENCE, write_modules_file=write_modules_file
         )
     if train_enabled:
         train_svc = ServicePackageFactory.get_service_package(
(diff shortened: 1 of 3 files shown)


caikit/runtime/dump_services.py
```diff
@@ -87,12 +87,15 @@ def dump_http_services(output_dir: str):
 
 
 if __name__ == "__main__":
-    assert len(sys.argv) == 2, f"Usage: {sys.argv[0]} <output_dir>"
+    assert (
+        len(sys.argv) == 2
+    ), f"Usage: {sys.argv[0]} <output_dir>, <write_modules_json>"
     out_dir = sys.argv[1]
+    write_modules_json = sys.argv[2]
```
gabe-l-hart, line 94: I think we might need some `bool` parsing on this since `python -m caikit.runtime.dump_services . false` would be "truthy". Realistically, this probably means adding `argparse` to this command. I think we would want  to preserve `output_dir` as a positional arg to avoid breaking compatibility.

gabe-l-hart (changes requested):
One little cleanup on how the bool is parsed

caikit/runtime/dump_services.py
```diff
@@ -99,12 +101,37 @@ def dump_http_services(output_dir: str):
...
 if __name__ == "__main__":
-    assert len(sys.argv) == 2, f"Usage: {sys.argv[0]} <output_dir>"
-    out_dir = sys.argv[1]
+    parser = argparse.ArgumentParser(
+        description="Dump grpc and http services for inference and train"
+    )
+
+    # Add an argument for the output_dir
+    parser.add_argument(
+        "-o",
```
gabe-l-hart, line 110: To maintain backwards compatibility, I think we should make this a positional argument ([docs](https://docs.python.org/3/library/argparse.html#core-functionality))

gabe-l-hart (changes requested):
One little change request to use a positional arg and maintain compatibility

gabe-l-hart (approved):
Ship it!

PR #533, opened 2023-10-18
gha: simplify env matrix

dtrifiro:
This simplifies the matrix logic a little bit.

### DIFF
--- a/.github/workflows/build-library.yml
+++ b/.github/workflows/build-library.yml
@@ -25,31 +25,26 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version:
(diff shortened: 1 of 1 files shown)


joerunde:
@dtrifiro does this work because there's only one version of python installed for each matrix entry?

dtrifiro:
I updated the actions so that `proto3` and `core` tests also run as part of the 3.8 job. I'd have preferred to have them run as a separate matrix entry, but I guess this will do for now.

dtrifiro:
Fair enough. I'll try and offload `proto3` and `core` so separate matrix entries.

dtrifiro:
Sorry, no time to get back to this.

PR #535, opened 2023-10-19
Sort Modules in inference rpcs generation

aluu317:

When generating inference rpcs, we use `modules` as a set, not a list, so the order of modules could be undeterministic. 

This PR sorts that modules list by Module ID so the list is fixed and deterministic.

Ex: SampleTaskRequest contains inference signatures from ListModule, SampleModule, PrimitiveSampleModule. Because this is a set of modules, it depends on which module is processed first, SampleTaskRequest will have different field ordering.

May be one of the fixes for https://github.com/caikit/caikit-nlp/issues/237.

**If applicable**:


gabe-l-hart (approved):
Excellent! Great find @aluu317

Issue #532, opened 2023-10-17
regression? HTTP server won't start because: AttributeError: type object 'list' has no attribute 'model_json_schema'


markstur:
## Describe the bug

A list in a task (run) signature like this:

@task(
    required_parameters={"source_sentence": str, "sentences": List[str]},
    output_type=FloatSequence
)

Works for gRPC, but HTTP server fails with:

> AttributeError: type object 'list' has no attribute 'model_json_schema'

## Platform

Please provide details about the environment you are using, including the following:

- Interpreter version: 3.11
- Library version:  caikit==0.22.0

## Sample Code

See above.

## Expected behavior

I think this is a regression.  It used to work.  Just updated to new release.

## Observed behavior

What you see happening (error messages, stack traces, etc...)

## Additional context

Should've mentioned that I ran into this while trying the new multi-task, but I don't think that is relevant.

Added stack trace (in comment) showing where the pydantic_model.model_json_schema() is now a problem.  Looks like here: https://github.com/caikit/caikit/commit/eea690d3611182e953ea682e46398e397a33112a#diff-8c48c03bdb3588a96567eff810341f816e17768a82d89ab2a501e7d8649abc94R391

markstur:
Traceback (most recent call last):
  File "/Users/markstur/gh/caikit-serving/demo/server/start_runtime.py", line 41, in <module>
    main()
  File "/Users/markstur/gh/caikit-serving/venv/lib/python3.11/site-packages/caikit/runtime/__main__.py", line 87, in main
    _http_server = RuntimeHTTPServer()
                   ^^^^^^^^^^^^^^^^^^^
  File "/Users/markstur/gh/caikit-serving/venv/lib/python3.11/site-packages/caikit/runtime/http_server/http_server.py", line 150, in __init__
    self._bind_routes(self.inference_service)
  File "/Users/markstur/gh/caikit-serving/venv/lib/python3.11/site-packages/caikit/runtime/http_server/http_server.py", line 276, in _bind_routes
    self._add_unary_input_unary_output_handler(rpc)
  File "/Users/markstur/gh/caikit-serving/venv/lib/python3.11/site-packages/caikit/runtime/http_server/http_server.py", line 392, in _add_unary_input_unary_output_handler
    responses=self._get_response_openapi(
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #536, opened 2023-10-19
Fix bug in HTTP Server with Sequences

HonakerM:
This PR fixes issue #532. There was a slight regression with the HTTP server that this fixes and adds a test case to ensure this doesn't happen in the future.

**If applicable**:


tests/fixtures/sample_lib/data_model/sample.py
```diff
@@ -103,6 +103,19 @@ class FileTask(TaskBase):
...
+    required_parameters={
+        "str_type": str,
+        "list_str_type": List[str],
+        "int_type": int,
+        "list_int_type": List[int],
+    },
+    unary_output_type=StrSequence,
+)
+class PrimitiveTask(TaskBase):
+    """A sample task for testing ra"""
```
joerunde, line 116: What's `ra`?

HonakerM: 🤦 a spelling mistake. Just fixed it

tests/fixtures/sample_lib/modules/primitive/primitive_module.py
```diff
@@ -0,0 +1,52 @@
...
+from ...data_model.sample import PrimitiveTask
+from caikit.core.modules import ModuleLoader
+from caikit.interfaces.common.data_model import StrSequence
+import caikit.core
+
+
+@caikit.core.module(
+    "bd87fbbe-caad-4db6-93f5-37cc3dbedef5", "PrimitiveModule", "0.0.1", PrimitiveTask
+)
+class PrimitiveModule(caikit.core.ModuleBase):
```
joerunde, line 31: @HonakerM Does this problem only occur if the `List[T]` types are in the task's set of required parameters? We do have a `sample_lib.modules.sample_task.SamplePrimitiveModule` already which has a bunch of primitive types including `List[T]` and `Union[List[T]...]`, would that one work as is if we called it via the http server?

HonakerM: @joerunde I missed that one! It should be testing the same stuff so I'll remove my custom module.

joerunde (approved):
get it!

PR #538, opened 2023-10-20
✨ add stream event types

prashantgupta24:

This PR makes 2 changes:

### Change 1
Right now in streaming APIs, we only sent in the `data` but not the type of event.
```
data: {
data:   "greeting": "Hello world stream"
data: }
```
Users could not determine whether an event was an error or a message.

This PR adds a `event` type to the messages, so that they look like:
```
event: message
data: {"greeting": "Hello world stream"}
```
or
```
event: error
data: {"details": "ValueError('raising a ValueError')", "code": 400}
```
### Change 2

This PR also makes the `http_server` catch `TypeError, ValueError` type errors, so that they are now tagged with `400`s instead of `500`s which was happening before and which was incorrect. Currently, in case of streaming, if the error happens while the stream is consumed, the error handling has to be done by the `http_server`, which doesn't have separate cases for type/value error.

(Also note the `to_json` part has been added to return the result in a single line, thanks Gabe)

**If applicable**:

### DIFF
--- a/caikit/runtime/http_server/http_server.py
+++ b/caikit/runtime/http_server/http_server.py
@@ -32,6 +32,7 @@ import ssl
 import tempfile
 import threading
 import time
+import traceback
 
 # Third Party
 from fastapi import FastAPI, HTTPException, Request, Response, status
@@ -512,6 +513,18 @@ class RuntimeHTTPServer(RuntimeServerBase):
                     return
                 except HTTPException as err:
                     raise err
+                except (TypeError, ValueError) as err:
+                    log_dict = {
+                        "log_code": "<RUN76624264W>",
+                        "message": repr(err),
+                        "stack_trace": traceback.format_exc(),
+                    }
+                    log.warning(log_dict)
+                    error_code = 400
+                    error_content = {
+                        "details": repr(err),
+                        "code": error_code,
+                    }
                 except CaikitRuntimeException as err:
                     error_code = GRPC_CODE_TO_HTTP.get(err.status_code, 500)
                     error_content = {
--- a/caikit/runtime/http_server/http_server.py
+++ b/caikit/runtime/http_server/http_server.py
@@ -19,6 +19,7 @@ API based on the task definitions available at boot.
 # Standard
 from contextlib import contextmanager
 from dataclasses import dataclass
+from enum import Enum
 from functools import partial
 from typing import Any, Dict, Iterable, Optional, Type, Union, get_args
 import asyncio
@@ -105,6 +106,13 @@ MODEL_ID = "model_id"
 
 # Endpoint to use for health checks
 HEALTH_ENDPOINT = "/health"
+
+
+# Stream event types enum
+class StreamEventTypes(str, Enum):
+    MESSAGE = "message"
+    ERROR = "error"
+
 
 # Small dataclass for consolidating TLS files
 @dataclass
@@ -400,7 +408,6 @@ class RuntimeHTTPServer(RuntimeServerBase):
         async def _handler(
             context: Request,
         ) -> Response:
-
             request = await pydantic_from_request(pydantic_request, context)
             request_params = self._get_request_params(rpc, request)
 
@@ -499,7 +506,7 @@ class RuntimeHTTPServer(RuntimeServerBase):
                         )
                     ):
                         yield {
-                            "event": "message",
+                            "event": StreamEventTypes.MESSAGE,
                             "data": result.to_json(),
                         }
                     return
@@ -508,7 +515,7 @@ class RuntimeHTTPServer(RuntimeServerBase):
                 except CaikitRuntimeException as err:
                     error_code = GRPC_CODE_TO_HTTP.get(err.status_code, 500)
                     error_content = {
-                        "event": "error",
+                        "event": StreamEventTypes.ERROR,
                         "details": err.message,
                         "code": error_code,
                         "id": err.id,
@@ -516,7 +523,7 @@ class RuntimeHTTPServer(RuntimeServerBase):
                 except Exception as err:  # pylint: disable=broad-exception-caught
                     error_code = 500
                     error_content = {
-                        "event": "error",
+                        "event": StreamEventTypes.ERROR,
                         "details": f"Unhandled exception: {str(err)}",
                         "code": error_code,
                         "id": None,
--- a/caikit/runtime/http_server/http_server.py
+++ b/caikit/runtime/http_server/http_server.py
@@ -506,10 +506,9 @@ class RuntimeHTTPServer(RuntimeServerBase):
                             **request_params,
                         )
                     ):
-                        yield {
-                            "event": StreamEventTypes.MESSAGE.value,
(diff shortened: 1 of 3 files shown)


caikit/runtime/http_server/http_server.py
```diff
@@ -497,20 +497,25 @@ async def _generator() -> pydantic_response:
...
+                            "event": "message",
+                            "data": result.to_json(),
+                        }
                     return
                 except HTTPException as err:
                     raise err
                 except CaikitRuntimeException as err:
                     error_code = GRPC_CODE_TO_HTTP.get(err.status_code, 500)
                     error_content = {
+                        "event": "error",
```
prashantgupta24, line 510: Wondering how to simulate error cases in streaming 🤔

caikit/runtime/http_server/http_server.py
```diff
@@ -497,20 +497,25 @@ async def _generator() -> pydantic_response:
                             **request_params,
                         )
                     ):
-                        yield result
+                        yield {
+                            "event": "message",
```
gabe-l-hart, line 501: Since these will be standard values, we should use an enum to avoid typos down the road

gabe-l-hart (changes requested):
Thanks Prashant! This will definitely be an API breaking change. Have you vetted this with current stream users? I think it's a good fix, but we need to ensure it's well communicated.

Also, one change request for an enumerate with the event types

caikit/runtime/http_server/http_server.py
prashantgupta24: Ah good point

prashantgupta24:
> Have you vetted this with current stream users? 

We got a request from an internal team for something similar. Their issue was that they could not determine whether an event was an error or a message without a separate type.

tests/runtime/http_server/test_http_server.py
```diff
@@ -466,6 +466,36 @@ def test_inference_streaming_sample_module(sample_task_model_id, runtime_http_se
...
+        json_input = {
+            "model_id": sample_task_model_id,
+            "inputs": {"name": "world"},
+            "parameters": {"err_stream": True},
+        }
+        stream = client.post(
+            f"/api/v1/task/server-streaming-sample",
+            json=json_input,
+        )
+        assert stream.status_code == 200
```
prashantgupta24, line 483: is it expected that the status code remains 200 even if the stream errors out? 🤔

evaline-ju: I think this would be expected for streams

caikit/runtime/http_server/http_server.py
```diff
@@ -106,6 +107,13 @@
 # Endpoint to use for health checks
 HEALTH_ENDPOINT = "/health"
 
+
+# Stream event types enum
+class StreamEventTypes(str, Enum):
```
prashantgupta24, line 112: remove `str`

tests/runtime/http_server/test_http_server.py
```diff
@@ -477,41 +478,101 @@ def test_inference_streaming_sample_module(sample_task_model_id, client):
...
+                .replace("\r\n", "")
+                .replace("}{", "}, {")
+            )
+        )
+        assert len(stream_responses) == 2
+        assert stream_responses[0].get("event") == StreamEventTypes.ERROR.value
+        assert (
+            stream_responses[1].get("details") == "ValueError('raising a ValueError')"
         )
+        assert stream_responses[1].get("code") == 400
```
prashantgupta24, line 575: This is why we had to introduce catching `Value/Type` errors in the `http_server`. Previously, these errors would have been labeled witn `500`

caikit/runtime/http_server/http_server.py
```diff
@@ -519,7 +542,9 @@ async def _generator() -> pydantic_response:
                     log.error("<RUN51881106E>", err, exc_info=True)
 
                 # If an error occurs, yield an error response and terminate
-                yield ServerSentEvent(data=json.dumps(error_content))
+                yield ServerSentEvent(
+                    data=json.dumps(error_content), event=StreamEventTypes.ERROR.value
```
gabe-l-hart, line 546: Ok, looks like maybe `str` is intended?

tests/runtime/http_server/test_http_server.py
```diff
@@ -477,41 +478,101 @@ def test_inference_streaming_sample_module(sample_task_model_id, client):
...
-    but pints the actual running server"""
+    but pings the actual running server"""
 
     for i in range(10):
         input = {"model_id": sample_task_model_id, "inputs": {"name": f"world{i}"}}
         url = f"http://localhost:{runtime_http_server.port}/api/v1/task/server-streaming-sample"
         stream = requests.post(url=url, json=input, verify=False)
         assert stream.status_code == 200
         stream_content = stream.content.decode(stream.encoding)
+        print(stream_content)
```
gabe-l-hart, line 516: Errant `print`

tests/runtime/http_server/test_http_server.py
gabe-l-hart: ah! That makes sense.

gabe-l-hart (changes requested):
One errant `print` and a question about whether we want to double-serialize the `data` field

caikit/runtime/http_server/http_server.py
```diff
@@ -498,10 +506,25 @@ async def _generator() -> pydantic_response:
                             **request_params,
                         )
                     ):
-                        yield result
+                        yield {
+                            "event": StreamEventTypes.MESSAGE.value,
+                            "data": result.to_json(),
```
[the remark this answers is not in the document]
prashantgupta24: Hmm, I guess a single line str is easier to parse vs multi-line `data: dict` statements? Because without this `to_json`, we get
```
data: {
data:   "greeting": "Hello world stream"
data: }
```

That's what the internal team wanted us to do, which made sense to me. We can discuss this further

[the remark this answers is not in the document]
prashantgupta24: Also note that yielding just the `result` part back _still_ sends in something like this:
```
data: {
data:   "greeting": "Hello world0 stream"
data: }
```
Note that `data` is still part of the returned message

caikit/runtime/http_server/http_server.py
prashantgupta24: Yep in case of error we do dump out the str

caikit/runtime/http_server/http_server.py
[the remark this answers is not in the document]
gabe-l-hart: Ok, I found the context I was missing. It looks like `data` and `event` are "first class things" in SSE and not just random strings we happen to be using, so my concern about double-decoding is invalid

CITE: https://github.com/sysid/sse-starlette/blob/159bbece1a58434d12d67934dc99fab556737247/sse_starlette/sse.py#L50

gabe-l-hart (approved):
Looks great, thanks Prashant!

PR #539, opened 2023-10-20
🐛 Stream test bug fix

prashantgupta24:

This PR adds:
- A fixture for the `fastapi.testclient`
- Skips testing streaming cases with FastAPI's testclient, pending resolution of https://github.com/tiangolo/fastapi/discussions/10518
- Instead, test streaming tests against an actual server.

fix https://github.com/caikit/caikit/issues/540

- Following suggestions from:
-  https://github.com/encode/starlette/issues/1315#issuecomment-1166600293 
- https://github.com/sysid/sse-starlette/issues/68#issuecomment-1678266759

**If applicable**:

### DIFF
--- a/tests/runtime/http_server/test_http_server.py
+++ b/tests/runtime/http_server/test_http_server.py
@@ -461,6 +461,9 @@ def test_output_file_task(file_task_model_id, client):
             assert pdf_result.read() == b"bounding|pdfdata\xff\xff\x00|box"
 
 
+@pytest.mark.skip(
+    "Skipping testing streaming cases with FastAPI's testclient, pending resolution https://github.com/tiangolo/fastapi/discussions/10518"
+)
 def test_inference_streaming_sample_module(sample_task_model_id, client):
     """Simple check for testing a happy path unary-stream case"""
     json_input = {"model_id": sample_task_model_id, "inputs": {"name": "world"}}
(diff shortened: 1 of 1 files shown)


tests/runtime/http_server/test_http_server.py
```diff
@@ -457,6 +457,34 @@ def test_inference_streaming_sample_module(sample_task_model_id, runtime_http_se
         )
 
 
+def test_inference_streaming_sample_module_another_test(
```
prashantgupta24, line 460: This is the exact same test as the `test_inference_streaming_sample_module`, but for some reason, this test hands indefinitely if they are both run together

prashantgupta24: FYI the grpc test cases run perfectly fine, from my initial investigation, it seems like a bug with asyncio or sse, both of which target the http server

tests/runtime/http_server/test_http_server.py
```diff
@@ -43,6 +43,15 @@
     runtime_http_test_server,
 )
 
+## Fixtures #####################################################################
+
+
+@pytest.fixture(scope="session")
```
gabe-l-hart, line 49: Why is this session scoped? I'm wondering if this is hiding the real issue of concurrent requests with multiple connections?

joerunde:
> ... so that it's used within a context manager.

But all the tests shown already were using it within a context manager? 😕😕😕 

It looks like what this is doing is really moving to a single `TestClient` instead of having one per test. I'm totally down for this anyway since it's much DRYer but still confused about the root cause here.

tests/runtime/http_server/test_http_server.py
prashantgupta24: Yeah without the session scope it's basically the same thing we had before, and it fails on the second stream test

prashantgupta24:
@joerunde yeah the only difference was the session scoping of the fixture that was making it work. It doesn't work without that unfortunately :(

joerunde (approved):
Honestly, I'm here for the test code simplification, the fixture for the client is really nice. Let us know if you find out any more details about why this thing hangs!

prashantgupta24:
@joerunde Now I'm thinking making the fixture session-scoped is not the right thing to do. Adding another test that pings an actual running server and using that along with either a function scoped or a session scoped test fixture doesn't seem to work :( 

My gut feeling is that `TestClient` and `EventSourceResponse` don't seem to work well together. If I don't use `TestClient`, then running any number of tests against the stream endpoint seem to work.

tests/runtime/http_server/test_http_server.py
gabe-l-hart: Hm, ok, that actually seems like this doesn't really solve the problem then, right? Is this a property of the `TestClient` itself or rather a property of the server app being connected to? If it's the former, this is fine as the "problem" lies in the testing infrastructure, but if it's the later, this is just hiding that and is probably not the right fix.

prashantgupta24: Removing session scoping, I agree it's not the right thing to do in this case.

PR #542, opened 2023-10-23
Fix GRPC server over Unix socket

HonakerM:
This PR updates the GRPCServer unix socket implementation to actually allow listening over unix sockets. Prior to this the server would always skip adding the unix socket even if the target directory existed.

**If applicable**:

### DIFF
--- a/tests/runtime/test_grpc_server.py
+++ b/tests/runtime/test_grpc_server.py
@@ -1283,23 +1283,23 @@ def test_construct_with_options(open_port, sample_train_service, sample_int_file
             assert context.value.code() == grpc.StatusCode.RESOURCE_EXHAUSTED
 
 
-def test_grpc_server_startup(open_port, sample_train_service, sample_int_file):
-    """Make sure that the server can be booted with config options"""
-    socket_dir = tempfile.TemporaryDirectory()
-    with temp_config(
-        {"runtime": {"grpc": {"unix_socket_path": socket_dir.name + "/grpc.sock"}}},
-        "merge",
-    ):
-        server = RuntimeGRPCServer()
-        with server:
-            stub = model_runtime_pb2_grpc.ModelRuntimeStub(
-                grpc.insecure_channel(f"unix://{socket_dir.name}/grpc.sock")
-            )
-            runtime_status_request = model_runtime_pb2.RuntimeStatusRequest()
-            actual_response = stub.runtimeStatus(runtime_status_request)
-            assert (
-                actual_response.status == model_runtime_pb2.RuntimeStatusResponse.READY
-            )
+def test_grpc_server_socket_listen():
+    """Make sure that the server correctly listen on a unix socket"""
+    with tempfile.TemporaryDirectory() as socket_dir:
+        with temp_config(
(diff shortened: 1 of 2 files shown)


tests/runtime/test_grpc_server.py
```diff
@@ -1283,6 +1283,25 @@ def test_construct_with_options(open_port, sample_train_service, sample_int_file
             assert context.value.code() == grpc.StatusCode.RESOURCE_EXHAUSTED
 
 
+def test_grpc_server_startup(open_port, sample_train_service, sample_int_file):
+    """Make sure that the server can be booted with config options"""
+    socket_dir = tempfile.TemporaryDirectory()
```
joerunde, line 1288: You can request `tmp_path` in the test args and pytest will give you a temporary directory: https://docs.pytest.org/en/7.4.x/how-to/tmp_path.html

But if this was meant to be named `socket_dir` for readability that's 👍

tests/runtime/test_grpc_server.py
```diff
@@ -1283,6 +1283,25 @@ def test_construct_with_options(open_port, sample_train_service, sample_int_file
             assert context.value.code() == grpc.StatusCode.RESOURCE_EXHAUSTED
 
 
+def test_grpc_server_startup(open_port, sample_train_service, sample_int_file):
+    """Make sure that the server can be booted with config options"""
```
joerunde, line 1287: "...booted with config options to enable listening on a socket"?

tests/runtime/test_grpc_server.py
```diff
@@ -1283,6 +1283,25 @@ def test_construct_with_options(open_port, sample_train_service, sample_int_file
...
+    socket_dir = tempfile.TemporaryDirectory()
+    with temp_config(
+        {"runtime": {"grpc": {"unix_socket_path": socket_dir.name + "/grpc.sock"}}},
+        "merge",
+    ):
+        server = RuntimeGRPCServer()
+        with server:
```
joerunde, line 1294: `with RuntimeGRPCServer() as server:`?

joerunde (approved):
super slick! Thanks for the bugfix

tests/runtime/test_grpc_server.py
HonakerM: I didn't know about `tmp_path`! However, when I tried it I started to get socket bind errors. I believe it's because unix sockets are limited to [paths with less than ~100 characters](https://stackoverflow.com/questions/34829600/why-is-the-maximal-path-length-allowed-for-unix-sockets-on-linux-108) and tmp_path gives you a very long dir name.

I did update the `TemporaryDirectory` to use a context manager just to make it a bit cleaner

tests/runtime/test_grpc_server.py
HonakerM: 🤦 thank you for catching me going to fast. I also updated the function name to be more descriptive

tests/runtime/test_grpc_server.py
HonakerM: Done! Didn't even need the `as server`.

tests/runtime/test_grpc_server.py
joerunde: woah, that's wild. TIL!

PR #543, opened 2023-10-23
feat: add info servicer and endpoint with version details

anhuong:
Expose caikit version via API for ease of debugging from users

**If applicable**:

### DIFF
--- a/caikit/config/config.yml
+++ b/caikit/config/config.yml
@@ -195,6 +195,11 @@ runtime:
         output_dir: training_output
         save_with_id: true
 
+    # Version details to retrieve
+    version_info:
+        all_packages: false
+        runtime_image: ""
+
 inference_plugin:
     model_mesh:
         # Model Mesh specific versions
--- a/caikit/config/config.yml
+++ b/caikit/config/config.yml
@@ -198,7 +198,7 @@ runtime:
     # Version details to retrieve
     version_info:
         python_packages:
-            all_packages: false
+            all: false
         runtime_image: ""
 
 inference_plugin:
--- a/caikit/config/config.yml
+++ b/caikit/config/config.yml
@@ -197,7 +197,8 @@ runtime:
 
     # Version details to retrieve
     version_info:
-        all_packages: false
+        python_packages:
+            all_packages: false
         runtime_image: ""
 
 inference_plugin:
--- a/caikit/interfaces/runtime/data_model/info.py
+++ b/caikit/interfaces/runtime/data_model/info.py
@@ -11,6 +11,9 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+"""
+This file contains interfaces to handle information requests
+"""
 
 # Standard
 from typing import Dict
--- a/caikit/interfaces/runtime/data_model/info.py
+++ b/caikit/interfaces/runtime/data_model/info.py
@@ -22,12 +22,12 @@ from typing import Dict
 import alog
 
 # Local
-from caikit.core.data_model import DataObjectBase, dataobject
+from caikit.core.data_model import PACKAGE_COMMON, DataObjectBase, dataobject
 from caikit.core.toolkit.wip_decorator import Action, WipCategory, work_in_progress
 
 log = alog.use_channel("RUNTIMEOPS")
 
-RUNTIME_PACKAGE = "caikit_data_model.runtime"
+RUNTIME_PACKAGE = f"{PACKAGE_COMMON}.runtime"
 
 
 @work_in_progress(action=Action.WARNING, category=WipCategory.BETA)
--- a/caikit/interfaces/runtime/data_model/info.py
+++ b/caikit/interfaces/runtime/data_model/info.py
@@ -23,21 +23,17 @@ import alog
 
 # Local
 from caikit.core.data_model import PACKAGE_COMMON, DataObjectBase, dataobject
-from caikit.core.toolkit.wip_decorator import Action, WipCategory, work_in_progress
 
 log = alog.use_channel("RUNTIMEOPS")
 
 RUNTIME_PACKAGE = f"{PACKAGE_COMMON}.runtime"
 
 
-@work_in_progress(action=Action.WARNING, category=WipCategory.BETA)
 @dataobject(RUNTIME_PACKAGE)
 class RuntimeInfoRequest(DataObjectBase):
     pass
 
 
-# TODO: what version details are in here -- is this meant to be more generic?
-@work_in_progress(action=Action.WARNING, category=WipCategory.BETA)
 @dataobject(RUNTIME_PACKAGE)
 class RuntimeInfoResponse(DataObjectBase):
     runtime_version: Optional[str]
--- a/caikit/interfaces/runtime/data_model/info.py
+++ b/caikit/interfaces/runtime/data_model/info.py
@@ -16,7 +16,7 @@ This file contains interfaces to handle information requests
 """
 
 # Standard
-from typing import Dict
+from typing import Dict, Optional
 
 # First Party
 import alog
@@ -40,4 +40,5 @@ class RuntimeInfoRequest(DataObjectBase):
 @work_in_progress(action=Action.WARNING, category=WipCategory.BETA)
 @dataobject(RUNTIME_PACKAGE)
 class RuntimeInfoResponse(DataObjectBase):
-    version_info: Dict[str, str]
+    runtime_version: Optional[str]
+    python_packages: Dict[str, str]
--- a/caikit/interfaces/runtime/data_model/info.py
+++ b/caikit/interfaces/runtime/data_model/info.py
@@ -19,6 +19,7 @@ This file contains interfaces to handle information requests
 from typing import Dict, Optional
 
 # First Party
+from py_to_proto.dataclass_to_proto import Annotated, FieldNumber
 import alog
 
 # Local
@@ -31,10 +32,10 @@ RUNTIME_PACKAGE = f"{PACKAGE_COMMON}.runtime"
 
 @dataobject(RUNTIME_PACKAGE)
 class RuntimeInfoRequest(DataObjectBase):
-    pass
+    """Empty request for runtime server information"""
 
 
 @dataobject(RUNTIME_PACKAGE)
 class RuntimeInfoResponse(DataObjectBase):
-    runtime_version: Optional[str]
-    python_packages: Dict[str, str]
+    runtime_version: Annotated[Optional[str], FieldNumber(1)]
+    python_packages: Annotated[Dict[str, str], FieldNumber(2)]
--- a/caikit/interfaces/runtime/data_model/runtime_info.py
+++ b/caikit/interfaces/runtime/data_model/runtime_info.py
@@ -29,7 +29,7 @@ RUNTIME_PACKAGE = "caikit_data_model.runtime"
 
 @work_in_progress(action=Action.WARNING, category=WipCategory.BETA)
 @dataobject(RUNTIME_PACKAGE)
-class RuntimeRequest(DataObjectBase):
+class RuntimeInfoRequest(DataObjectBase):
     pass
 
 
--- a/caikit/interfaces/runtime/data_model/runtime_info.py
+++ b/caikit/interfaces/runtime/data_model/runtime_info.py
@@ -1,0 +1,40 @@
+# Copyright The Caikit Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Standard
+from typing import List
+
+# First Party
+import alog
+
+# Local
+from caikit.core.data_model import DataObjectBase, dataobject
+from caikit.core.toolkit.wip_decorator import Action, WipCategory, work_in_progress
+
+log = alog.use_channel("RUNTIMEOPS")
+
+RUNTIME_PACKAGE = "caikit_data_model.runtime"
+
+# @work_in_progress(action=Action.WARNING, category=WipCategory.BETA)
+# @dataobject(RUNTIME_PACKAGE)
+# class ModelPointer(DataObjectBase):
+#     model_id: str
+
+# TODO: what version details are in here -- is this meant to be more generic?
+@work_in_progress(action=Action.WARNING, category=WipCategory.BETA)
+@dataobject(RUNTIME_PACKAGE)
+class RuntimeInfoStatusResponse(DataObjectBase):
+    caikit_version: str
+    runtime_image_version: str
+    caikit_nlp_version: str
--- a/caikit/interfaces/runtime/data_model/runtime_info.py
+++ b/caikit/interfaces/runtime/data_model/runtime_info.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 # Standard
-from typing import List
+from typing import Dict
 
 # First Party
 import alog
@@ -38,3 +38,4 @@ class RuntimeInfoStatusResponse(DataObjectBase):
     caikit_version: str
     runtime_image_version: str
     caikit_nlp_version: str
+    version_info: Dict[str, str]
--- a/caikit/interfaces/runtime/data_model/runtime_info.py
+++ b/caikit/interfaces/runtime/data_model/runtime_info.py
@@ -26,16 +26,15 @@ log = alog.use_channel("RUNTIMEOPS")
 
 RUNTIME_PACKAGE = "caikit_data_model.runtime"
 
-# @work_in_progress(action=Action.WARNING, category=WipCategory.BETA)
-# @dataobject(RUNTIME_PACKAGE)
-# class ModelPointer(DataObjectBase):
-#     model_id: str
+
+@work_in_progress(action=Action.WARNING, category=WipCategory.BETA)
+@dataobject(RUNTIME_PACKAGE)
+class RuntimeRequest(DataObjectBase):
+    pass
+
 
 # TODO: what version details are in here -- is this meant to be more generic?
 @work_in_progress(action=Action.WARNING, category=WipCategory.BETA)
 @dataobject(RUNTIME_PACKAGE)
-class RuntimeInfoStatusResponse(DataObjectBase):
-    caikit_version: str
-    runtime_image_version: str
-    caikit_nlp_version: str
+class RuntimeInfoResponse(DataObjectBase):
     version_info: Dict[str, str]
--- a/caikit/runtime/http_server/http_server.py
+++ b/caikit/runtime/http_server/http_server.py
@@ -66,7 +66,7 @@ from caikit.runtime.service_generation.rpcs import (
 )
 from caikit.runtime.servicers.global_predict_servicer import GlobalPredictServicer
 from caikit.runtime.servicers.global_train_servicer import GlobalTrainServicer
-from caikit.runtime.servicers.info_servicer import InfoServicerImpl
+from caikit.runtime.servicers.info_servicer import InfoServicer
 from caikit.runtime.types.caikit_runtime_exception import CaikitRuntimeException
 
 ## Globals #####################################################################
@@ -107,7 +107,7 @@ MODEL_ID = "model_id"
 HEALTH_ENDPOINT = "/health"
 
 # Endpoint to use for server info
-RUNTIME_INFO_ENDPOINT = "/info/runtime"
+RUNTIME_INFO_ENDPOINT = "/info/version"
 
 # Small dataclass for consolidating TLS files
 @dataclass
@@ -146,7 +146,7 @@ class RuntimeHTTPServer(RuntimeServerBase):
         # Placeholders for global servicers
         self.global_predict_servicer = None
         self.global_train_servicer = None
-        self.runtime_info_servicer = InfoServicerImpl()
+        self.runtime_info_servicer = InfoServicer()
 
         # Set up inference if enabled
         if self.enable_inference:
--- a/caikit/runtime/http_server/http_server.py
+++ b/caikit/runtime/http_server/http_server.py
@@ -67,6 +67,9 @@ from caikit.runtime.service_generation.rpcs import (
 from caikit.runtime.servicers.global_predict_servicer import GlobalPredictServicer
 from caikit.runtime.servicers.global_train_servicer import GlobalTrainServicer
 from caikit.runtime.types.caikit_runtime_exception import CaikitRuntimeException
+from caikit.runtime.servicers.runtime_info_servicer import (
+    RuntimeInfoServicerImpl,
+)
 
 ## Globals #####################################################################
 
@@ -163,8 +166,10 @@ class RuntimeHTTPServer(RuntimeServerBase):
             self._health_check
         )
 
+        # Add runtime info endpoint
+        # self.runtime_info_servicer = RuntimeInfoServicerImpl(self.inference_service)
         self.app.get(RUNTIME_INFO_ENDPOINT, response_class=JSONResponse)(
-            self._runtime_info
+            RuntimeInfoServicerImpl().GetRuntimeInfo().version_info
         )
 
         # Parse TLS configuration
@@ -261,9 +266,9 @@ class RuntimeHTTPServer(RuntimeServerBase):
             time.sleep(1e-3)
         log.info("HTTP Server is running in thread")
 
-    def _runtime_info(self) -> Dict[str, str]:
-        log.debug4("Server info details")
-        return self._get_runtime_info()
+    # def _runtime_info(self) -> Dict[str, str]:
+    #     log.debug4("Server info details")
+    #     return self._get_runtime_info()
 
     ##########
     ## Impl ##
--- a/caikit/runtime/http_server/http_server.py
+++ b/caikit/runtime/http_server/http_server.py
@@ -66,10 +66,8 @@ from caikit.runtime.service_generation.rpcs import (
 )
 from caikit.runtime.servicers.global_predict_servicer import GlobalPredictServicer
 from caikit.runtime.servicers.global_train_servicer import GlobalTrainServicer
+from caikit.runtime.servicers.runtime_info_servicer import RuntimeInfoServicerImpl
 from caikit.runtime.types.caikit_runtime_exception import CaikitRuntimeException
-from caikit.runtime.servicers.runtime_info_servicer import (
-    RuntimeInfoServicerImpl,
-)
 
 ## Globals #####################################################################
 
--- a/caikit/runtime/http_server/http_server.py
+++ b/caikit/runtime/http_server/http_server.py
@@ -146,6 +146,7 @@ class RuntimeHTTPServer(RuntimeServerBase):
         # Placeholders for global servicers
         self.global_predict_servicer = None
         self.global_train_servicer = None
+        self.runtime_info_servicer = RuntimeInfoServicerImpl()
 
         # Set up inference if enabled
         if self.enable_inference:
@@ -165,9 +166,8 @@ class RuntimeHTTPServer(RuntimeServerBase):
         )
 
         # Add runtime info endpoint
-        # self.runtime_info_servicer = RuntimeInfoServicerImpl(self.inference_service)
         self.app.get(RUNTIME_INFO_ENDPOINT, response_class=JSONResponse)(
-            RuntimeInfoServicerImpl().GetRuntimeInfo().version_info
+            self.runtime_info_servicer.get_version_dict
         )
 
         # Parse TLS configuration
@@ -263,10 +263,6 @@ class RuntimeHTTPServer(RuntimeServerBase):
         while not self.server.started:
             time.sleep(1e-3)
         log.info("HTTP Server is running in thread")
-
-    # def _runtime_info(self) -> Dict[str, str]:
-    #     log.debug4("Server info details")
-    #     return self._get_runtime_info()
 
     ##########
     ## Impl ##
--- a/caikit/runtime/http_server/http_server.py
+++ b/caikit/runtime/http_server/http_server.py
@@ -146,7 +146,7 @@ class RuntimeHTTPServer(RuntimeServerBase):
         # Placeholders for global servicers
         self.global_predict_servicer = None
         self.global_train_servicer = None
-        self.runtime_info_servicer = InfoServicer()
+        self.info_servicer = InfoServicer()
 
         # Set up inference if enabled
         if self.enable_inference:
@@ -167,7 +167,7 @@ class RuntimeHTTPServer(RuntimeServerBase):
 
         # Add runtime info endpoint
         self.app.get(RUNTIME_INFO_ENDPOINT, response_class=JSONResponse)(
-            self.runtime_info_servicer.get_version_dict
+            self.info_servicer.get_version_dict
         )
 
         # Parse TLS configuration
--- a/caikit/runtime/http_server/http_server.py
+++ b/caikit/runtime/http_server/http_server.py
@@ -105,6 +105,9 @@ MODEL_ID = "model_id"
 # Endpoint to use for health checks
 HEALTH_ENDPOINT = "/health"
 
+# Endpoint to use for server info
+RUNTIME_INFO_ENDPOINT = "/runtimeInfo"
+
 # Small dataclass for consolidating TLS files
 @dataclass
 class _TlsFiles:
@@ -158,6 +161,10 @@ class RuntimeHTTPServer(RuntimeServerBase):
         # Add the health endpoint
         self.app.get(HEALTH_ENDPOINT, response_class=PlainTextResponse)(
             self._health_check
+        )
+
+        self.app.get(RUNTIME_INFO_ENDPOINT, response_class=JSONResponse)(
+            self._runtime_info
         )
 
         # Parse TLS configuration
@@ -253,6 +260,10 @@ class RuntimeHTTPServer(RuntimeServerBase):
         while not self.server.started:
             time.sleep(1e-3)
         log.info("HTTP Server is running in thread")
+
+    def _runtime_info(self) -> Dict[str, str]:
+        log.debug4("Server info details")
+        return self._get_runtime_info()
 
     ##########
     ## Impl ##
--- a/caikit/runtime/http_server/http_server.py
+++ b/caikit/runtime/http_server/http_server.py
@@ -66,7 +66,7 @@ from caikit.runtime.service_generation.rpcs import (
 )
 from caikit.runtime.servicers.global_predict_servicer import GlobalPredictServicer
 from caikit.runtime.servicers.global_train_servicer import GlobalTrainServicer
-from caikit.runtime.servicers.runtime_info_servicer import RuntimeInfoServicerImpl
+from caikit.runtime.servicers.info_servicer import InfoServicerImpl
 from caikit.runtime.types.caikit_runtime_exception import CaikitRuntimeException
 
 ## Globals #####################################################################
@@ -107,7 +107,7 @@ MODEL_ID = "model_id"
 HEALTH_ENDPOINT = "/health"
 
 # Endpoint to use for server info
-RUNTIME_INFO_ENDPOINT = "/runtimeInfo"
+RUNTIME_INFO_ENDPOINT = "/info/runtime"
 
 # Small dataclass for consolidating TLS files
 @dataclass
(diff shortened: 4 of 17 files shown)


caikit/runtime/server_base.py
```diff
@@ -88,6 +96,23 @@ def _start_metrics_server(cls) -> None:
                 start_http_server(get_config().runtime.metrics.port)
             cls._metrics_server_started = True
 
+    @classmethod
+    def _get_runtime_info(cls) -> Dict[str, str]:
+        """Get information on versions of libraries and server from config"""
+        version_dict = {}
+        # TODO: how to get library version for extensions?
+        # version_dict["caikit_nlp_version"] = pip_version("caikit_nlp")
+        version_dict["caikit_version"] = pip_version("caikit")
```
anhuong, line 105: how do I get the library extension name without explicitly setting caikit_nlp since this should be generic for other extensions as well

tests/fixtures/config/config.yml
```diff
@@ -11,6 +11,10 @@ runtime:
         # Write to log file every N seconds
         log_interval: 5
 
+    versioning:
+        runtime_image: "1.2.3"
+        foo: bar
+
```
anhuong, line 17: I don’t fully understand the flow of how a user sets details in `config.yml` to be populated in the endpoint, what type of info would this use and how does this work for flowing into RPC..?

caikit/runtime/service_factory.py
```diff
@@ -66,6 +67,20 @@
...
 }
 
+RUNTIME_INFO_SERVICE_NAME = "RuntimeInfo"
+RUNTIME_INFO_SERVICE_SPEC = {
+    "service": {
+        "rpcs": [
+            {
+                "name": "GetServerInfo",
+                # TODO: this has no input param...but unsure how to codify that here since this json schema is required? str input_type? - was hitting validation errors in py_to_proto
+                "input_type": TrainingInfoRequest.get_proto_class().DESCRIPTOR.full_name,
```
anhuong, line 77: there is no input param for this endpoint so I wasn’t sure what to put for input_type . I tried deleting the line, having None , or setting an empty string but these all failed py_to_proto validation because I think [this](https://github.com/IBM/py-to-proto/blob/main/py_to_proto/json_to_service.py#L35) shows the expected schema which is a string value for the input_type?

tests/runtime/http_server/test_http_server.py
```diff
@@ -544,6 +544,18 @@ def test_health_check_ok(runtime_http_server):
...
+def test_runtime_info_ok(runtime_http_server):
+    """Make sure the runtime info returns version data"""
+    with TestClient(runtime_http_server.app) as client:
+        response = client.get(http_server.RUNTIME_INFO_ENDPOINT)
+        assert response.status_code == 200
+        json_response = json.loads(response.content.decode(response.default_encoding))
+        assert json_response["foo"] == "bar"
+        assert json_response["runtime_image"] == "1.2.3"
+        assert json_response["caikit_version"] == "0.0.1"
+        assert response.text == "OK"
```
anhuong, line 556: This is the line that fails, added in just to see output, TODO: remove

tests/runtime/test_grpc_server.py
```diff
@@ -953,6 +954,28 @@ def test_runtime_status_ok_response(runtime_grpc_server):
...
+
+    runtime_info_stub = runtime_info_service.stub_class(
+        runtime_grpc_server.make_local_channel()
+    )
+
+    training_info_request = TrainingInfoRequest(training_id=sample_task_model_id)
+    runtime_info_response: RuntimeInfoStatusResponse = (
+        RuntimeInfoStatusResponse.from_proto(
+            runtime_info_stub.GetServerInfo(training_info_request.to_proto())
+        )
```
anhuong, line 971: right way to test a new GRPC endpoint/RPC?

tests/runtime/test_grpc_server.py
```diff
@@ -953,6 +954,28 @@ def test_runtime_status_ok_response(runtime_grpc_server):
     assert actual_response.numericRuntimeVersion == 0
 
 
+# TODO: remove input of TrainingInfoRequest
+def test_runtime_info_ok_response(sample_task_model_id, runtime_grpc_server):
```
anhuong, line 958: Currently failing with error `AttributeError: 'dict' object has no attribute 'GetServerInfo'` on [this line](https://github.com/caikit/caikit/compare/main...anhuong:caikit:server-info?expand=1#diff-b272a1480be61d2c6958dbb793953698ffa95dd8bfae011d5ba9ad3e3a2369ebR151-R152) which makes a lot of sense since the method I attached is not a Class that implements `GetServerInfo` . I see how it’s done for TrainingManagement but what I don’t understand is how HTTP server implements the TrainingManagement portions? Because I thought I had to implement it in `server_base` so that I could reference it within each server like [here](https://github.com/caikit/caikit/compare/main...anhuong:caikit:server-info?expand=1#diff-8c48c03bdb3588a96567eff810341f816e17768a82d89ab2a501e7d8649abc94R266)

caikit/runtime/service_factory.py
anhuong: error message:
```
tests/runtime/conftest.py:157: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/usr/local/Cellar/python@3.11/3.11.3/Frameworks/Python.framework/Versions/3.11/lib/python3.11/contextlib.py:137: in __enter__
    return next(self.gen)
tests/runtime/conftest.py:147: in runtime_grpc_test_server
    with RuntimeGRPCServer(*args, **kwargs) as server:
caikit/runtime/grpc_server.py:125: in __init__
    ServicePackageFactory.get_service_package(
caikit/runtime/service_factory.py:133: in get_service_package
    grpc_service = json_to_service(
.....
  log.debug2("Validating service json")
        if not validate_jtd(json_service_def, SERVICE_JTD_SCHEMA, EXTENDED_TYPE_VALIDATORS):
>           raise ValueError("Invalid service json")
E           ValueError: Invalid service json
```

caikit/runtime/servicers/runtime_info_servicer.py
```diff
@@ -0,0 +1,68 @@
...
+from caikit.config import get_config
+
+log = alog.use_channel("RI-SERVICR-I")
+
+
+class RuntimeInfoServicerImpl:
+    """This class contains the implementation of all of the RPCs that are required to run a
+    service in Model Mesh as a Model-Runtime."""
+
+    def GetRuntimeInfo(self, request, context):
```
anhuong, line 43: What is the `request` and `context` here? It is needed in order for the GRPC tests to run otherwise I get error
```
ERROR    grpc._server:_server.py:580 Exception calling application: RuntimeInfoServicerImpl.GetRuntimeInfo() takes 1 positional argument but 3 were given
  File "/Users/anhuong/github.com/anhuong/caikit/.tox/py/lib/python3.11/site-packages/grpc/_server.py", line 552, in _call_behavior
    response_or_iterator = behavior(argument, context)
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/anhuong/github.com/anhuong/caikit/.tox/py/lib/python3.11/site-packages/py_grpc_prometheus/prometheus_server_interceptor.py", line 118, in new_behavior
    raise e
  File "/Users/anhuong/github.com/anhuong/caikit/.tox/py/lib/python3.11/site-packages/py_grpc_prometheus/prometheus_server_interceptor.py", line 67, in new_behavior
    response_or_iterator = behavior(request_or_iterator, servicer_context)
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/anhuong/github.com/anhuong/caikit/caikit/runtime/interceptors/caikit_runtime_server_wrapper.py", line 149, in safe_rpc_call
    return rpc(request, context)
           ^^^^^^^^^^^^^^^^^^^^^
TypeError: RuntimeInfoServicerImpl.GetRuntimeInfo() takes 1 positional argument but 3 were given
```

but when I run with request and context, HTTP server fails with error: `TypeError: RuntimeInfoServicerImpl.GetRuntimeInfo() missing 2 required positional arguments: 'request' and 'context'`

caikit/runtime/http_server/http_server.py
```diff
@@ -160,6 +166,12 @@ async def validation_exception_handler(_, exc: ResponseValidationError):
             self._health_check
         )
 
+        # Add runtime info endpoint
+        # self.runtime_info_servicer = RuntimeInfoServicerImpl(self.inference_service)
+        self.app.get(RUNTIME_INFO_ENDPOINT, response_class=JSONResponse)(
+            RuntimeInfoServicerImpl().GetRuntimeInfo().version_info
```
anhuong, line 172: Currently runs (when remove request and context params) but fails with error

```
obj = {'runtime_image': '1.2.3', 'caikit_version': '0.0.1', 'foo': 'bar'}

    def _signature_from_callable(obj, *,
                                 follow_wrapper_chains=True,
                                 skip_bound_arg=True,
                                 globals=None,
                                 locals=None,
                                 eval_str=False,
                                 sigcls):
    
        """Private helper function to get signature for arbitrary
        callable objects.
        """
    
        _get_signature_of = functools.partial(_signature_from_callable,
                                    follow_wrapper_chains=follow_wrapper_chains,
                                    skip_bound_arg=skip_bound_arg,
                                    globals=globals,
                                    locals=locals,
                                    sigcls=sigcls,
                                    eval_str=eval_str)
    
        if not callable(obj):
>           raise TypeError('{!r} is not a callable object'.format(obj))
E           TypeError: {'runtime_image': '1.2.3', 'caikit_version': '0.0.1', 'foo': 'bar'} is not a callable object
```
How do I match the `JSONResponse` type with the output from the method?

caikit/runtime/server_base.py
anhuong: --> not possible to get `caikit_nlp` from within caikit, just can get caikit version. Is that sufficient/what is wanted?

tests/runtime/test_grpc_server.py
anhuong: --> moved from `server_base` into own servicer implementation

caikit/runtime/servicers/runtime_info_servicer.py
evaline-ju: `request`(protobuf message defined for the endpoint)  and `context` (grpc context) are both necessary for GRPC endpoint handling

joerunde: @anhuong Since this is a grpc endpoint handler, I probably wouldn't try to call it directly from an http endpoint handler. It should be simple to refactor the main logic out of this and into a separate public method that both the grpc and http endpoint handlers can call.

As for the `context`, it's suuuuuper hard to build a real one of those objects yourself, they're an implementation detail of the grpc server itself. But we do have a test fixture that can give you a fake one: https://github.com/caikit/caikit/blob/main/tests/fixtures/fixtures.py#L42

caikit/runtime/service_factory.py
joerunde: Shoot, yeah there's an `Empty.proto` that would be designed for this but I don't think that's hooked up in `py_to_proto`. I'd try creating a `@dataobject` that has no attributes, and that way it's at least extendable in the future if we need to add anything. I could see taking a list of extra package names that you want to get the versions for.

tests/runtime/test_grpc_server.py
joerunde: @anhuong I don't think the http server supports the `TrainingManagementService`

caikit/runtime/servicers/runtime_info_servicer.py
```diff
@@ -0,0 +1,68 @@
...
+        # TODO: how to get library version for extensions?
+        # version_dict["caikit_nlp_version"] = pip_version("caikit_nlp")
+        version_dict["caikit_version"] = pip_version("caikit")
+        # TODO: why is cls.config != get_config() within this method --> fails with error  AttributeError: type object 'RuntimeHTTPServer' has no attribute 'config'
+        print("DOES VERSIONING EXIST", get_config().runtime.versioning)
+
+        if get_config().runtime.versioning:
+            print("GET_CONFIG", get_config().runtime.versioning)
+            # TODO: how to get library that is being run -- aka get caikit_nlp part dynamically
+            version_dict.update(get_config().runtime.versioning)
```
joerunde, line 55: @anhuong have you looked into importlib.metadata.packages_distributions for the heavy lifting here? or figuring out what's installed?
https://docs.python.org/3.11/library/importlib.metadata.html#module-importlib.metadata

Something like this might get us 90% there:
```
from importlib.metadata import version, packages_distributions

for p in packages_distributions().keys():
    if "caikit" in p:
        version_dict[p] = version(p)
    # or anything extra the user requested:
    if p in request.packages:
        version_dict[p] = version(p)
```

anhuong: Oh yeah we were just looking at that yesterday as well but that seems like a package that is new in 3.10 so we won't be able to use since we support 3.8, 3.9

joerunde: Ah right. I think the `importlib_metadata` package exists for this reason, it looks like it's a full backport of `importlib.metadata`

anhuong: Interesting! How does that work to `import importlib_metadata` for <3.10 and do `import importlib.metadata` for >=3.10?

caikit/interfaces/runtime/data_model/runtime_info.py
```diff
@@ -0,0 +1,40 @@
...
+@work_in_progress(action=Action.WARNING, category=WipCategory.BETA)
+@dataobject(RUNTIME_PACKAGE)
+class RuntimeInfoRequest(DataObjectBase):
+    pass
+
+
+# TODO: what version details are in here -- is this meant to be more generic?
+@work_in_progress(action=Action.WARNING, category=WipCategory.BETA)
+@dataobject(RUNTIME_PACKAGE)
+class RuntimeInfoResponse(DataObjectBase):
```
anhuong, line 39: Thoughts on name RuntimeInfo everywhere? @gabe-l-hart I know you said you wanted it to be more generic so it could be used for other things?

caikit/runtime/servicers/runtime_info_servicer.py
```diff
@@ -0,0 +1,78 @@
...
+        return RuntimeInfoResponse(
+            version_info=versions,
+        ).to_proto()
+
+    def get_version_dict(self) -> Dict[str, str]:
+        """Get information on versions of libraries and server for HTTP"""
+        runtime_info_response = self.get_runtime_info_impl()
+        return runtime_info_response.version_info
+
+    # TODO: fix so can get versions for something like alog --> alchemy_logging
```
anhuong, line 72: How to use `importlib_metadata` for <3.10 and `importlib.metadata` for >=3.10 in order to utilize `packages_distributions()`?`

tests/runtime/http_server/test_http_server.py
```diff
@@ -544,6 +545,80 @@ def test_health_check_ok(runtime_http_server):
...
+        assert "py_to_proto" not in json_response
+
+
+def test_runtime_info_ok_response_all_sys_modules(runtime_http_server):
+    with temp_config(
+        {
+            "runtime": {
+                "versioning": {
+                    "sys_modules": True,
+                    "runtime_image": "1.2.3",
```
anhuong, line 567: This is the config details I have so far, but open to other thoughts @gabe-l-hart ! I know we talked about user requested packages in a list vs all sys modules vs just caikit. Current logic is just caikit libraries if nothing is set and adds in `runtime_image` and all sys.modules if set

caikit/runtime/server_base.py
```diff
@@ -76,6 +76,13 @@ def __init__(self, base_port: int, tls_config_override: Optional[aconfig.Config]
 
         self.training_service = training_service
 
+        # create runtime info service
+        self.runtime_info_service: Optional[
```
anhuong, line 80: This is currently failing linting with error [`too-many-instance-attributes`](https://pylint.readthedocs.io/en/latest/user_guide/messages/refactor/too-many-instance-attributes.html), thoughts? Ignore or fix?

tests/runtime/test_grpc_server.py
```diff
@@ -953,6 +956,103 @@ def test_runtime_status_ok_response(runtime_grpc_server):
...
+        ServicePackageFactory.ServiceType.RUNTIME_INFO,
+    )
+
+    runtime_info_stub = runtime_info_service.stub_class(
+        runtime_grpc_server.make_local_channel()
+    )
+
+    runtime_request = RuntimeInfoRequest()
+    runtime_info_response: RuntimeInfoResponse = RuntimeInfoResponse.from_proto(
+        runtime_info_stub.GetRuntimeInfo(runtime_request.to_proto())
```
anhuong, line 970: This test passes but how does one test this locally? Wwhen I startup the server for samplelib, I see the `/runtimeInfo` endpoint in REST (see image) and I see protos files `caikit_data_model.runtime.runtimeinforesponse.proto` and `caikit_data_model.runtime.runtimerequest.proto` but I don't see `GetRuntimeInfo` within protos

Screenshot 2023-10-24 at 6 15 24 PM

caikit/runtime/servicers/runtime_info_servicer.py
```diff
@@ -0,0 +1,78 @@
...
+                get_config().runtime.versioning
+                and get_config().runtime.versioning.sys_modules
+            ):
+                if len(lib.split(".")) == 1:
+                    version = self.try_lib_version(lib)
+                    if version:
+                        versions[lib] = version
+            # just get caikit versions
+            else:
+                if len(lib.split(".")) == 1 and lib.startswith("caikit"):
```
dtrifiro, line 55: You can easily get `caikit`  version information using `caikit.__version__` or `caikit.__version__tuple__` :

```console
>>> caikit.__version__
'0.22.1.dev10+g11bf1ab'
>>> caikit.__version_tuple__
(0, 22, 1, 'dev10', 'g11bf1ab')
```

anhuong: Good to know! Although we want this to capture all caikit libraries so that can include `caikit_nlp` and `caikit_tgis_backend` as well

dtrifiro: Yeah, I just figured this was helpful, note that both `caikit_nlp` and `caikit_tgis_backend` also have `__version__`

caikit/runtime/servicers/runtime_info_servicer.py
joerunde: @anhuong you could probably import `importlib_metadata` in all cases, or..
- Do a try/catch
```
try:
    from importlib.metadata import version
except ImportError:
    from importlib_metadata import version
```
- grab the python version off sys
```
if sys.version_info < (3, 11):
    from importlib_metadata import version
else:
    from importlib.metadata import version
```

anhuong: Doing `importlib_metadata` didn't work in py311 for me but the alternative is great!

caikit/runtime/servicers/runtime_info_servicer.py
```diff
@@ -0,0 +1,79 @@
...
+# Have pylint ignore Class XXXX has no YYYY member so that we can use gRPC enums.
+# pylint: disable=E1101
+
+# Standard
+from typing import Dict
+import sys
+if sys.version_info < (3, 10):
+    import importlib_metadata as importlib_meta
+else:
+    import importlib.metadata as importlib_meta
```
anhuong, line 24: This is failing with error `ModuleNotFoundError: No module named 'importlib_metadata'` in tox but when I run it locally it works...
```
$ python3.9
Python 3.9.16 (main, Dec  7 2022, 10:15:43) 
[Clang 14.0.0 (clang-1400.0.29.202)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import importlib_metadata
>>> importlib_metadata.packages_distributions()
{'tensorflow_text': ['tensorflow-text']
```

and should be compatible with py3.8 as seen from [here](https://github.com/python/importlib_metadata#compatibility)

tests/runtime/http_server/test_http_server.py
gabe-l-hart: Ok, makes sense! A couple of thoughts:

1. We need to add default values to `caikit/config/config.yml`
2. I think it might be good to change `"versioning"` to `"version_info"` to avoid any confusion about things like "how to handle multiple service interface versions" in the future
3. I still see `"foo": "bar"` in here which I think should get removed?
4. I think we should change `"sys_modules"` to `"all_packages"` or something similar to avoid confusion with "standard modules"

tests/runtime/http_server/test_http_server.py
```diff
@@ -544,6 +545,78 @@ def test_health_check_ok(runtime_http_server):
...
+
+            json_response = json.loads(
+                response.content.decode(response.default_encoding)
+            )
+            assert json_response["runtime_image"] == "1.2.3"
+            assert "caikit" in json_response
+            # dependent libraries versions added
+            assert "alog" in json_response
+            assert "py_to_proto" in json_response
+            # additional config values ignored
```
gabe-l-hart, line 586: Once this is done through `config.yml`, I don't think we need to explicitly test this since that's pretty much standard across all usages of `caikit.config`.

tests/runtime/http_server/test_http_server.py
```diff
@@ -544,6 +545,78 @@ def test_health_check_ok(runtime_http_server):
...
+            assert "foo" not in json_response
+
+@patch.dict(
```
gabe-l-hart, line 589: I also think we don't need specific tests for env var overrides on this once it's part of the standard `caikit.config` since that's standard and assumed to be tested in `aconfig` iteself.

gabe-l-hart: (but also, nice patch for `os.environ`!)

tests/runtime/test_grpc_server.py
gabe-l-hart: Is the issue that the `service`/`rpc` is not getting generated with `dump_services`? If this test is passing I think it means that the servicer is getting bound correctly. Do you see it with reflection using `grpcui`?

tests/runtime/test_grpc_server.py
```diff
@@ -953,6 +956,102 @@ def test_runtime_status_ok_response(runtime_grpc_server):
...
+        assert "alog" in runtime_info_response.version_info
+        assert "py_to_proto" in runtime_info_response.version_info
+        # additional config values ignored
+        assert "foo" not in runtime_info_response.version_info
+
+
+@patch.dict(
+    os.environ,
+    {
+        "RUNTIME_VERSIONING_SYS_MODULES": "false",
```
gabe-l-hart, line 1019: Same point here about not needing to test env var overrides

gabe-l-hart (changes requested):
Thanks Anh! This looks really good. The main request I have is to remove the `runtime` scoping from all of the service names so that we can use this as the place for additional `info` RPCs (e.g. `model-info`). A few other little NITs, and we need to add the config to `caikit/config/config.yml`.

tests/runtime/http_server/test_http_server.py
anhuong: 1,2,3 is good feedback, thanks! `foo: bar` is in as a test example that if other values are added in, they are not included in the final versioning response/dict.

tests/runtime/test_grpc_server.py
anhuong: > Is the issue that the service/rpc is not getting generated with dump_services
Yes I believe so, there is only `samplelibservice.proto` and `samplelibtrainingservice.proto` but you're right that with `grpcui` it does appear. Should there be a proto file for samplelibruntimeinfoservice?

Screenshot 2023-10-25 at 2 28 14 PM

caikit/runtime/service_factory.py
```diff
@@ -66,6 +68,19 @@
     }
 }
 
+RUNTIME_INFO_SERVICE_NAME = "RuntimeInfo"
+RUNTIME_INFO_SERVICE_SPEC = {
```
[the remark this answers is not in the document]
anhuong: Gotcha, thanks, appreciate the feedback on where it makes sense to be more specific with runtimeInfo and where we want to generalize for reuse

tests/runtime/test_grpc_server.py
joerunde: As defined I don't think the service would have `sample_lib` prepended to the name- it should just be named `runtimeinfoservice.proto`. That's probably what we want- I'd imagine this service definition should stay constant no matter which libraries you're serving in the runtime

caikit/runtime/servicers/runtime_info_servicer.py
```diff
@@ -0,0 +1,79 @@
...
+import alog
+
+# Local
+from caikit.config import get_config
+from caikit.interfaces.runtime.data_model import RuntimeInfoResponse
+
+log = alog.use_channel("RI-SERVICR-I")
+
+
+class RuntimeInfoServicerImpl:
```
gkumbhat, line 36: Why not just `InfoServicer` ?

caikit/runtime/servicers/runtime_info_servicer.py
```diff
@@ -0,0 +1,79 @@
...
+    """This class contains the implementation of all of the RPCs that are required to run a
```
gkumbhat, line 37: nit: the docstrict isn't accurate ?

caikit/runtime/servicers/runtime_info_servicer.py
gkumbhat: `__version__` seems cleaner way.. @dtrifiro can it also get the version if the package is installed via git commit ?

caikit/runtime/http_server/http_server.py
```diff
@@ -105,6 +106,9 @@
 # Endpoint to use for health checks
 HEALTH_ENDPOINT = "/health"
 
+# Endpoint to use for server info
+RUNTIME_INFO_ENDPOINT = "/runtimeInfo"
```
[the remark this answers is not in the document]
anhuong: Would making this configurable be something that can be added to `config.yml`?

tests/runtime/test_grpc_server.py
anhuong: that makes sense but there isn't a `runtimeinfoservice.proto`...
```
$ ls protos | grep service
samplelibservice.proto
samplelibtrainingservice.proto
```

caikit/runtime/servicers/runtime_info_servicer.py
anhuong: > you could probably import importlib_metadata in all cases, or..

i see what you're saying now, took a little while but it finally clicked 😅

caikit/runtime/servicers/runtime_info_servicer.py
dtrifiro: Yep, that's the main advantage in using `setuptools_scm`: version is set at build time by using git tags.

gabe-l-hart (changes requested):
A few more NITs and some changes to not return `null` values in the info dict. Very close now!

caikit/config/config.yml
```diff
@@ -195,6 +195,12 @@ runtime:
         output_dir: training_output
         save_with_id: true
 
+    # Version details to retrieve
+    version_info:
+        python_packages:
+            all_packages: false
```
gabe-l-hart, line 201: I'm the worst, but...

```suggestion
            all: false
```

caikit/interfaces/runtime/data_model/info.py
```diff
@@ -0,0 +1,43 @@
...
+# Local
+from caikit.core.data_model import PACKAGE_COMMON, DataObjectBase, dataobject
+from caikit.core.toolkit.wip_decorator import Action, WipCategory, work_in_progress
+
+log = alog.use_channel("RUNTIMEOPS")
+
+RUNTIME_PACKAGE = f"{PACKAGE_COMMON}.runtime"
+
+
+@work_in_progress(action=Action.WARNING, category=WipCategory.BETA)
```
gabe-l-hart, line 33: No need for WIP decorator!

caikit/interfaces/runtime/data_model/info.py
```diff
@@ -0,0 +1,43 @@
...
+@dataobject(RUNTIME_PACKAGE)
+class RuntimeInfoRequest(DataObjectBase):
+    pass
+
+
+# TODO: what version details are in here -- is this meant to be more generic?
+@work_in_progress(action=Action.WARNING, category=WipCategory.BETA)
+@dataobject(RUNTIME_PACKAGE)
+class RuntimeInfoResponse(DataObjectBase):
+    version_info: Dict[str, str]
```
gabe-l-hart, line 43: I think you flipped what I meant by suggesting that we nest python_packages! I was suggesting that on the output, so something like:

```json
{
  "runtime_version": "1.2.3",
  "python_packages": {
    "foo": "2.3.4"
  }
}
```

caikit/runtime/servicers/info_servicer.py
```diff
@@ -0,0 +1,84 @@
...
+    ) -> RuntimeInfoResponse:
+        """Get information on versions of libraries and server for GRPC"""
+        return self.get_runtime_info_impl()
+
+    def get_runtime_info_impl(self) -> RuntimeInfoResponse:
+        """Get information on versions of libraries and server from config"""
+        versions = {}
+        version_info = get_config().runtime.version_info or {}
+        if version_info.get("python_packages"):
+            all_packages = version_info.get("python_packages").get("all_packages")
```
gabe-l-hart, line 52: This is actually unsafe since `all_packages` may not be defined below

```suggestion
        all_packages = (version_info.get("python_packages") or {}).get("all_packages")
```

tests/runtime/test_grpc_server.py
gabe-l-hart: I suspect this is because of how [dump_grpc_services](https://github.com/caikit/caikit/blob/main/caikit/runtime/dump_services.py#L32) is implemented. I think you need to add this new service there!

gabe-l-hart (changes requested):
A few more NITs, and I think you need to update `dump_grpc_services`

tests/runtime/test_grpc_server.py
anhuong: ah-ha that was it! Now the protos are `info.proto`, `caikit_data_model.common.runtime.runtimeinforequest.proto`, and `caikit_data_model.common.runtime.runtimeinforesponse.proto`. 

With info.proto looking like:
```
service Info {
  rpc GetRuntimeInfo(caikit_data_model.common.runtime.RuntimeInfoRequest) returns (caikit_data_model.common.runtime.RuntimeInfoResponse);
}
```

tests/runtime/http_server/test_http_server.py
```diff
@@ -544,6 +545,48 @@ def test_health_check_ok(runtime_http_server):
...
+def test_runtime_info_ok(runtime_http_server):
+    """Make sure the runtime info returns version data"""
+    with TestClient(runtime_http_server.app) as client:
+        response = client.get(http_server.RUNTIME_INFO_ENDPOINT)
+        assert response.status_code == 200
+
+        json_response = json.loads(response.content.decode(response.default_encoding))
+        assert "caikit" in json_response["python_packages"]
+        # runtime_version not added if not set
+        assert json_response["runtime_version"] == ""
```
anhuong, line 557: Note the output now looks like
```json
{
  "python_packages": {
    "caikit": "0.0.1"
  },
  "runtime_version": ""
}
```

caikit/runtime/servicers/info_servicer.py
```diff
@@ -0,0 +1,85 @@
...
+            runtime_version=runtime_image,
+        ).to_proto()
+
+    def get_version_dict(self) -> Dict[str, Union[str, Dict]]:
+        """Get information on versions of libraries and server for HTTP"""
+        versions = {}
+        runtime_info_response = self.get_runtime_info_impl()
+        versions["python_packages"] = runtime_info_response.python_packages
+        versions["runtime_version"] = runtime_info_response.runtime_version
+        return versions
```
anhuong, line 78: I _think_ this is the correct things to do to return the dictionary for HTTP? THe outputs from GRPC and HTTP are the same with this so that looks good

caikit/interfaces/runtime/data_model/info.py
```diff
@@ -0,0 +1,40 @@
...
+
+@dataobject(RUNTIME_PACKAGE)
+class RuntimeInfoRequest(DataObjectBase):
+    pass
+
+
+@dataobject(RUNTIME_PACKAGE)
+class RuntimeInfoResponse(DataObjectBase):
+    runtime_version: Optional[str]
+    python_packages: Dict[str, str]
```
gabe-l-hart, line 40: Crap, One more thing: We should explicitly set `FieldNumber`s on these:

```py
from py_to_proto.dataclass_to_proto import Annotated, FieldNumber
...

class RuntimeInfoResponse(DataObjectBase):
    runtime_version: Annotated[Optional[str], FieldNumber(1)]
    python_packages: Annotated[Dict[str, str], FieldNumber(2)]
```

caikit/interfaces/runtime/data_model/info.py
```diff
@@ -0,0 +1,40 @@
...
+from caikit.core.data_model import PACKAGE_COMMON, DataObjectBase, dataobject
+
+log = alog.use_channel("RUNTIMEOPS")
+
+RUNTIME_PACKAGE = f"{PACKAGE_COMMON}.runtime"
+
+
+@dataobject(RUNTIME_PACKAGE)
+class RuntimeInfoRequest(DataObjectBase):
+    pass
```
gabe-l-hart, line 34: If you put a docstring here, you don't need `pass`

```suggestion
   """Empty request for runtime server information"""
```

caikit/runtime/servicers/info_servicer.py
```diff
@@ -0,0 +1,85 @@
...
+            if all_packages:
+                lib_version = self.try_lib_version(dist_names[0])
+                if lib_version:
+                    python_packages[lib] = lib_version
+            # just get caikit versions
+            else:
+                if len(lib.split(".")) == 1 and lib.startswith("caikit"):
+                    version = self.try_lib_version(dist_names[0])
+                    if version:
+                        python_packages[lib] = version
```
gabe-l-hart, line 63:
```suggestion
            elif len(lib.split(".")) == 1 and lib.startswith("caikit"):
                  version = self.try_lib_version(dist_names[0])
                  if version:
                      python_packages[lib] = version
```

caikit/runtime/servicers/info_servicer.py
gabe-l-hart, line 78: I think you can boil all of this down to:

```suggestion
return self.get_runtime_info_impl().to_dict()
```

tests/runtime/test_grpc_server.py
gabe-l-hart: Yep, that would be the place to change it

gabe-l-hart (changes requested):
Sooo close! Just a couple more little tweaks

caikit/interfaces/runtime/data_model/info.py
anhuong: I noticed the `Annotated` on other data models and was wondering, I'm guessing these should be set for all responses?

gkumbhat: yep, those would need to be for all responses.

gabe-l-hart: It's not strictly required, but it guarantees that the messages will not accidentally change their proto representations by changing field orders. We require that all messages in `caikit` itself have them.

caikit/runtime/servicers/info_servicer.py
```diff
@@ -0,0 +1,80 @@
...
+        for lib, dist_names in importlib_metadata.packages_distributions().items():
+            if all_packages:
+                lib_version = self.try_lib_version(dist_names[0])
+                if lib_version:
+                    python_packages[lib] = lib_version
+            # just get caikit versions
+            elif len(lib.split(".")) == 1 and lib.startswith("caikit"):
+                version = self.try_lib_version(dist_names[0])
+                if version:
+                    python_packages[lib] = version
```
gabe-l-hart, line 62: Oooh, I just remembered how to code golf this down! The walrus operator (`:=`)

```suggestion
            if (
                (all_packages or (len(lib.split(".")) == 1 and lib.startswith("caikit")))
                and version := self.try_lib_version(dist_names[0])
            ):
                python_packages[lib] = version
```

gabe-l-hart (approved):
Looks great Anh! Thanks for all the iteration. I have one more fun pythonism suggestion, but it's purely syntactic sugar, so you can take it or leave it. This is ready to merge any time!

caikit/runtime/servicers/info_servicer.py
anhuong: Very nice!

anhuong:
I don't have merge powers so if @gabe-l-hart or @gkumbhat you can merge this in, that would be much appreciated!

evaline-ju:
@anhuong Seems to be some test syntax errors, otherwise can help merge

gabe-l-hart (approved):
Ship it!

anhuong:
Thanks @evaline-ju ! Fixed the issue and thanks Gabe for rerunning the build that failed

PR #546, opened 2023-10-26
Output target

joerunde:

This PR introduces the concept of the model saver plugins as the output counterpart to #520, to allow for configurable output paths for models to be saved to, for external library extensions.

Note: S3 tests have been removed here to be tested separately in implementations that require S3

Addresses remaining portions of: https://github.com/caikit/caikit/issues/517

**If applicable**:


caikit/config/config.yml
```diff
@@ -195,6 +195,12 @@ runtime:
         output_dir: training_output
         save_with_id: true
 
+        model_saver_plugins:
```
gabe-l-hart, line 198: We should think about whether this lives outside of `runtime` near `data_streams`

caikit/core/model_management/model_saver.py
```diff
@@ -0,0 +1,117 @@
...
+import typing
+
+# Local
+from ..modules import ModuleBase
+from caikit.interfaces.common.data_model.stream_sources import PathReference
+
+T = typing.TypeVar("T")
+
+
+class ModelSaver(typing.Generic[T]):
```
gabe-l-hart, line 28:
```suggestion
class ModelSaver(typing.Generic[T], abc.ABC):
```

caikit/core/model_management/model_saver.py
```diff
@@ -0,0 +1,117 @@
...
+    """Generic-typed, abstract base for model saver"""
+
+    @abc.abstractmethod
+    def save_model(
+        self, model: ModuleBase, model_name: str, training_id: Optional[str]
+    ) -> typing.Any:
```
gabe-l-hart, line 34: Potentially this is just `T` (i.e. saver for `str` path, saver for `s3Path`, etc...)

caikit/core/model_management/model_saver.py
```diff
@@ -0,0 +1,117 @@
...
+            final_path_parts.append(model_name)
+
+        return os.path.join(*final_path_parts)
+
+
+# Extend ModelSaver with a concrete message type
+class LocalPathModelSaver(ModelSaver[PathReference]):
+    """Holds the actual impl for saving the model"""
+
+    # TODO: Do we even keep the base `training_output_dir`?
```
gabe-l-hart, line 96: Looks like this got TODONE!

caikit/core/model_management/model_saver.py
```diff
@@ -0,0 +1,117 @@
...
+            training_id (str | None): The globally unique id tracking the training run that
+                created this model
+
+        Returns:
+            Any: Some representation of where the model was saved. This could be a path on disk.
+                    None if not applicable
+        """
+
+    @classmethod
+    def _save_path_with_id(
```
gabe-l-hart, line 65: NIT: This feels like a name for a mutating function rather than a `get` function?

caikit/core/model_management/model_saver.py
```diff
@@ -0,0 +1,117 @@
...
+
+        Returns:
+            Any: Some representation of where the model was saved. This could be a path on disk.
+
+        Raises:
+            Any appropriate exception if saving the model fails
+        """
+
+    @abc.abstractmethod
+    def save_path(self, model_name: str, training_id: Optional[str]) -> typing.Any:
```
gabe-l-hart, line 51: I think all of this "look up the save path" was added thinking we _might_ need it, but in reality we don't actually use it. Since the typing semantics are gorpy anyway, let's remove `save_path` and the return value from `save_model`. We can add this back in if needed down the road.

caikit/core/model_management/model_saver.py
```diff
@@ -0,0 +1,117 @@
...
+
+T = typing.TypeVar("T")
+
+
+class ModelSaver(typing.Generic[T]):
+    """Generic-typed, abstract base for model saver"""
+
+    @abc.abstractmethod
+    def save_model(
+        self, model: ModuleBase, model_name: str, training_id: Optional[str]
```
gabe-l-hart, line 33: I think we can go one of two ways here:

1. Move `T` into being an arg and have the lifecycle of a `Saver` be the duration of the server's life
2. Move `model_name`/`training_id` into `__init__` and explicitly state that a `Saver` lives for the duration of saving a single model

caikit/core/model_management/model_trainer_base.py
```diff
@@ -61,27 +60,26 @@ def __init__(
             self,
             trainer_name: str,
             training_id: str,
-            save_with_id: bool,
```
gabe-l-hart, line 64: Definitely a breaking API change because it breaks existing trainer implementations

caikit/core/model_management/model_trainer_base.py
```diff
@@ -61,27 +60,26 @@ def __init__(
...
             )
-            self._save_path = self.__class__._save_path_with_id(
-                save_path,
-                save_with_id,
-                self._id,
-                model_name,
-            )
+            self._saver = saver
+            # TODO: figure out what to do with this
+            if saver:
```
gabe-l-hart, line 76: I think, per discussion above, we can probably just remove this

caikit/core/model_management/model_trainer_base.py
```diff
@@ -90,11 +88,24 @@ def id(self) -> str:
...
+            the trained model somewhere
+            """
+            return self._saver
+
         @property
         def save_path(self) -> Optional[str]:
             """If created with a save path, the future must expose it, including
             any injected training id
             """
+            # TODO: ????
```
gabe-l-hart, line 108: Get rid of it!

caikit/interfaces/common/data_model/stream_sources.py
```diff
@@ -30,6 +30,13 @@ class FileReference(DataObjectBase):
     filename: Annotated[str, FieldNumber(1)]
 
 
+@dataobject(PACKAGE_COMMON)
+class PathReference(DataObjectBase):
```
gabe-l-hart, line 34: This is a bit confusing because it's not really for a `DataStreamSource`

caikit/core/model_management/model_saver.py
```diff
@@ -0,0 +1,117 @@
...
+"""This file defines the abstraction for handling the output of training a model"""
+# Standard
+from typing import Optional
+import abc
+import os
+import typing
+
+# Local
+from ..modules import ModuleBase
+from caikit.interfaces.common.data_model.stream_sources import PathReference
```
gabe-l-hart, line 23: We can't do this! `core` is not allowed to import `interfaces`

caikit/interfaces/common/data_model/stream_sources.py
gabe-l-hart: The pattern that would follow how we do things currently would be to put this in `core.data_model.<somefile>.py`

caikit/runtime/service_factory.py
```diff
@@ -42,6 +42,10 @@
     TrainingStatusResponse,
 )
 from caikit.runtime import service_generation
+from caikit.runtime.service_generation.output_target import (
+    PluginFactory as OutputTargetPluginFactory,
```
gabe-l-hart, line 46: I'd probably be in favor of calling this `OutputTargetPluginFactory` in the definition (and maybe fixing the corresponding one in `DataStreamSource` too)

caikit/runtime/service_factory.py
```diff
@@ -115,6 +119,7 @@ def get_service_package(
                 concrete Servicer implementation to a protobufs Service and grpc Server
         """
         if service_type == cls.ServiceType.TRAINING_MANAGEMENT:
+            # TODO: should this use service_generation.get_runtime_service_package()?
```
gabe-l-hart, line 122: We're breaking the API, so yes, we should NOT be hard-coding this for the management service

caikit/runtime/service_generation/output_target.py
```diff
@@ -0,0 +1,258 @@
...
+        """The name of the field that this plugin will use in the output target oneof"""
+        return self.get_output_target_type().__name__.lower()
+
+    # Default impls
+    def get_output_target_type(self) -> Type[DataBase]:
+        """Return the output target message expected by your model saver"""
+        # Gorp here to get the generic type T of the target class
+        # Could just be overriden directly by your plugin
+        bases = self.get_model_saver_class().__orig_bases__
+        output_target_base = [b for b in bases if typing.get_origin(b) == ModelSaver][0]
```
gabe-l-hart, line 83: Any chance of an index error here?

caikit/core/model_management/model_saver.py
```diff
@@ -0,0 +1,117 @@
...
+
+        if model_name and model_name not in save_path:
+            # Don't inject model name if it's already in the path
+            final_path_parts.append(model_name)
+
+        return os.path.join(*final_path_parts)
+
+
+# Extend ModelSaver with a concrete message type
+class LocalPathModelSaver(ModelSaver[PathReference]):
```
gabe-l-hart, line 93: I think it might be cleanest if this just used `str` instead of `PathReference` and the down in `runtime` we wrap the `str` as needed in the plugin

caikit/runtime/servicers/global_train_servicer.py
```diff
@@ -214,23 +215,27 @@ def run_training_job(
...
-            model_path: S3Path = request_data_model.output_path
-        else:
-            # Otherwise, use either:
+        # Create the model_saver to handle saving the training output
+        model_saver = (
+            request_data_model.output_target.make_model_saver()
+            if request_data_model.output_target
+            else None
+        )
+        if model_saver is None:
```
gabe-l-hart, line 224:
```suggestion
        if model_saver := (
            request_data_model.output_target.make_model_saver()
            if request_data_model.output_target
            else None
        ) is None:
```

pyproject.toml
```diff
@@ -21,7 +21,7 @@ dependencies = [
     "munch>=2.5.0,<5.0",
     "numpy>=1.22.2,<2",
     "protobuf>=3.19.0,<5",
-    "py-to-proto>=0.5.0,<0.6.0,!=0.2.1",
+    "py-to-proto>=0.5.2,<0.6.0,!=0.2.1",
```
gabe-l-hart, line 24:
```suggestion
    "py-to-proto>=0.5.2,<0.6.0",
```

gabe-l-hart (changes requested):
Really nice stuff! I think we need to sort out the boundary between `core`, `interfaces`, and `runtime` a little bit, and a bunch of little NITs, but otherwise I think this is right on target (pun intended 😉)

caikit/runtime/service_factory.py
```diff
@@ -187,7 +189,12 @@ def get_service_package(
             )
             service_name = f"{ai_domain_name}Service"
         else:  # service_type == cls.ServiceType.TRAINING
-            rpc_list = service_generation.create_training_rpcs(clean_modules)
+            output_target_type = make_output_target_message(
+                # plugin_factory=OutputTargetPluginFactory
```
gabe-l-hart, line 193: What's this about?

caikit/runtime/service_generation/output_target.py
```diff
@@ -0,0 +1,64 @@
...
+        model_savers_config = get_config().model_management.savers
+
+    annotation_list = []
+
+    field_number = 1
+    for saver_name in model_savers_config:
+        saver_builder = caikit.core.MODEL_MANAGER.get_saver_builder(saver_name)
+        output_target_type = saver_builder.output_target_type()
+        annotation_list.append(
+            Annotated[output_target_type, OneofField(saver_name), field_number]
```
gabe-l-hart, line 52: Don't you need `FieldNumber(field_number)` here?

pyproject.toml
```diff
@@ -22,7 +22,7 @@ dependencies = [
     "munch>=2.5.0,<5.0",
     "numpy>=1.22.2,<2",
     "protobuf>=3.19.0,<5",
-    "py-to-proto>=0.5.0,<0.6.0,!=0.2.1",
+    "py-to-proto>=0.5.2,<0.6.0",
```
gabe-l-hart, line 25: Was something released recently to make us want to have this lower bound bump? Unless we NEED a feature in `0.5.2`, I think we should try to keep the range as open as possible.

tests/core/test_model_manager.py
```diff
@@ -675,8 +681,9 @@ def test_train_by_module_class(reset_globals):
     with setup_test_trainer():
         train_future = caikit.train(SampleModule, DataStream.from_iterable([]))
         assert isinstance(train_future, ModelTrainerBase.ModelFutureBase)
-        loaded_model = train_future.load()
-        assert isinstance(loaded_model, SampleModule)
+        # ensure model actually trained
+        trained_model = train_future._trained_model
```
gabe-l-hart, line 685: I wonder if we should make a non-interface function on the `LocalModelTrainer` for this? We could then `assert isinstance(trainer, LocalModelTrainer)` and safely call a non-impl function (instead of relying on this hidden property)

tests/runtime/conftest.py
```diff
@@ -359,18 +366,31 @@ def register_trained_model(
...
+    training_future.wait()
+    # Make sure it passed
+    assert training_future.get_info().status == TrainingStatus.COMPLETED
+    # And that it was saved
+    assert training_future.saver is not None
+    # ...And that it was saved to disk
+    assert isinstance(training_future.saver, LocalModelSaver)
+
+    # Get the save path
+    save_path = training_future.saver._get_save_path(
```
gabe-l-hart, line 380: Nice, this is cleaner!

gabe-l-hart (commented):
A couple of NITs and a little bit of lingering confusion on `saver builder` vs `saver`, but I think this is pretty darn close. I really like the move to have this in the core under `model_management`!

caikit/core/model_management/factories.py
```diff
@@ -38,3 +38,10 @@
 # location.
 model_initializer_factory = ImportableFactory("ModelInitializer")
 model_initializer_factory.register(LocalModelInitializer)
+
+# Model saver builder factory. A model saver builder is responsible for taking
+# an output_target and returning a valid model saver for that output type.
```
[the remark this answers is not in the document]
joerunde: It's... confusing but consistent!

caikit/core/model_management/local_model_trainer.py
```diff
@@ -111,10 +109,10 @@ def __init__(
                 )
                 # If training in a subprocess without a save path, the model
                 # will be unreachable once trained!
-                if not self.save_path:
+                if not self.saver:
```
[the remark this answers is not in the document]
joerunde: That, and if we wanted to auto-create a `saver` in `caikit.train` then we would need to pass in an `output_target` instead/as-well to create the saver with. Then we would probably decide we want to support both parameters to allow either the auto-creation or an explicit saver, and then we'd have to check that both aren't set, and and and...

Figured it's simpler to just accept the saver, but we could talk about it

caikit/core/model_management/model_saver_base.py
```diff
@@ -0,0 +1,139 @@
...
+    """Generic-typed, abstract base for model saver.
+
+    Model savers are lifetime-scoped to saving a single model to a single target.
+    They encapsulate the target information (where the model should be saved) so that the
+    save_model API only needs to be passed the model itself and some identifiers.
+    """
+
+    def __new__(cls, *args, **kwargs):
+        origins = [typing.get_origin(b) for b in cls.__orig_bases__]
+        error.value_check(
```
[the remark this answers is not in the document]
joerunde: Yep!

Maybe a bit wasteful since it checks on _every_ instance construction, but it works!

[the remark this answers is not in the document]
joerunde: (and no I don't want to write a meta-class do do this check)

caikit/core/model_management/model_saver_base.py
```diff
@@ -0,0 +1,139 @@
...
+
+error = error_handler.get(log)
+
+OutputTargetType = typing.TypeVar("OutputTargetType")
+
+
+class ModelSaverBase(typing.Generic[OutputTargetType], abc.ABC):
+    """Generic-typed, abstract base for model saver.
+
+    Model savers are lifetime-scoped to saving a single model to a single target.
```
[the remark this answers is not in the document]
joerunde: Yeah we can talk about names / if we want this to be a class + closure combo instead of two classes

caikit/core/model_manager.py
```diff
@@ -423,6 +421,58 @@ def get_initializer(
...
+        Otherwise, look for a saver that supports the supplied OutputTargetType.
+
+        Args:
+            output_target (OutputTargetType): An output target. There must be a
+                configured ModelSaverBase[OutputTargetType].
+
+        Returns:
+            ModelSaverBase[OutputTargetType]: A configured model saver
+        """
+        if saver_name is not None:
```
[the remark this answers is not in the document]
joerunde: Yes, though the separate `get_saver_builder` is what the runtime uses at service generation time to grab all the available saver builders to construct the `output_target` message 🤔

caikit/runtime/service_factory.py
joerunde: ah, just a leftover bit I meant to delete. But also... do we want to pass in some extra config here so users can re-name / re-number these oneof fields?

caikit/runtime/service_generation/data_stream_source.py
```diff
@@ -374,6 +374,9 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self._plugins = None
 
+    # Mostly duplicate code in output_target.py with different config
+    # could be refactored when n=3
+    # pylint: disable=duplicate-code
```
joerunde, line 379: oh TODO I think these can be deleted now

caikit/runtime/service_generation/output_target.py
joerunde: sure do!

pyproject.toml
joerunde: Yeah- we need 0.5.2 to support singular oneofs: https://github.com/IBM/py-to-proto/releases/tag/0.5.2

tests/core/test_model_manager.py
joerunde: I don't think this is the `LocalModelTrainer`, it's a test class

joerunde: `TestModelFuture` in helpers.py

caikit/config/config.yml
```diff
@@ -191,10 +198,18 @@ runtime:
 
     # Training configuration for server
     training:
-        # The directory to save trained models in
+        # These two configs are used when no `output_target` is sent with the training request
+        # They are used to construct a `Local` model saver + output target
         output_dir: training_output
         save_with_id: true
 
+        # TODO: delete?
```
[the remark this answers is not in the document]
joerunde: Ah- yeah it's unused now.

But I do wonder if we need some more configurability in `runtime` somewhere, otherwise the name and numbering of the output target fields are just gonna match what's in `model_management.savers`

gabe-l-hart:
@joerunde Let's revive this one and get it over the line if possible. It would be great to put this one to bed

joerunde:
Closing for now- can revive if the flexibility on save targets is required by more users

PR #551, opened 2023-10-30
:technologist: Add start_index in response

gkumbhat:
### Changes
- Add `start_index` as additional field to allow easy use of processed classification results instead of referencing it from "end" index with `processed_index`


evaline-ju (approved):
LGTM

Issue #548, opened 2023-10-26
Unified Readiness Probe


gabe-l-hart:
## Is your feature request related to a problem? Please describe.

When deployed in `kubernetes`, `caikit.runtime` needs a consistent way to perform [liveness and readiness proves](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/) for the combination of servers that are configured to run.

## Describe the solution you'd like

These probes will require intimate knowledge of the `caikit.runtime` configuration and should therefore be part of `caikit.runtime` itself. The proposed solution is to add a new executable module `caikit.runtime.readiness_probe` which parses the current configuration to match the config of the running server and makes `http` and/or `grpc` calls to the respective health endpoints.

## Describe alternatives you've considered

* We can let this be a user problem, but users will likely stumble over how to implement this consistently
* We could support using the existing `/health` endpoint for the `http` server, but `kubernetes` cannot perform a `grpc` call as a health check

## Additional context

* Health probes will need to work in the presence of TLS and mTLS configured servers. For mTLS, this means that the probe process will need a key/cert pair signed by a client cert that the server trusts which may be provided externally.

PR #552, opened 2023-10-30
Health probe command 548

gabe-l-hart:

Addresses #548 

This PR adds `caikit.runtime.health_probe` which runs a health probe against one or both runtime servers. It uses the existing environment config to match the running server.

**If applicable**:

### DIFF
--- a/caikit/runtime/grpc_server.py
+++ b/caikit/runtime/grpc_server.py
@@ -196,11 +196,20 @@ class RuntimeGRPCServer(RuntimeServerBase):
             )
             if self.tls_config.client.cert:
                 log.info("<RUN10001806I>", "Running with mutual TLS")
+                # Combine the client cert with the server's own cert so that
+                # health probes can use the server's key/cert instead of needing
+                # one signed by a potentially-external CA.
+                root_certificates = b"\n".join(
+                    [
+                        bytes(self._load_secret(self.tls_config.client.cert), "utf-8"),
+                        tls_server_pair[1],
+                    ]
+                )
                 # Client will verify the server using server cert and the server
                 # will verify the client using client cert.
                 server_credentials = grpc.ssl_server_credentials(
                     [tls_server_pair],
-                    root_certificates=self._load_secret(self.tls_config.client.cert),
+                    root_certificates=root_certificates,
                     require_client_auth=True,
                 )
             else:
--- a/caikit/runtime/health_probe.py
+++ b/caikit/runtime/health_probe.py
@@ -32,6 +32,7 @@ from ..core.toolkit import logging
 log = alog.use_channel("PROBE")
 
 
+@alog.timed_function(log.debug)
 def health_probe() -> bool:
     """Run a health probe against all running runtime servers.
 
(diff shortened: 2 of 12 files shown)


caikit/runtime/health_probe.py
```diff
@@ -0,0 +1,267 @@
...
+        # Third Party
+        import requests  # pylint: disable=import-outside-toplevel
+
+        # Local
+        from .http_server import (  # pylint: disable=import-outside-toplevel
+            HEALTH_ENDPOINT,
+        )
+
+    # Requests requires that the TLS information be in files
+    with _tls_files(tls_key, tls_cert) as tls_files:
```
gkumbhat, line 123: is this leading to writing files with every call making it slow and should we cache the files?

caikit/runtime/health_probe.py
```diff
@@ -0,0 +1,267 @@
...
+            )
+        else:
+            log.debug("Probing TLS gRPC server")
+            credentials = grpc.ssl_channel_credentials(
+                root_certificates=tls_server_cert,
+            )
+        channel = grpc.secure_channel(hostname, credentials=credentials)
+    else:
+        log.debug("Probing INSECURE gRPC server")
+        channel = grpc.insecure_channel(hostname)
```
gkumbhat, line 223: would creating a new channel with every call be slow?

caikit/runtime/grpc_server.py
```diff
@@ -195,11 +195,20 @@ def __init__(
             )
             if self.tls_config.client.cert:
                 log.info("<RUN10001806I>", "Running with mutual TLS")
+                # Combine the client cert with the server's own cert so that
+                # health probes can use the server's key/cert instead of needing
+                # one signed by a potentially-external CA.
+                root_certificates = b"\n".join(
```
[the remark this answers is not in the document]
gabe-l-hart: Yep, just `cat` them together is how you merge PEM files

caikit/runtime/health_probe.py
gabe-l-hart: Oooooh, very good point

caikit/runtime/health_probe.py
gabe-l-hart: Yes-ish. This will be run as a standalone process, so we can't hold anything in memory between calls. I do think the tls files should be cached though

joerunde: it sure would!

But like the files bit, this is a binary meant to be run once on each check. Dunno how you cache a `channel` across multiple processes like that :(

joerunde:
I'd ask if we could vendor in the stock grpc health probe binary for non-mtls cases but that would of course lower our ossf scorecard

caikit/runtime/health_probe.py
gabe-l-hart: As I think about this more, I don't think this is worth doing as it would require having non-temp write access and a configured location where the "cached" files get put. If this is too slow, users can simply use mounted files to begin with and not use inline TLS values.

joerunde (approved):
nice, love the new entrypoints!

PR #553, opened 2023-10-30
:zap::bug: Add proper caching of streams in sources

joerunde:

We came across two issues while testing custom plugins for DataStreamSources:
1. The `.to_data_stream` method is called on each read through the stream, which is really slow when your plugin implementation is reading from a remote resource, and unnecessary when your plugin already cached the remote data locally for stream re-entrancy.
2.  Calling `len` on unpickled DataStreamSources fails, because the base `DataStream` class sets a `_length` attribute that is not re-created on the DataStreamSourceBase's `__set_state__`

This PR caches both the `_length` property and the result of calling `to_data_stream` to fix both

**If applicable**:

### DIFF
--- a/caikit/runtime/service_generation/data_stream_source.py
+++ b/caikit/runtime/service_generation/data_stream_source.py
@@ -422,8 +422,9 @@ class DataStreamSourceBase(DataStream):
         super().__init__(self._generator)
 
     def _generator(self):
-        stream = self.to_data_stream()
-        return stream.generator_func(*stream.generator_args, **stream.generator_kwargs)
+        return self._stream.generator_func(
+            *self._stream.generator_args, **self._stream.generator_kwargs
+        )
 
     def __getstate__(self) -> bytes:
         """A DataStreamSource is pickled by serializing its source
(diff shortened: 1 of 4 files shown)


tests/runtime/service_generation/test_data_stream_source.py
```diff
@@ -104,6 +104,8 @@ def test_pickle_round_trip_primitive():
     round_trip = pickle.loads(pickle.dumps(inst))
     assert round_trip.to_dict() == inst.to_dict()
 
+    validate_data_stream(round_trip, 2, float)
```
joerunde, line 107: These tests were hiding failures, we never actually _read_ the unpickled stream

caikit/runtime/service_generation/data_stream_source.py
```diff
@@ -450,6 +451,13 @@ def name_to_plugin_map(self):
             plugin.get_field_name(self.ELEMENT_TYPE): plugin for plugin in self.PLUGINS
         }
 
+    @cached_property
+    def _stream(self):
```
gabe-l-hart, line 455: Do we need to worry about OOMs here for large data sets?

gabe-l-hart (commented):
One question about OOM concerns, but otherwise looks good

caikit/runtime/service_generation/data_stream_source.py
joerunde: Well, if `_to_data_stream` is loading a bunch of data into memory then it'd still be better to only do it once, right? 

I think it'd be up to the plugin authors to make sure that large datasets are properly handled. The WIP plugin we have now is loading datasets into memory after downloading, but we would probably write into a tempfile instead if we ever configured it to allow large downloads

gabe-l-hart: Make sense!

gabe-l-hart (approved):

PR #554, opened 2023-10-31
🐛 return 404 instead of 500 for incorrect model id with multi model finder

prashantgupta24:

In case of `local_model_finder`, if we send in a non-existent model, it throws a `FileNotFoundError` if the model_path is not found - https://github.com/caikit/caikit/blob/main/caikit/core/model_management/local_model_finder.py#L71

`multi_model_finder` on the other hand, returns a `None` if the model is not found - 
https://github.com/caikit/caikit/blob/main/caikit/core/model_management/multi_model_finder.py#L137

This was causing a scenario to occur in which `multi_model_finder` was throwing a 500 if we sent it a non-existent model:

> {"channel": "MODEL-LOADER", "exception": "Traceback (most recent call last):\n  File \"/home/runner/work/caikit/caikit/caikit/runtime/model_management/model_loader.py\", line 118, in _load_module\n    model = MODEL_MANAGER.load(model_path)\n  File \"/home/runner/work/caikit/caikit/caikit/core/model_manager.py\", line 273, in load\n    return self._do_load(\n  File \"/home/runner/work/caikit/caikit/caikit/core/model_manager.py\", line 459, in _do_load\n    error.value_check(\n  File \"/home/runner/work/caikit/caikit/caikit/core/exceptions/error_handler.py\", line 384, in value_check\n    self(\n  File \"/home/runner/work/caikit/caikit/caikit/core/exceptions/error_handler.py\", line 130, in log_raise\n    raise exception\nValueError: value check failed: Unable to find a ModuleConfig for not_an_id", "level": "error", "log_code": "<RUN62912924E>", "message": "load failed when processing path: not_an_id with error: ValueError('value check failed: Unable to find a ModuleConfig for not_an_id')", "model_id": "not_an_id", "num_indent": 0, "thread_id": 140647025604160, "timestamp": "2023-11-01T19:17:57.688955"}

This PR adds a `ValueError` exception block while loading models, that can catch the `ValueError` thrown by `multi_model_finder` in case it doesn't find a model.

**If applicable**:

### DIFF
--- a/tests/runtime/test_grpc_server.py
+++ b/tests/runtime/test_grpc_server.py
@@ -835,7 +835,7 @@ def test_load_model_badmodel_error_response(runtime_grpc_server):
             modelKey="baz",
         )
         stub.loadModel(load_model_request)
-    assert context.value.code() == grpc.StatusCode.INTERNAL
+    assert context.value.code() == grpc.StatusCode.NOT_FOUND
 
 
 def test_unload_model_ok_response(sample_task_model_id, runtime_grpc_server):
--- a/caikit/runtime/model_management/model_loader.py
+++ b/caikit/runtime/model_management/model_loader.py
@@ -135,6 +135,18 @@ class ModelLoader:
                 StatusCode.NOT_FOUND,
                 f"Model {model_id} not found. Nested error: {fnfe}",
             ) from fnfe
+        except ValueError as ve:
+            log_dict = {
+                "log_code": "<RUN38617724E>",
+                "message": "load failed to find model: %s with error: %s"
+                % (model_path, repr(ve)),
+                "model_id": model_id,
+            }
+            log.error(log_dict)
+            raise CaikitRuntimeException(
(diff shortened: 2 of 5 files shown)


tests/runtime/test_grpc_server.py
```diff
@@ -835,7 +835,7 @@ def test_load_model_badmodel_error_response(runtime_grpc_server):
             modelKey="baz",
         )
         stub.loadModel(load_model_request)
-    assert context.value.code() == grpc.StatusCode.INTERNAL
+    assert context.value.code() == grpc.StatusCode.NOT_FOUND
```
prashantgupta24, line 838: Shouldn't these have been `NOT_FOUND` from the beginning? Maybe I'm missing some context

gabe-l-hart: Nope, you're probably right!

gabe-l-hart (approved):
Ship it! Nice fix Prashant

PR #556, opened 2023-11-01
🎨 fix request validation response

prashantgupta24:

**If applicable**:


caikit/runtime/http_server/http_server.py
```diff
@@ -141,6 +141,21 @@ def __init__(self, tls_config_override: Optional[aconfig.Config] = None):
...
+        @self.app.exception_handler(RequestValidationError)
+        async def request_validation_exception_handler(_, exc: RequestValidationError):
+            err_code = status.HTTP_422_UNPROCESSABLE_ENTITY
+            error_content = {
+                "details": exc.errors()[0]["msg"]
+                if len(exc.errors()) > 0 and "msg" in exc.errors()[0]
+                else exc.errors(),
+                "additional_info": exc.errors(),
+                "code": err_code,
+                "id": None,
```
prashantgupta24, line 154: Not sure if we want to include an always null ID though? 🤔

joerunde: What ids are normally set here by other services that follow this convention? Are they internal ids similar to log codes?

prashantgupta24: Raising a `CaikitRuntimeException` adds a `uuid.uuid4().hex` to the err.id, which we then use. Would it make sense to add it here too?

```
            except CaikitRuntimeException as err:
                error_code = GRPC_CODE_TO_HTTP.get(err.status_code, 500)
                error_content = {
                    "details": err.message,
                    "code": error_code,
                    "id": err.id,
                }
```

joerunde: Right, I have no context on if those IDs are supposed to be
1. unique per instance of an error, in order to help you trace a single failure through a system or
2. unique per error case, so that you could identify the root cause in the code, like we use log codes for

I was wondering what other services that follow the `id / details / code` do. I'd be fine going with case (1) for now though since I dunno how we'd handle (2) effectively across all the places our code is scattered.

> Would it make sense to add it here too?

TL;DR: yeah I think so. Let's make sure that ID is logged by our server before returning the response so that somebody who is digging through logs can find that log statement, and use it to grab a thread id or something that they can use to find more details down the stack

prashantgupta24: Makes sense!

caikit/runtime/http_server/http_server.py
```diff
@@ -393,9 +409,9 @@ async def _handler(context: Request) -> Response:
                 error_content = {
                     "details": f"Unhandled exception: {str(err)}",
                     "code": error_code,
-                    "id": None,
+                    "id": uuid.uuid4().hex,
                 }
-                log.error("<RUN51881106E>", err, exc_info=True)
+                log.error("<RUN51231106E>", error_content, exc_info=True)
```
joerunde, line 414: @prashantgupta24 did we want a `log.error` in the `except CaikitRuntimeException` case as well?

joerunde (approved):
lpgtm! Just wondering about whether we need to make sure to re-log the `CaikitRunitmeExceptions` to get that error.id into the logs

(and lint failures)

caikit/runtime/http_server/http_server.py
prashantgupta24: done

joerunde: hmm, duplicate logging can for sure be annoying...

if a quick test shows that the CREs are already logged with the id in a case or two, then it's probably fine to omit? We can always add it here later if we find the traceability hole

prashantgupta24: Hmm, we log it most of the times, but not all 😞 I'm inclining towards having duplicate logs vs not having logs at all in some cases

joerunde: sounds good to me too!

PR #562, opened 2023-11-03
proposal: replace pylint with ruff

dtrifiro:
[`ruff`](https://github.com/astral-sh/ruff) is a linter written in rust. It is much faster than mypy and supports a wide range of rules. It has also recently gotten formatter functionality (can replace `black` with no changes)

Just as a comparison, on `main`:
```console
$ time tox -e lint
...
  lint: OK (27.41=setup[14.34]+cmd[13.07] seconds)
  congratulations :) (27.46 seconds)
tox -e lint  22.19s user 2.24s system 88% cpu 27.597 total
```
on this branch:
```console
$ time tox -e lint
...
  lint: OK (12.60=setup[12.58]+cmd[0.02] seconds)
  congratulations :) (12.66 seconds)
tox -e lint  8.76s user 0.78s system 74% cpu 12.791 total
```

These times include env setup times. If we only run pylint/ruff on an existing virtualenv:

```console
$ time pylint caikit examples
 pylint caikit examples  20.62s user 1.57s system 107% cpu 20.615 total
```

```console
$ time ruff check examples
$ ruff check caikit examples  0.01s user 0.00s system 130% cpu 0.009 total
```

Other good things about ruff:
- includes a built in fixer, which can fix basic issues by calling `ruff check --fix`, 
- `pre-commit` hooks already exist and can be included in `.pre-commitc-config`
-


dtrifiro:
Please note that I did change more than a few lines of code to fix some ruff warnings.  All of these are in the second commit of this PR and can be validated by running `ruff` on main cherry-picking the first commit of this branch.

pyproject.toml
```diff
@@ -117,3 +117,64 @@ Source = "https://github.com/caikit/caikit"
...
+    # "R0903", # too-few-public-methods
+    # "W0212", # protected-access
+    # "W0511", # fixme
+    # "W1202", # logging-format-interpolation
+    # "E1205", # logging-too-many-args
+    # "W0201", # attribute-defined-outside-init
+    # "W0223", # abstract-method
+    # "W0104", # pointless-statement
+    # "C0411", # wrong-import-order
+]
```
dtrifiro, line 171: These are the original disabled messages in `.pylintrc`

pyproject.toml
```diff
@@ -117,3 +117,64 @@ Source = "https://github.com/caikit/caikit"
...
+    "I001", # import block unsorted/unformatted
+    "E402", # module level import not at top of file
+    # "B028", # warnings: no explicit stacklevel keyword argument found
+    # "I0001", # raw-checker-failed
+    # "I0010", # bad-inline-option
+    # "I0011", # locally-disabled
+    # "I0013", # file-ignored
+    # "I0020", # suppressed-message
+    # "I0021", # useless-suppression
+    # "I0022", # deprecated-pragma
```
dtrifiro, line 146: These are some of the messages that were enabled by default in pylint

gabe-l-hart:
Thank you for tackling this @dtrifiro! I [started looking at exactly this transition](https://github.com/gabe-l-hart/caikit/tree/UseRuff) quite a while ago, but never got back to it. This would be a huge improvement to be able to do inline linting as a commit hook

pyproject.toml
```diff
@@ -117,3 +117,64 @@ Source = "https://github.com/caikit/caikit"
...
+
+
+[tool.ruff]
+line-length = 100
+target-version = "py38"
+exclude = ["caikit/runtime/protobufs/*.py"]
+
+
+[tool.ruff.lint]
+select = [ "E", "F", "UP", "B", "SIM", "I"]
```
dtrifiro, line 129: `select` marks the rules that we do want to check the code for. Here's an extract from the ruff docs

```toml
[tool.ruff.lint]
select = [
  # pycodestyle
  "E",
  # Pyflakes
  "F",
  # pyupgrade
  "UP",
  # flake8-bugbear
  "B",
  # flake8-simplify
  "SIM",
  # isort
  "I",
]

[List of available rules](https://docs.astral.sh/ruff/rules/)

File: caikit/core/data_model/base.py [hunk 4]
```diff
@@ -37,6 +37,10 @@
 from ..exceptions import error_handler
 from . import enums, json_dict, timestamp
 
+if TYPE_CHECKING:
+    # Local
+    from caikit.core.data_model.data_backends import DataModelBackendBase
+    from caikit.interfaces.common.data_model.file import File
```
dtrifiro, line 43: This is what is currently causing CI to fail `tox -e imports`/`./scripts/check_deps.sh`.

I'm assuming `pydeps` has no concept of `TYPE_CHECKING` resulting in this test failing even though this is only being imported when running in type checking mode. I can comment this out, but it would be interesting to have working type checking

caikit/runtime/http_server/http_server.py
```diff
@@ -112,6 +112,7 @@
 # Endpoint to use for server info
 RUNTIME_INFO_ENDPOINT = "/info/version"
 
+
```
gabe-l-hart, line 115: I've seen a number of these additional newlines. I'm surprised our previous `black` formatter didn't catch these OR that it's not complaining about them now. I care zero about how it looks as long as `black` is happy though

caikit/runtime/servicers/global_predict_servicer.py
```diff
@@ -158,61 +158,58 @@ def Predict(
...
 
-        with self._handle_predict_exceptions(model_id, request_name):
-            with alog.ContextLog(
-                log.debug, "GlobalPredictServicer.Predict:%s", request_name
-            ):
-                # Retrieve the model from the model manager
-                log.debug("<RUN52259029D>", "Retrieving model '%s'", model_id)
-                model = self._model_manager.retrieve_model(model_id)
-                model_class = type(model)
+        with self._handle_predict_exceptions(model_id, request_name), alog.ContextLog(
```
gabe-l-hart, line 161: Magic! I did not realize you could chain `with` statements! This changes everything about gorpy extra indentation

pyproject.toml
dtrifiro: We could also leave those in `pyproject.toml`:
```toml
[tool.pylint.message_control]
disable = [
    "use-symbolic-message-instead",
    ...
]
```

pyproject.toml
```diff
@@ -81,8 +81,8 @@ dev-docs = [
 ]
 
 dev-fmt = [
+    "ruff==0.1.3",
```
gabe-l-hart, line 84: Is strict pinning the right way to go here? We've been trying to move away from that towards semver ranges, but for dev tools it's _probably_ fine to pin. Since `ruff` is in `0.`, I'd assume the range here would be `>=0.1.3,<0.2`.

pyproject.toml
```diff
@@ -117,3 +117,64 @@ Source = "https://github.com/caikit/caikit"
 markers = [
     "examples: marks tests as e2e examples (deselect with '-m \"not examples\"')",
 ]
+
+
+[tool.ruff]
+line-length = 100
+target-version = "py38"
```
gabe-l-hart, line 124: Interesting. We aim to support 3.8+, but I'm guessing this is the right call so we don't get flagged for a bunch of syntactic improvements in later versions that we can't use, right?

caikit/core/data_model/base.py
gabe-l-hart: Yeah, that makes sense. I think this may have slipped through my review, but looking more closely, since `File` is _itself_ a `DataBase` derived class, I think we need to rethink this a bit. It's used as a forward declaration here for the `to_file` method, but I think we may want to change that method to instead return something more python-native.

Thoughts @joerunde?

dtrifiro: I ran a few tests with `mypy` and right now I'm getting `Found 761 errors in 84 files (checked 156 source files)`. So we can leave this commented out and get back to it when/if we decide to tackle typing.

caikit/core/data_model/base.py
```diff
@@ -37,6 +37,10 @@
 from ..exceptions import error_handler
 from . import enums, json_dict, timestamp
 
+if TYPE_CHECKING:
```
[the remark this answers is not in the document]
dtrifiro: From the docs:


`typing.TYPE_CHECKING special constant that is assumed to be True by 3rd party static type checkers. It is False at runtime.`


Reference: https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING

caikit/core/data_model/streams/data_stream.py
```diff
@@ -314,15 +315,14 @@ def from_csv(cls, filename: str, *args, skip=0, **kwargs) -> "DataStream[List]":
     @classmethod
     def _from_csv_generator(cls, filename, skip, *csv_args, **csv_kwargs):
         # open the csv file (closure around `filename`)
-        with open(filename, mode="r", encoding="utf8") as fh:
+        with open(filename, encoding="utf8") as fh:
```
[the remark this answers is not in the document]
dtrifiro: Yeah, `ruff` does flag this and I tend to agree, I'm not sure we gain anything by explicitly defining a default for a builtin like `open`

pyproject.toml
dtrifiro: We should be getting PRs from @dependabot for every ruff update. That will also let us know when a `ruff` upgrade introduces something that breaks our linting and let us take action.

pyproject.toml
dtrifiro: Indeed. We previously were targeting 3.9, even if we state that the minimum supported version is 3.8. 

When we drop 3.8, we can bump this up and fix all of the deprecated imports and more, such as 

`from typing import Dict` (we can use `dict` as a type int since 3.9)

caikit/core/data_model/base.py
dtrifiro: I ended up commenting this out and adding pragmas to ignore warnings for now.

caikit/core/data_model/streams/csv_column_formatter.py
```diff
@@ -104,14 +104,14 @@ def _convert(data_item):
...
-                if len(data_item) > len(self._expected_columns):
-                    # More data in the data item left...
-                    if last_type == list:
-                        # Last element was a list, so slurp the rest of the row in
-                        length = len(self._expected_columns)
-
-                        data_item_copy[length - 1].extend(data_item[length:])
-                        data_item_copy = data_item_copy[0:length]
+                if (len(data_item) > len(self._expected_columns)) and (
+                    last_type == list  # More data in the data item left...
```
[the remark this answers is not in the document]
dtrifiro: I changed this a little bit in my last push to clean it up while I'm at it

gabe-l-hart (approved):
Looks great. I see that DCO needs to be rebased and for some reason the CI didn't run on the last push, but I'm happy with where the code is at this point.

.pre-commit-config.yaml
```diff
@@ -13,3 +13,10 @@ repos:
       hooks:
           - id: isort
             exclude: imports
+    - repo: https://github.com/astral-sh/ruff-pre-commit
+      rev: v0.1.3
+      hooks:
+          - id: ruff
+            args:
+                [--fix, --exit-non-zero-on-fix]
+                # - id: ruff-format #TODO: this could replace black
```
dtrifiro, line 22: @gabe-l-hart squeezed this in at the last minute

PR #565, opened 2023-11-03
Health probes no localhost

gabe-l-hart:

This PR fixes the `caikit-health-probe` to support probing a (m)TLS server over `gRPC` when the server's cert does not contain `localhost` as a Subject Alternate Name.

This PR does introduce a dependency on `cryptography` which is a widely used third party library in the python ecosystem. The need for `cryptography` stems from the fact that `grpc` does not allow TLS hostname verification to be fully disabled on the client side. As such, we need to parse the correct `hostname` override from the server's cert in order to set the value correctly in the client. There are numerous libraries to do this, and [it is technically possible using the builtin `ssl`](https://stackoverflow.com/a/50072461), but that requires accessing implementation details that are not guaranteed to be API stable.

### DIFF
--- a/tests/runtime/conftest.py
+++ b/tests/runtime/conftest.py
@@ -438,12 +438,10 @@ def _check_server_readiness(server):
 
 def _check_http_server_readiness(server, config_overrides: Dict[str, Dict]):
     mode = "http"
-    verify = None
     cert = None
     # tls
     if config_overrides:
         mode = "https"
-        verify = config_overrides["use_in_test"]["ca_cert"]
         # mtls
         if "client_cert" and "client_key" in config_overrides["use_in_test"]:
             cert = (
@@ -455,7 +453,7 @@ def _check_http_server_readiness(server, config_overrides: Dict[str, Dict]):
         try:
             response = requests.get(
                 f"{mode}://localhost:{server.port}{http_server.HEALTH_ENDPOINT}",
-                verify=verify,
+                verify=False,
                 cert=cert,
             )
             assert response.status_code == 200
--- a/tests/runtime/test_caikit_health_probe.py
+++ b/tests/runtime/test_caikit_health_probe.py
@@ -31,6 +31,7 @@ import tls_test_tools
 
 # First Party
 from caikit_health_probe import __main__ as caikit_health_probe
+import alog
 
 # Local
 from caikit import get_config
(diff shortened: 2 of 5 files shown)


tests/runtime/conftest.py
```diff
@@ -455,7 +453,7 @@ def _check_http_server_readiness(server, config_overrides: Dict[str, Dict]):
         try:
             response = requests.get(
                 f"{mode}://localhost:{server.port}{http_server.HEALTH_ENDPOINT}",
-                verify=verify,
+                verify=False,
```
joerunde, line 456: #secure!

tests/runtime/test_caikit_health_probe.py
```diff
@@ -93,52 +98,71 @@ def temp_probe_config(*args, **kwargs):
...
+        grpc_port = tls_test_tools.open_port()
+
+        # Set up SAN lists if not putting "localhost" in
+        server_sans, client_sans = None, None
+        if "NO_LOCALHOST" in tls_mode.name:
+            server_sans = ["foo.bar"]
+            client_sans = ["baz.bat"]
+
+        # Set up tls values if needed
+        tls = tls_mode == TlsMode.TLS
```
joerunde, line 128: Since there's now a `TlsMode.TLS_NO_LOCALHOST` does this need to be
```
tls = tls_mode.name.startswith("TLS")
```

tests/runtime/test_caikit_health_probe.py
```diff
@@ -93,52 +98,71 @@ def temp_probe_config(*args, **kwargs):
...
+                with maybe_runtime_grpc_test_server(grpc_port):
+                    # If only running gRPC, health probe should pass
+                    assert caikit_health_probe.health_probe() == (
+                        server_mode == ServerMode.GRPC
+                    )
+                    # If booting the HTTP server, do so
+                    with maybe_runtime_http_test_server(
+                        http_port, tls_config_override=config_overrides
+                    ):
+                        # Probe should always pass with both possible servers up
```
joerunde, line 167: Is there no test for what happens when there's no server up?

joerunde (approved):
Thanks for digging through this!

lgtm as long as the test is doing the tls-no-localhost case correctly

tests/runtime/conftest.py
gabe-l-hart: This is setting up the tests, not the body of the test, so I felt it was appropriate to have this not verify and delegate that responsibility to a test that explicitly verifies.

tests/runtime/test_caikit_health_probe.py
gabe-l-hart: Good find!

tests/runtime/test_caikit_health_probe.py
gabe-l-hart: https://github.com/caikit/caikit/pull/565/files#diff-7b05366351799f2ca03e0507a6bf16c4b2f9b1e47c3535f45cdcd5bffe0c7500R155

PR #567, opened 2023-11-06
No crypto health probes

gabe-l-hart:

This PR is one more attempt to build a robust `caikit-health-probe`, but this time without relying on the `cryptography` package. The decision to not add `cryptography` as a dependency is based on the significant security surface opened up by having this tool in the dependency chain.

With this PR, the health probe can fail when `localhost` is not in the CN/SANs for he server's cert AND the server is not running with the insecure unix socket open. This is an acceptable limitation to trade for not adding a crypto-sensitive library in the dependency list.

### DIFF
--- a/caikit_health_probe/__main__.py
+++ b/caikit_health_probe/__main__.py
@@ -23,10 +23,6 @@ import sys
 import tempfile
 import warnings
 
-# Third Party
-from cryptography import x509
-from cryptography.hazmat.backends import default_backend
-
 # First Party
 import alog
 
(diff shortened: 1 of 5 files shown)


tests/runtime/test_caikit_health_probe.py
```diff
@@ -85,54 +71,101 @@ def temp_probe_config(*args, **kwargs):
...
-        tls = tls_mode.name.startswith("TLS")
-        mtls = tls_mode.name.startswith("MTLS")
         with generate_tls_configs(
             port=http_port,
-            tls=tls,
-            mtls=mtls,
-            inline=inline,
-            separate_client_ca=tls_mode == TlsMode.MTLS_SEPARATE_CLIENT_CA,
+            tls=test_config.tls_mode == TlsMode.TLS,
+            mtls=test_config.tls_mode == TlsMode.MTLS,
```
joerunde, line 166: nice, this is way better than name parsing

gabe-l-hart: Yeah, that was getting out of hand

caikit_health_probe/__main__.py
```diff
@@ -205,8 +201,17 @@ def _grpc_health_probe(
         )
         import grpc  # pylint: disable=import-outside-toplevel
 
+    # Server hostname to use unless using socket mode
     hostname = f"localhost:{port}"
-    if tls_key and tls_cert:
+    socket_file = get_config().runtime.grpc.unix_socket_path
+
+    # If available, use a unix socket
+    if os.path.exists(os.path.dirname(socket_file)):
```
gabe-l-hart, line 209: Oops, this is not safe

```suggestion
    if socket_file and os.path.exists(os.path.dirname(socket_file)):
```

joerunde (approved):
lpgtm?
I'm really surprised that you can specify an insecure unix socket + secure port, but not an insecure port + secure port. Fun quirks and features of grpc, I guess 🤷

caikit_health_probe/__main__.py
joerunde: oh good catch

PR #568, opened 2023-11-06
added new dataobject for user warnings

swith004:
This PR adds a warnings field and associated reason dataobject in order to have the ability to issue a warning when returning classification results.

### DIFF
--- a/caikit/interfaces/nlp/data_model/classification.py
+++ b/caikit/interfaces/nlp/data_model/classification.py
@@ -32,7 +32,7 @@ log = alog.use_channel("DATAM")
 
 @dataobject(package=NLP_PACKAGE)
 class InputWarnings(DataObjectBase):
-    """Input Warning data objects, which returns a reason and message associated with warnings 
+    """Input Warning data object, which returns a reason and message associated with warnings 
     to issue to a user that causes errors (such as failed text generation)
     """
     id: str
--- a/caikit/interfaces/nlp/data_model/classification.py
+++ b/caikit/interfaces/nlp/data_model/classification.py
@@ -14,6 +14,7 @@
 """Data structures for classification representations"""
 
 # Standard
+from enum import Enum
 from typing import List, Optional
 
 # Third Party
@@ -32,13 +33,18 @@ log = alog.use_channel("DATAM")
 
 
 @dataobject(package=NLP_PACKAGE)
-class InputWarnings(DataObjectBase):
+class InputWarningReason(Enum):
+    UNSUITABLE_INPUT = 0
+
+
+@dataobject(package=NLP_PACKAGE)
+class InputWarning(DataObjectBase):
     """Input Warning data object, which returns a reason and message associated with warnings
     to issue to a user that causes errors (such as failed text generation)
     """
 
-    id: str
-    message: str
+    id: Annotated[InputWarningReason, FieldNumber(1)]  # id of input error
+    message: Annotated[str, FieldNumber(2)]  # Error message detailing Warning
 
 
 @dataobject(package=NLP_PACKAGE)
@@ -92,9 +98,6 @@ class TokenClassificationResult(DataObjectBase):
     token_count: Annotated[
         Optional[int], FieldNumber(7)
     ]  # Length of tokens in the text
-    warnings: Annotated[
-        Optional[InputWarnings], FieldNumber(8)
-    ]  # Warning to user in the event of input errors
 
 
 @dataobject(package=NLP_PACKAGE)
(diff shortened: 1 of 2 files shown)


caikit/interfaces/nlp/data_model/classification.py
```diff
@@ -30,6 +30,13 @@
 
 log = alog.use_channel("DATAM")
 
+@dataobject(package=NLP_PACKAGE)
+class InputWarnings(DataObjectBase):
+    """Input Warning data object, which returns a reason and message associated with warnings 
+    to issue to a user that causes errors (such as failed text generation)
+    """
+    id: str
+    message: str
```
joerunde, line 39: These should probably be annotated with field numbers, looks like we're making sure to do that everywhere else

caikit/interfaces/nlp/data_model/classification.py
```diff
@@ -139,6 +152,6 @@ class ClassifiedGeneratedTextStreamResult(ClassifiedGeneratedTextResult):
     """
 
     processed_index: Annotated[
-        Optional[int], FieldNumber(7)
+        Optional[int], FieldNumber(8)
```
joerunde, line 155: I know this probably isn't gonna break anybody if these interfaces are only used in an http server, but `FieldNumber` here maps to a protobuf field number, which cannot be changed without breaking the wire format serialization. See: https://github.com/caikit/caikit/pull/568/checks?check_run_id=18419777460

joerunde: Hah, and this case is called out specifically 😉 

> Common causes of field number reuse:
> - renumbering fields (sometimes done to achieve a more aesthetically pleasing number order for fields). Renumbering effectively deletes and re-adds all the fields involved in the renumbering, resulting in incompatible wire-format changes.
> - deleting a field and not [reserving](https://protobuf.dev/programming-guides/proto3/#fieldreserved) the number to prevent future reuse.

swith004: Oh, I was concerned about renumbering, but wasn't sure it would break anything since it still worked in the guardrails test.. In that case, if I add the warnings field number to be a new field number that doesn't disrupt the order, will that suffice or still cause issues @joerunde ?

caikit/interfaces/nlp/data_model/classification.py
```diff
@@ -82,6 +89,9 @@ class TokenClassificationResult(DataObjectBase):
     token_count: Annotated[
         Optional[int], FieldNumber(7)
     ]  # Length of tokens in the text
+    warnings: Annotated[
+        Optional[InputWarnings], FieldNumber(8)
```
evaline-ju, line 96: Two separate concerns on this
- `warnings` leads me to believe there could be multiple warnings, but the `InputWarnings` data object seems to only allow one message. Should that be singular and a list allowed here?
- Currently none of the tasks expect or return this https://github.com/caikit/caikit/blob/main/caikit/interfaces/nlp/tasks.py, so what is the rationale on putting the warnings on individual `TokenClassificationResult`? i.e. what kind of warnings are we expecting to return on a token level basis rather than an overall results basis?

caikit/interfaces/nlp/data_model/classification.py
joerunde: Yeah you'll need to make sure to not change any existing numbers, and new fields need to use a number that isn't already in use. You would be able to see this fail if you ran a grpc server without the changes and a grpc client with your new changes. Since guardrails is http only, you probably wouldn't even be able to trigger the failure, but we should still make sure we understand how the field numbering works and practice not breaking it.

caikit/interfaces/nlp/data_model/classification.py
swith004: 1. Per SW,  The ideal warning in the event of the guardrails user input detection warning should look something like this: 
`warnings" : [  {
      "id": "generation_prevented",
       "message": "Unsuitable input detected. Please check the detected entities on your input and try again with the unsuitable input removed."  } ]`
      
So the idea could be either a list of InputWarnings if multiple warnings? 

2. You are right - this should only be in ClassifiedGeneratedTextResult (in which case, I don't think I'd have to worry about breaking the order for field numbers

joerunde (approved):
lgtm

prashantgupta24 (changes requested):
looks good, pending unit test

tests/interfaces/nlp/test_classification.py
```diff
@@ -260,3 +270,4 @@ def _validate_classification_generated_text_result(obj):
     assert obj.generated_token_count == 7
     assert obj.finish_reason == 5
     assert obj.seed == 42
+    assert obj.warnings[0].id == InputWarningReason.UNSUITABLE_INPUT.value
```
prashantgupta24, line 273: can you also assert
```
assert obj.warnings[0].message == "unsuitable input detected"
```

swith004: thanks; done!

prashantgupta24 (approved):

PR #574, opened 2023-11-11
Fix test_dict_to_struct_to_dict for proto3 issue

markstur:
Fix test_dict_to_struct_to_dict for proto3 issue.
Adding a dataobject that uses JsonDict made the instanceof tests
fail on proto3 due to the use of the temporary descriptor pool.
This only impacted proto3.

This is a minimal change that I'm expecting to reproduce a fail the proto3 CI tests.
I need to get past that fail for other reasons.  This sample object should help isolate/fix/test.

### DIFF
--- a/tests/fixtures/sample_lib/data_model/sample.py
+++ b/tests/fixtures/sample_lib/data_model/sample.py
@@ -11,6 +11,7 @@ import zipfile
 # Local
 from caikit.core import DataObjectBase, TaskBase, dataobject, task
 from caikit.core.data_model import ProducerId
+from caikit.core.data_model.json_dict import JsonDict
 from caikit.interfaces.common.data_model import File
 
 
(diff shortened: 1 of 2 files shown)


tests/fixtures/sample_lib/data_model/sample.py
```diff
@@ -29,6 +30,18 @@ class SampleListInputType(DataObjectBase):
     inputs: List[SampleInputType]
 
 
+# Test w/ just import and no dataobject
+@dataobject(package="caikit_data_model.sample_lib")
+class JsonDictInputType(DataObjectBase):
```
joerunde, line 35: where / how is this used?

markstur: @joerunde see below.  I'd probably remove it, but I think @prashantgupta24 wanted it kept in.  It does impact the descriptor pool which affects proto3 as described below.

Commented best I could, but "less is more"?  Just asking @prashantgupta24 if I should remove it.  He might want to add.

Issue #570, opened 2023-11-08
Return of 500 on malformed http parameters


evaline-ju:
## Describe the bug

When http parameters are malformed, a `500` `Internal Server Error` is returned rather than an expected `400`.

The validation error can only be seen in the logs

> {"channel": "uvicorn.error", "exception": "Traceback (most recent call last):\n  File \"/opt/miniconda/lib/python3.9/site-packages/caikit/runtime/http_server/pydantic_wrapper.py\", line 200, in pydantic_from_request\n    return pydantic_model.model_validate_json(raw_content)\n  File \"/opt/miniconda/lib/python3.9/site-packages/pydantic/main.py\", line 530, in model_validate_json\n    return cls.__pydantic_validator__.validate_json(json_data, strict=strict, context=context)\npydantic_core._pydantic_core.ValidationError: 1 validation error for caikit.http.TextGenerationTask.TextGenerationTaskRequestHttpRequest\n  Invalid JSON: trailing comma at line 6 column 3 [type=json_invalid, input_value=b'{\\n  \"inputs\": \"what is...el_id\": \"bloom-560m\"\\n}', input_type=bytes]\n    For further information visit https://errors.pydantic.dev/2.4/v/json_invalid\n\nThe above exception was the direct cause of the following exception:\n\nTraceback (most recent call last):\n

## Platform

Please provide details about the environment you are using, including the following:

- Interpreter version:
- Library version: Latest `0.24.1` at time of writing

## Sample Code

This example is with a `caikit-nlp` endpoint but parameter parsing is done through `caikit`

Note trailing comma at `min_new_tokens`
```
curl -k 'POST' \
  'https://localhost:8080/api/v1/task/text-generation' \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -d '{
  "inputs": "string",
  "parameters": {
    "max_new_tokens": 2,
    "min_new_tokens": 0,
  },
  "model_id": "bloom-560m"
}'
```

## Expected behavior

`400`

## Observed behavior

`500`

## Additional context

Would suggest verifying through adding a failing unit test in an http server test first

prashantgupta24:
Hmm, I would have imagined the `RequestValidationError` catching this 😞

evaline-ju:
I looked for that in the logs and it had seemed to be triggered/appear but not sure if there's something else going on 

Sample call
```
curl -k 'POST' \
  'https://localhost:8080/api/v1/task/text-generation' \
[...]

PR #577, opened 2023-11-13
Handle malformed json data for inference in http

waleedqk:
Related to issue: https://github.com/caikit/caikit/issues/570

### DIFF
--- a/caikit/runtime/http_server/http_server.py
+++ b/caikit/runtime/http_server/http_server.py
@@ -148,9 +148,9 @@ class RuntimeHTTPServer(RuntimeServerBase):
         async def request_validation_exception_handler(
             _, exc: RequestValidationError
         ) -> Response:
-            err_code = status.HTTP_422_UNPROCESSABLE_ENTITY 
-            if isinstance(exc.errors()[0]['input'], bytes):
-                exc.errors()[0]['input'] = exc.errors()[0]['input'].decode('utf-8')
+            err_code = status.HTTP_422_UNPROCESSABLE_ENTITY
+            if isinstance(exc.errors()[0]["input"], bytes):
+                exc.errors()[0]["input"] = exc.errors()[0]["input"].decode("utf-8")
             error_content = {
                 "details": exc.errors()[0]["msg"]
                 if len(exc.errors()) > 0 and "msg" in exc.errors()[0]
--- a/caikit/runtime/http_server/http_server.py
+++ b/caikit/runtime/http_server/http_server.py
@@ -149,6 +149,13 @@ class RuntimeHTTPServer(RuntimeServerBase):
             _, exc: RequestValidationError
         ) -> Response:
             err_code = status.HTTP_422_UNPROCESSABLE_ENTITY
+            # try:
+            #     # Attempt to decode the byte string using UTF-8 encoding
+            #     exc.errors()[0]['input'] = exc.errors()[0]['input'].decode('utf-8')
+            # except UnicodeDecodeError as e:
+            #     # Handle the exception if the byte string cannot be decoded using UTF-8
+            #     print(f'Error occurred while converting to UTF-8: {str(e)}')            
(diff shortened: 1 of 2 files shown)


waleedqk:
should a 
```
try:
except UnicodeDecodeError as e:
```
block be added around the decode call?

caikit/runtime/http_server/http_server.py
```diff
@@ -158,7 +158,10 @@ async def request_validation_exception_handler(
                 "id": uuid.uuid4().hex,
             }
             log.error("<RUN59871106E>", error_content, exc_info=True)
-            return Response(content=json.dumps(error_content), status_code=err_code)
+            return Response(
+                content=json.dumps(jsonable_encoder(error_content)),
+                status_code=err_code,
```
prashantgupta24, line 163: We already use `jsonable_encoder` in the response validation! Worth investigating what the difference is between `Response` and `JSONResponse`. We can think about returning the same in both validations

prashantgupta24: Also we should just be able to remove `json.dumps` altogether?
```
(content=jsonable_encoder(error_content), status_code=err_code)
```

tests/runtime/http_server/test_http_server.py
```diff
@@ -1025,3 +1025,14 @@ def test_train_other_task(client, runtime_http_server):
...
+
+def test_inference_malformed_param(client):
+    """Send a malformed data parameter field to the the inference call to induce the correct HTTP error"""
+
+    response = client.post(
+        f"/api/v1/task/sample",
+        data='{"bad_input": 100,}',  # send intentionally bad json
+        headers={"Content-Type": "application/json"},
+    )
+    assert response.status_code == 422
```
prashantgupta24, line 1038: let's assert everything else that comes along with response as well

tests/runtime/http_server/test_http_server.py
```diff
@@ -1025,3 +1025,14 @@ def test_train_other_task(client, runtime_http_server):
     json_response = json.loads(response.content.decode(response.default_encoding))
     assert response.status_code == 200, json_response
     assert json_response["farewell"] == "goodbye: world 64 times"
+
+
+def test_inference_malformed_param(client):
+    """Send a malformed data parameter field to the the inference call to induce the correct HTTP error"""
```
prashantgupta24, line 1031: typo - multiple `the`

caikit/runtime/http_server/http_server.py
waleedqk: Using `jsonable_encoder` returns the dict, that no longer has the byte object, but still the bad json
so using the `json.dumps` converts it to `str` so it can be passed without error

waleedqk: The `JSONResponse` does not get called when I put a breakpoint in the `validation_exception_handler` method. Looking into the call stack to figure out how / when that is called?

prashantgupta24: Yeah it's hard to trigger the response validation since I don't think we have a test for that. Better to read up on the implementation differences between the 2.

prashantgupta24 (changes requested):
Looking good! Couple of nits:
1. Let's change the PR title
2. Standardize on `JSONResponse` to be returned

tests/runtime/http_server/test_http_server.py
```diff
@@ -627,6 +627,22 @@ def test_inference_streaming_sample_module_actual_server_throws(
         assert stream_responses[1].get("code") == 400
 
 
+def test_inference_malformed_param(client):
+    """Send a malformed data parameter field to the inference call to induce the correct HTTP error"""
+
+    response = client.post(
+        f"/api/v1/task/sample",
```
prashantgupta24, line 634: no need for the `f` here anymore

prashantgupta24 (approved):
LGTM! Thanks Waleed

gabe-l-hart (approved):
Looks good!

PR #578, opened 2023-11-13
Nested model names

gabe-l-hart:

This PR makes it possible to load nested models inside of `local_models_dir` without them being immediately unloaded by the periodic sync logic.

This _does_ change a log code from `W` to `D` because we're changing the semantics of finding non-model directories in `local_models_dir` from being "bad" to being expected. It does NOT silence `COR80419785E` which is raised [here](https://github.com/caikit/caikit/blob/main/caikit/core/model_manager.py#L280). From the perspective of `load`, this _is_ an error, but when used by `runtime` it is a valid outcome of attempting to load something that isn't actually a model.

The fix I attempted for this was to [filter `disk_models`](https://github.com/caikit/caikit/blob/main/caikit/runtime/model_management/model_manager.py#L431) to only those containing a `config.yml` file, but that breaks files with `.zip` suffixes. I could certainly adjust the filter to allow `.zip` files, but this gets into reimplementing the logic of the core `load` function and could make for difficult maintenance if more archive formats are added.

**If applicable**:

### DIFF
--- a/caikit/runtime/model_management/model_manager.py
+++ b/caikit/runtime/model_management/model_manager.py
@@ -443,10 +443,8 @@ class ModelManager:  # pylint: disable=too-many-instance-attributes
         unload_models = [
             model_id
             for model_id, loaded_model in self.loaded_models.items()
-            if model_id not in disk_models
-            and loaded_model.path().startswith(
-                self._local_models_dir,
-            )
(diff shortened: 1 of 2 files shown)


joerunde (approved):
🌶️🌶️🌶️

PR #580, opened 2023-11-13
Unload model race

gabe-l-hart:

This PR addresses a race condition that was showing up due to another bug where models in the `local_models_dir` that were nested inside subdirectories were auto-unloaded when the sync cron fired. Because of how often models were being unloaded, it was reasonably common that a model would be unloaded _after_ being added to the global `self.loaded_models` map, but _before_ the `retrieve_model` call finished, causing [the re-lookup](https://github.com/caikit/caikit/blob/main/caikit/runtime/model_management/model_manager.py#L413) to raise a `KeyError`.

The fix in this PR is to not re-index into `self.loaded_models`. This does still expose the possible race condition where a request might appear as though a model were still loaded, even after it were removed from the map, but that would only manifest for a single request and subsequent requests would fail to find the model.

* The bug that revealed this race was fixed in the [previous PR](https://github.com/caikit/caikit/pull/578)
* The change to the return value of `caikit.runtime.model_management.ModelManager.load_model` is NOT considered an API breaking change. The public API of `caikit.runtime` is only the server APIs and the extensible interfaces, so even though it's a public function on the class, it's not API breaking to the overall `caikit.runtime` API.

**If applicable**:

### DIFF
--- a/caikit/runtime/model_management/model_manager.py
+++ b/caikit/runtime/model_management/model_manager.py
@@ -378,8 +378,8 @@ class ModelManager:  # pylint: disable=too-many-instance-attributes
             )
 
         # Now retrieve the model and fall back to lazy loading
-        model_loaded = model_id in self.loaded_models
-        if not model_loaded and self._lazy_load_local_models:
+        loaded_model = self.loaded_models.get(model_id)
+        if not loaded_model and self._lazy_load_local_models:
             local_model_path = os.path.join(self._local_models_dir, model_id)
             log.debug2(
                 "Lazy loading local model %s from %s", model_id, local_model_path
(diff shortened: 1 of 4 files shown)


tests/runtime/model_management/test_model_manager.py
```diff
@@ -525,6 +525,44 @@ def test_nested_local_model_load_unload(good_model_path):
...
+            model_cache_path = os.path.join(cache_dir, model_id)
+            shutil.copytree(good_model_path, model_cache_path)
+
+            # Patch the manager's load_model to immediately unload the model
+            orig_load_model = manager.load_model
+
+            def load_and_unload_model(self, model_id: str, *args, **kwargs):
+                res = orig_load_model(model_id, *args, **kwargs)
+                manager.unload_model(model_id)
+                return res
```
joerunde, line 558: 🤯 💥

caikit/runtime/model_management/model_manager.py
```diff
@@ -410,7 +411,7 @@ def retrieve_model(self, model_id: str) -> ModuleBase:
 
         # NOTE: If the model is partially loaded, this call will wait on the
         #   model future in the LoadedModel
-        return self.loaded_models[model_id].model()
+        return loaded_model.model()
```
joerunde, line 414: 👍  this just looks better too

tests/runtime/model_management/test_model_manager.py
```diff
@@ -525,6 +525,44 @@ def test_nested_local_model_load_unload(good_model_path):
...
+
+            with patch.object(manager.__class__, "load_model", load_and_unload_model):
+
+                # Retrieve the model and make sure there's no error
+                assert manager.retrieve_model(model_id)
```
joerunde, line 563: Should there be any other check for completeness here that the model id isn't in the model manager's map?

joerunde (approved):
looking 😎

Issue #575, opened 2023-11-13
Add support for cancelling HTTP requests


HonakerM:
## Is your feature request related to a problem? Please describe.

Currently only the GRPC runtime server supports aborting work. It would be nice if the HTTP server could do something similar. Without this the runtime server will continue processing requests even if there is no client to receive  

## Describe the solution you'd like

I would like there to be a new `RequestAborter` class which handles aborting HTTP work.

HonakerM:
With the merging of the above Pr this can be closed!

PR #582, opened 2023-11-14
Add HTTP Request Aborter

HonakerM:
This PR supports #575 by adding a new `RequestAborter` class to abort in-progress HTTP requests. This allows the HTTP server to stop work when a client is no longer listening for the response. This is especially important for long running inference requests in-which the client might timeout. 

I put the `RequestAborter` implementation in the `work_management` module but it might make more sense keeping it in `http_server`.

**If applicable**:


caikit/runtime/work_management/request_aborter.py
```diff
@@ -0,0 +1,107 @@
...
+"""
+This module helps us know when a HTTP client call is cancelled, and we need to stop or undo work
+"""
+# Standard
+from typing import Optional
+import asyncio
+import threading
+
+# Third Party
+from fastapi import Request
```
gabe-l-hart, line 24: Yeah, this needs to go in `http_server` since `fastapi` is an optional dependency. We're trying to keep all usage of these optional dependencies as isolated as possible to avoid needing to track how/when they get imported.

caikit/runtime/work_management/request_aborter.py
```diff
@@ -0,0 +1,107 @@
...
+# First Party
+import alog
+
+# Local
+from caikit.runtime.work_management.abortable_action import ActionAborter
+
+log = alog.use_channel("REQUEST-ABORTER")
+
+
+class RequestAborter(ActionAborter):
```
gabe-l-hart, line 35: I think it would be cleaner to call this `HttpRequestAborter` and we might also want to rename `RpcAborter` to `GrpcRequestAborter` or something similar

caikit/runtime/work_management/request_aborter.py
gabe-l-hart: In that same vein, we probably want/need to consolidate the places where `grpc` is used into a sub-module that's quarantined from proactive imports as well

tests/runtime/work_management/test_request_aborter.py
```diff
@@ -0,0 +1,105 @@
...
+    # Wait for uvicorn to start
+    while not server.started:
+        check_server_timeout(start_time, server)
+        time.sleep(0.01)
+
+    # Try the endpoint but timeout after 1 second
+    with pytest.raises(ReadTimeout):
+        requests.get(
+            f"http://localhost:{open_port}/test_abort",
+            timeout=1,
```
gabe-l-hart, line 90: Can we make this timeout tighter?

tests/runtime/work_management/test_request_aborter.py
```diff
@@ -0,0 +1,105 @@
...
+        check_server_timeout(start_time, server)
+        time.sleep(0.01)
+
+    # Assert the request aborter actually aborted
+    assert test_request_aborter.TEST_ABORTER.must_abort()
+    assert abort_event.is_set()
+    assert request_finished.is_set()
+
+    # Clean up the server
+    server.should_exit = True
```
gabe-l-hart, line 104: Let's put all of this in a `try/finally` so it doesn't hang other tests if this one starts failing

gabe-l-hart (changes requested):
A couple of quick suggestions in the test, but otherwise this looks great. Thanks for the contribution Casey!

caikit/runtime/work_management/request_aborter.py
HonakerM: Okay that's what I thought! I'll move it over.

caikit/runtime/work_management/request_aborter.py
HonakerM: Okay! I wasn't sure due to the naming of `RpcAborter` but I definitely like `HttpRequestAborter` more.

caikit/runtime/work_management/request_aborter.py
```diff
@@ -0,0 +1,107 @@
...
+        self.poll_time = poll_time
+
+        # Set initial
+        self.is_terminated = threading.Event()
+        # Add an empty list for event variables that will be notified on termination
+        self.events: list[threading.Event] = []
+
+        # Start request aborter task
+        log.debug("<RUN81824293>", "Watching for request disconnect")
+        self.event_loop.create_task(self.watch_for_disconnect())
```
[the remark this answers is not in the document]
HonakerM: Uhhhhhhh, that's a good question. I'm not the most familiar with asyncio so let me take a closer look. There's probably a better way to implement this than `create_task`

tests/runtime/work_management/test_request_aborter.py
HonakerM: Yes? Just to make sure by tighter you mean lower right? I'm assuming to make the test finish a bit quicker

gabe-l-hart: Yep. @joerunde yells at me every time I increase the overall test duration by `0.01`, so I think he'd have strong words for `1`

caikit/runtime/http_server/request_aborter.py
```diff
@@ -0,0 +1,108 @@
...
+    work to be terminated when a client time's out or stops listening.
+
+    IFF the client request has been terminated, `must_abort` will return True.
+    """
+
+    def __init__(
+        self,
+        context: Request,
+        loop: Optional[asyncio.AbstractEventLoop] = None,
+        poll_time: Optional[float] = 1,
```
gabe-l-hart, line 48: It seems like we might want to make this default a little smaller since many requests will not last much longer than 1s even if not aborted.

gabe-l-hart (changes requested):
One tiny request that we make the default `poll_time` shorter so we catch canceled inference requests quicker and save the work associated. Otherwise, it's good to go!

caikit/runtime/http_server/http_server.py
```diff
@@ -534,12 +551,18 @@ async def _generator() -> pydantic_response:
...
                             **request_params,
                         )
                     ):
                         yield ServerSentEvent(
                             data=result.to_json(), event=StreamEventTypes.MESSAGE.value
                         )
+
+                    # Clean up Aborter to ensure object is deleted
+                    if aborter:
+                        aborter.abort()
```
joerunde, line 564: This smells like it should be a context manager?

```
with HttpRequestAborter(context) as aborter:
    ...
    self.global_predict_servicer.predict_model(
        ...
        aborter=aborter
    )
```
(I'm just guessing there's a case where a non-aborting-related exception could cause this cleanup to be skipped)

If constructing one of these is light, we can probably construct it always since the `AbortableAction` won't be created later if `config.runtime.use_abortable_threads` is false

caikit/runtime/http_server/request_aborter.py
HonakerM: Can do! I'm coming from a land where the average response time can be as high as 75s...

caikit/runtime/http_server/http_server.py
HonakerM: That's a good question. I'm not sure how lightweight it is since it runs an async task in the background.  However, I think this might be a perfect usecase for [nullcontext!](https://docs.python.org/3.7/library/contextlib.html#contextlib.nullcontext) (I've never had the chance to use one for real!). I'll push a commit with what I mean

caikit/runtime/http_server/http_server.py
```diff
@@ -460,18 +461,26 @@ async def _handler(
...
-                    model_id=model_id,
-                    request_name=rpc.request.name,
-                    inference_func_name=model.get_inference_signature(
-                        output_streaming=False, input_streaming=False, task=rpc.task
-                    ).method_name,
-                    **request_params,
+                aborter_context = (
+                    HttpRequestAborter(context)
+                    if get_config().runtime.use_abortable_threads
+                    else nullcontext()
```
gabe-l-hart, line 467: Ah, very nice! I wasn't aware of `nullcontext`

gabe-l-hart (approved):
Looks great, thanks Casey!

joerunde (approved):
🌶️🌶️🌶️

PR #583, opened 2023-11-14
Add exist_ok option to ModuleSaver and only rmtree when exist_ok=False

### LINKED ISSUE #525: Unsafe rmtree in ModuleSaver on any error can cause data loss :bomb:
## Describe the bug

ModuleSaver wants to rmtree to remove any files it created (or might have touched) if any error occurs.  This is assuming that the given model_path was either newly created or no longer usable due to some saver error (harsh but maybe appropriate in specific use cases).

But! model_path may contain important files that should not be deleted.  In the worst case, if the error was caused by an inappropriate model_path (e.g. "" or "/") then the _exit will just wipeout files.  This would not only clobber some model directory, or files created/touched, but potentially anything on your disk.

Even if the error was on a good model_path with existing files, reverting to prior state would be more appropriate than :bomb:

## Sample Code

don't try it

## Expected behavior

shutil.rmtree() is dangerous and should never be used except with tightly controlled parameters.

## Observed behavior

My local caikit-nlp repo was wiped out, when testing without setting a model_path.  I think this is what caused it.  Just a bootstrap/save script that lacked arg checks (which is apparently how folks are supposed to get model config files).

## Additional context

The general behavior of rmtree is the bigger concern (please don't), but the use case here where ModuleSaver seems to be designed with exist_ok=True and yet expects to wipeout that path when errors occur is questionable.  This really needs to be able to revert any changed file on error.

markstur:
The prior implementation was dangerous because it could remove existing files including existing models or any accidental path passed in. This is a bit safer.

Fixes #525

**If applicable**:

### DIFF
--- a/caikit/core/modules/saver.py
+++ b/caikit/core/modules/saver.py
@@ -41,7 +41,8 @@ error = error_handler.get(log)
 
 class ModuleSaver:
     """A module saver that provides common functionality used for saving modules and also a context
-    manager that cleans up gracefully in case an error is encountered during the save process.
+    manager that cleans up in case an error is encountered during the save process for a model_path
+    that did not already exist.
     """
 
     SAVED_KEY_NAME = "saved"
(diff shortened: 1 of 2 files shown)


markstur:
@gkumbhat

joerunde (approved):
LPGTM!

PR #584, opened 2023-11-15
♻️ move stuff into their own packages

prashantgupta24:

It seems like the bulk of work needed for https://github.com/caikit/caikit/issues/585 and the separation of the `http_server` and the `grpc_server` dependency overlap was in `tests` - and in order to maintain consistency, changes had to be done in code as well. So this PR:

in code:
- introduces 2 new packages, `grpc_server` and `http_server` and puts specific modules within those packages.

in tests:
- introduces 2 new conftests within `http_server` and `grpc_server` packages to add specific information.
- refactors the `tests/runtime/conftest.py` to only contain pieces of information that's needed by both. 

**If applicable**:


tests/runtime/grpc/conftest.py
prashantgupta24, line 1: Let me know if the `grpc` directory and moving stuff that requires grpc specific tests within it makes sense

joerunde:
I wonder what the limits on usage of github actions are? Will we find them?

tests/runtime/grpc/conftest.py
joerunde: Yeah I think this makes a lot of sense here. I also think that these two statements are true:
- We like to have the `tests` package structure mirror the `src` package structure
- We think it's valuable to use package boundaries to confine optional dependencies

So it sounds like it's time to do a little package refactoring in caikit.runtime.
`caikit.runtime.http` with the http server and friends, and `caikit.runtime.grpc` with the grpc server and friends?

prashantgupta24: Yep that would make a lot of sense! Does it make sense to do that in this PR?

gabe-l-hart: 100% agree Joe! If we're doing it, seems like this is as good a PR as any to do it in

gabe-l-hart (changes requested):
Thank you for tackling this Prashant! I'm really glad to see that all of the changes are in `tests`. Per the code-structure comment, I think Joe is right that we should use this as an opportunity to shuffle things around and get all of the optional deps to be cleanly isolated.

pyproject.toml
```diff
@@ -40,14 +40,19 @@ caikit-render-interfaces = "caikit.runtime.dump_services:main"
 
 ## Runtime Extra Sets ##
 
+runtime-base = [
```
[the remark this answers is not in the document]
prashantgupta24: Hmm, the only reason why I didn't do that is because `prometheus_client` is a base requirement for the runtime and not the core? Not sure if that makes a difference

caikit/runtime/interceptors/caikit_runtime_server_wrapper.py
prashantgupta24, line 1: I guess technically this can go under `grpc` as well. Thoughts?

pyproject.toml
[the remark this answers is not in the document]
joerunde: 🤞 that we can remove this req with the move to open telemetry... at some point.

But what do we do about requirements that in general are required for metrics and would be required for either the http or grpc server? `runtime-metrics`?

caikit/runtime/__init__.py
```diff
@@ -13,4 +13,8 @@
 # limitations under the License.
 
 # Local
-from .service_factory import get_inference_request, get_train_params, get_train_request
+from .grpc_server.service_factory import (
```
gabe-l-hart, line 16: Hm, I'm still confused here. The `service_factory` stuff seems like it should _not_ be specific to the `grpc_server` since it is also required by the `http_server` in order to wrap the deduced services into HTTP APIs?

prashantgupta24: Yeaaah, that's unfortunately because of the fact that in `test_service_factory`, we use the `runtime_grpc_server` fixture, which can only be available within `grpc_server` package. And since we want to mirror the code to the tests, I'm stuck with this

prashantgupta24:
Hmm, for some reason, a test case keeps failing on 3.11 only on GH. Using 3.11 locally works fine :/

prashantgupta24:
For some inexplicable reason, the changes in this PR make one test in `test_abortable_action.py` fail only with linux’s python 3.11 version! It works if we run our GH actions on `darwin` instead of `linux` 😭

prashantgupta24:
Things that make this PR build successfully as of now:
1. Use `darwin` instead of linux for the GH code runners
2. Not raise errors for `UserWarning`
3. Explicitly ignore `PytestUnraisableExceptionWarning`

gabe-l-hart:
@prashantgupta24 I know this one got pretty out of control with moving things around. Let's discuss how we want to handle this going forward since this would potentially cause some serious rearrangements and breaking changes to semi-public APIs (usage of modules within `caikit.runtime` such as how the guardrails server operates).

gabe-l-hart:
Closing as Stale. @prashantgupta24 we can revisit this later if it bubbles back up

Issue #467, opened 2023-09-13
Full error handling with lazy loading models in runtime


gabe-l-hart:
## Is your feature request related to a problem? Please describe.

As a user of `caikit.runtime`, I need to be able to know what happens when an inference call for a model fails for some reason. In particular, the following scenarios need to be handled:

1. Inference of a model tied to a TGIS backend that is not currently available
    * This would include if that TGIS instance is in the process of booting and not yet Ready OR the instance is simply not available
2. Something unexpected failed in the model load process (random exception thrown, resources missing, etc...)

We may also want to consider the case where a user does not want to wait for cold inference (inference that triggers lazy loading).

PR #588, opened 2023-11-27
Don't lazy load models that are actively being written

joerunde:

This fixes an issue where the lazy load loop can load a large model that is in the progress of being written, where the module itself may gracefully handle partial model artifacts.

This is done by running two back-to-back size checks on the model directory before loading, and only loading if the sizes match. This could have performance implications for models with a very large number of file artifacts. The amount of time that needs to elapse between these two checks to correctly catch models mid-write could vary depending on the specifics about how the bytes are making it to disk.

Possibly related to #467 

This is a very simple fix, it may be _too_ simple

**If applicable**:

### DIFF
--- a/caikit/runtime/model_management/model_manager.py
+++ b/caikit/runtime/model_management/model_manager.py
@@ -508,6 +508,13 @@ class ModelManager:  # pylint: disable=too-many-instance-attributes
                     )
 
     def _model_write_in_progress(self, model_dir: str) -> bool:
+        """Returns true if model_dir is currently being written to. Uses the
+        runtime.lazy_load_write_detection_period_seconds configuration to sleep between
+        consecutive size checks of the directory.
+
+        Always returns false if runtime.lazy_load_write_detection_period_seconds is zero,
+        negative, or None.
+        """
         if (
             self._lazy_load_write_detection_period_seconds is None
             or self._lazy_load_write_detection_period_seconds <= 0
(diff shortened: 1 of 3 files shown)


caikit/runtime/model_management/model_manager.py
```diff
@@ -490,6 +507,35 @@ def _local_models_dir_sync(self, wait: bool = False):
...
+        Always returns false if runtime.lazy_load_write_detection_period_seconds is zero,
+        negative, or None.
+        """
+        if (
+            self._lazy_load_write_detection_period_seconds is None
+            or self._lazy_load_write_detection_period_seconds <= 0
+        ):
+            return False
+
+        # Get the current directory size
```
gabe-l-hart, line 524: How expensive is this? Is there any worry that it would be extremely expensive when "disk" is implemented as a shim to S3?

gabe-l-hart (approved):
This looks great! Very nice test too. The only question I have is around the cost of computing the disk size for mounted filesystems, but I think we would expect that to be intelligently implemented by the volume implementation.

caikit/runtime/model_management/model_manager.py
joerunde: Yeah idk!

I'm making the blind assumption that mounted filesystems like `s3fs` would do something smarter than a full remote read just to stat a file 🤷 

Might research it up a bit later

PR #590, opened 2023-11-29
Add endpoint to get information about loaded models

HonakerM:
This PR allows for end-users to get information about the currently loaded models. This includes things like the model name, path, version, and size. The information is gathered from the `runtime.ModelManager.loaded_models` values.

This endpoint will be useful for the RemoteModel work and for any thin-client to discover models.

**If applicable**:

### DIFF
--- a/caikit/core/modules/base.py
+++ b/caikit/core/modules/base.py
@@ -79,6 +79,17 @@ class ModuleBase(metaclass=_ModuleBaseMeta):
             self._metadata = {}
         return self._metadata
 
+    def get_metadata(self) -> Dict[str, str]:
+        """Helper function to return public metadata about a Module. This is kept
+        separate from the metadata parameter to allow for fields to be excluded. This
+        function also requires a flat metadata structure without nested dictionaries
+
+        Returns:
+            Dict[str, str]: A dictionary of this module's public metadata
+        """
+
+        return {"name": self.MODULE_NAME, "version": self.MODULE_VERSION}
+
     def set_load_backend(self, load_backend):
         """Method used by the model manager to indicate the load backend that
         was used to load this module
--- a/caikit/core/modules/base.py
+++ b/caikit/core/modules/base.py
@@ -80,16 +80,31 @@ class ModuleBase(metaclass=_ModuleBaseMeta):
         return self._metadata
 
     @property
-    def module_metadata(self) -> Dict[str, Any]:
-        """Helper property to return public metadata about a Module. This function
-        is separate from `metdata` as this is specific for the root module. This
-        function also requires a flat metadata structure without nested dictionaries
+    def module_metadata(cls) -> Dict[str, Any]:
+        """Helper property to return metadata about a Module. This function
+        is separate from `metadata` as this is specific for the class module. This
+        function also requires a flat metadata structure without nested dictionaries.
+
+        NOTE: This should be a @classmethod but using @property/@classmethod together has
+        been deprecated
 
         Returns:
-            Dict[str, str]: A dictionary of this module's public metadata
+            Dict[str, str]: A dictionary of this ModuleBases's metadata
         """
 
-        return {"name": self.MODULE_NAME, "version": self.MODULE_VERSION}
+        return {"name": cls.MODULE_NAME, "version": cls.MODULE_VERSION}
+
+    @property
+    def public_model_info(cls) -> Dict[str, Any]:
+        """Helper property to return public metadata about a specific Model. This
+        function is separate from `metdata` as that contains the entire ModelConfig
+        which might not want to be shared/exposed.
+
+        Returns:
+            Dict[str, str]: A dictionary of this models's public metadata
+        """
+
+        return {}
 
     def set_load_backend(self, load_backend):
         """Method used by the model manager to indicate the load backend that
--- a/caikit/core/modules/base.py
+++ b/caikit/core/modules/base.py
@@ -79,9 +79,10 @@ class ModuleBase(metaclass=_ModuleBaseMeta):
             self._metadata = {}
         return self._metadata
 
-    def get_metadata(self) -> Dict[str, str]:
-        """Helper function to return public metadata about a Module. This is kept
-        separate from the metadata parameter to allow for fields to be excluded. This
+    @property
+    def module_metadata(self) -> Dict[str, Any]:
+        """Helper property to return public metadata about a Module. This function
+        is separate from `metdata` as this is specific for the root module. This
         function also requires a flat metadata structure without nested dictionaries
 
         Returns:
--- a/caikit/interfaces/runtime/data_model/info.py
+++ b/caikit/interfaces/runtime/data_model/info.py
@@ -16,7 +16,7 @@ This file contains interfaces to handle information requests
 """
 
 # Standard
-from typing import Dict, Optional
+from typing import Dict, List, Optional
 
 # First Party
 from py_to_proto.dataclass_to_proto import Annotated, FieldNumber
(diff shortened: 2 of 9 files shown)


tests/runtime/http_server/test_http_server.py
```diff
@@ -824,6 +824,27 @@ def test_runtime_info_ok_custom_python_packages(runtime_http_server):
             assert "py_to_proto" not in json_response["python_packages"]
 
 
+def test_models_info_ok(runtime_http_server, sample_task_model_id):
+    """Make sure the runtime info returns version data"""
+    with TestClient(runtime_http_server.app) as client:
```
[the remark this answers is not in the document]
HonakerM: Good call! I just pushed an update

caikit/interfaces/runtime/data_model/info.py
```diff
@@ -39,3 +39,30 @@ class RuntimeInfoRequest(DataObjectBase):
 class RuntimeInfoResponse(DataObjectBase):
     runtime_version: Annotated[Optional[str], FieldNumber(1)]
     python_packages: Annotated[Dict[str, str], FieldNumber(2)]
+
+
+@dataobject(RUNTIME_PACKAGE)
+class ModelInfoRequest(DataObjectBase):
+    """Empty request for runtime server information"""
```
gabe-l-hart, line 46: Does this need to specify a specific `model_id`?

caikit/interfaces/runtime/data_model/info.py
```diff
@@ -39,3 +39,30 @@ class RuntimeInfoRequest(DataObjectBase):
...
+
+    # Model information
+    model_path: Annotated[str, FieldNumber(1)]
+    name: Annotated[str, FieldNumber(2)]
+    size: Annotated[int, FieldNumber(3)]
+
+    # Module Information
+    module_id: Annotated[str, FieldNumber(4)]
+    module_name: Annotated[str, FieldNumber(5)]
+    version: Annotated[str, FieldNumber(6)]
```
gabe-l-hart, line 61: I haven't made it to the service implementation yet, but I'm assuming that it is completely agnostic to the specific model and only reports the metadata known by `caikit.runtime`. I suspect that we'll want to extend this with a freeform field that can call an overridable function on an individual `model` instance and provide model-type-specific information. Here, I would think that would look something like `metadata: Dict[str, str]`

caikit/runtime/servicers/info_servicer.py
```diff
@@ -38,12 +44,65 @@ class InfoServicer:
...
+
+        Args:
+            request: ModelInfoRequest
+            context
+
+        Returns:
+            models_info: ModelInfoResponse
+                DataObject containing the model info
+        """
+        return self.get_models_info_impl().to_proto()
```
gabe-l-hart, line 60: NIT: We've tended to use `_get_models_info` rather than `get_models_info_impl` naming to indicate private impl functions

caikit/runtime/servicers/info_servicer.py
```diff
@@ -38,12 +44,65 @@ class InfoServicer:
...
+            model_info: ModelInfoResponse
+                DataObject with model information
+        """
+        model_manager = ModelManager.get_instance()
+
+        # Get all loaded models
+        response = ModelInfoResponse(models=[])
+        for name, loaded_module in model_manager.loaded_models.items():
+            model_instance = loaded_module.model()
+            response.models.append(
```
gabe-l-hart, line 84: Here's my proposal to make this more flexible going forward (could be in a follow-on PR):

1. Add a `def get_metadata(self) -> Dict[str, str]` method to `ModuleBase`
2. Add a `metadata: Dict[str, str]` field to `ModelInfo` that defaults to `{}`
3. Populate (2) from (1) here if (2) returns anything

This would allow for custom logic in things like remote TGIS implementations of `text_generation`.

gabe-l-hart (changes requested):
I think there are two things we can do to make this immediately more flexible and extensible:

1. Add `model_ids: List[str]` to the request so that users can request a subset of the models
2. Add a `metadata` field to the output that is populated from a new base-class function `get_metadata` on `ModuleBase`

caikit/runtime/servicers/info_servicer.py
HonakerM: Yeah, that's my preferred as well.  I saw that the info servicer was already using `get_runtime_info_impl` so I tried to follow the same convention. I'll update both to use underscores

caikit/interfaces/runtime/data_model/info.py
HonakerM: Good call on the list! Updating the endpoint/tests now

caikit/interfaces/runtime/data_model/info.py
gabe-l-hart: Sounds great. I think those other fields make sense in `metadata`

caikit/runtime/servicers/info_servicer.py
joerunde: @HonakerM @gabe-l-hart the metadata property already exists?
```
    @property
    def metadata(self) -> Dict[str, Any]:
```

HonakerM: @joerunde I saw that but I think having a separate `get_metadata` function would be useful. The `metadata` property returns the full Moduleconfig that was used during loading and I don't think we want to return that (especially if weights/info could be proprietary)

caikit/core/modules/base.py
```diff
@@ -79,6 +79,17 @@ def metadata(self) -> Dict[str, Any]:
             self._metadata = {}
         return self._metadata
 
+    def get_metadata(self) -> Dict[str, str]:
```
gabe-l-hart, line 82: After @joerunde pointed out that we already have `def metadata` above, I think we should figure out how to combine these two. My thought would be that this addition of `"name"` and `"version"` could/should be added by the service on top of the output of the `metadata` property.

caikit/runtime/http_server/http_server.py
```diff
@@ -760,6 +768,33 @@ def _get_response_openapi(
         output = {200: {"content": response_schema}}
         return output
 
+    def _model_info(
+        self, model_ids: Annotated[List[str], Query(default_factory=list)]
+    ) -> Dict[str, Any]:
+        """Create wrapper for get_models_info so model_ids can be marked as a query parameter"""
```
gabe-l-hart, line 774: Does this mean it will _always_ be a query parameter or that it _can_ be in a query parameter?

caikit/runtime/servicers/info_servicer.py
gabe-l-hart: Ok, interesting. I think we should change the name of `get_metadata` to `model_info` then to avoid the naming confusion

gabe-l-hart (changes requested):
I think we need to disambiguate `metadata` and `get_metadata` either by making them the same thing or by changing `get_metadata` to `model_info` or something similar. Otherwise, I think this is ready to go!

caikit/core/modules/base.py
```diff
@@ -79,6 +79,33 @@ def metadata(self) -> Dict[str, Any]:
             self._metadata = {}
         return self._metadata
 
+    @property
```
gabe-l-hart, line 82:
```suggestion
    @classmethod
    @property
```

caikit/runtime/http_server/http_server.py
gabe-l-hart: I'd still like to understand the semantics of this one

gabe-l-hart (changes requested):
One more tiny fix request (missing a `@classproperty`) otherwise I think this is good to go!

caikit/core/modules/base.py
HonakerM: @gabe-l-hart see my comment in the docstring. Combing @property and @classmethod seemed to break the py3.8 tests. Using them together was also [deprecated in 3.11](https://docs.python.org/3.11/library/functions.html#classmethod)

gabe-l-hart: Read better Gabe! I know we have done this in places, so let me see what the right fully compatible way to do it is

caikit/runtime/http_server/http_server.py
HonakerM: @gabe-l-hart sorry I missed this one! It will always be an optional query parameter. From [my understanding](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/GET) it's an anti-pattern to send data in the body of a GET request

gabe-l-hart: Ah, yes, that's definitely true. I'm so used to `gRPC` where everything is a POST!

HonakerM:
@gabe-l-hart Sorry to ping but anything else needed for this PR? I have a PR in the wings that depends on this one

caikit/core/modules/base.py
gabe-l-hart: Ok, I think with this NOTE in the docstring, we're good. Thanks!

gabe-l-hart (approved):

Issue #545, opened 2023-10-26
RUN44773514E obscures true cause of runtime startup failure


ssiegel95:
## Is your feature request related to a problem? Please describe.

Consider the following environment configuration:

```sh
env:
  - name: RUNTIME_LAZY_LOAD_LOCAL_MODELS
     value: "TRUE"
    - name: RUNTIME_TRAINING_SAVE_WITH_ID
     value: "FALSE"
   - name: RUNTIME_TRAINING_OUTPUT_DIR
     value: "models"
   - name: RUNTIME_LOCAL_MODELS_DIR
     value: "/app/models"
```

This will lead to the following being printed as the last line of the stack:

```
  File "/usr/local/lib/python3.10/site-packages/caikit/core/exceptions/error_handler.py", line 384, in value_check
    self(
  File "/usr/local/lib/python3.10/site-packages/caikit/core/exceptions/error_handler.py", line 130, in log_raise
    raise exception
ValueError: value check failed: Must set runtime.local_models_dir with runtime.lazy_load_local_models
```

This is a confusing message because the user has specified both `RUNTIME_LOCAL_MODELS_DIR` and `RUNTIME_LAZY_LOAD_LOCAL_MODELS`. The actual cause of the failure is that the value for the former (`/app/models`) doesn't exist and it needs to if `RUNTIME_LAZY_LOAD_LOCAL_MODELS` is set to true. The user does see 

```{"channel": "MODEL-MANAGR", "exception": null, "level": "warning", "log_code": "<RUN53709826W>", "message": "Invalid runtime.local_models_dir /app/models. Does not exist", "num_indent": 0, "thread_id": 140593866667840, "timestamp": "2023-10-26T15:21:46.223266"}``` in the log file but seeing RUN44773514E as the final entry obscures the fact that the above warning is actually fatal.

## Describe the solution you'd like

Clearer error or warning messaging about the requirement that `RUNTIME_LOCAL_MODELS_DIR` needs to be present.

waleedqk:
Is the required solution here that the error message should be more expressive, signifying that the `_local_models_dir` while set, is not a valid dir path?

ssiegel95:
[...]

PR #591, opened 2023-11-29
lazy_load_local_models obscures error update

waleedqk:
fix [issue545](https://github.com/caikit/caikit/issues/545)

The`ValueError` message seen as the last statement, to provide more clarity towards the root cause of the issue.  

if `RUNTIME_LOCAL_MODELS_DIR` is not a valid path, then `RUNTIME_LAZY_LOAD_LOCAL_MODELS` set to `True` will result in the error. The existing message causes confusion at first glance, since the user has set both parameters, but the path does not exist.


prashantgupta24 (changes requested):
Couple of things:
- Can you add a test?
- Can you add a comment on the nested condition?

tests/examples/test_examples.py
```diff
@@ -110,3 +114,42 @@ def test_example_sample_lib():
 
             # Client worked well, let's stop the server
             server.terminate()
+
+
+def test_lazy_load_local_models_invalid_model_dir():
```
prashantgupta24, line 119: `examples` is not a good place to add this test. I'm assuming `test_model_manager` to test `model_manager`?

tests/examples/test_examples.py
```diff
@@ -110,3 +114,42 @@ def test_example_sample_lib():
...
+
+    with TemporaryDirectory() as workdir:
+
+        caikit.config.configure(
+            config_dict={
+                "merge_strategy": "merge",
+                "runtime": {
+                    "library": "sample_lib",
+                    "local_models_dir": invalid_local_models_dir,
+                    "lazy_load_local_models": True,
```
prashantgupta24, line 136: can you add 4 tests with `local_models_dir` false and true, `lazy_load_local_models` false and true.

tests/runtime/model_management/test_model_manager.py
```diff
@@ -30,6 +31,7 @@
 from caikit import get_config
 from caikit.core.model_manager import ModelManager as CoreModelManager
 from caikit.core.modules import ModuleBase
+from caikit.runtime.dump_services import dump_grpc_services, dump_http_services
```
prashantgupta24, line 34: not needed?

tests/runtime/model_management/test_model_manager.py
```diff
@@ -1054,3 +1056,126 @@ def test_lazy_load_handles_temporary_errors():
...
+                    "library": "sample_lib",
+                    "local_models_dir": cache_dir,
+                    "lazy_load_local_models": True,
+                    "grpc": {"enabled": True},
+                    "http": {"enabled": True},
+                    "training": {"save_with_id": False, "output_dir": cache_dir},
+                    "service_generation": {
+                        "package": "caikit_sample_lib"
+                    },  # This is done to avoid name collision with Caikit itself
+                },
```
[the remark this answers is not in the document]
prashantgupta24: same for all tests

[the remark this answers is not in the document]
prashantgupta24: I also feel the tests can use the signature defined within `test_model_manager_loads_local_models_on_init` instead of the `non_singleton_model_managers`.

tests/runtime/model_management/test_model_manager.py
```diff
@@ -1054,3 +1055,97 @@ def test_lazy_load_handles_temporary_errors():
...
+
+    with TemporaryDirectory() as cache_dir:
+
+        ModelManager._ModelManager__instance = None
+        with temp_config(
+            {
+                "runtime": {
+                    "local_models_dir": cache_dir,
+                    "lazy_load_local_models": True,
+                    "training": {"save_with_id": False, "output_dir": cache_dir},
```
prashantgupta24, line 1073: why do we need `training`?

waleedqk: Good question. I think we dont, so will take it out
Was trying to mimic the https://github.com/caikit/caikit/issues/545 env specifically

waleedqk:
@prashantgupta24 any suggestions as to why the builds are failing with the `RuntimeError: MODULE_ID ... conflicts for classes 'BoundingBoxModule' and 'BoundingBoxModule'`.
Updating the tests only in the latest commits, so unsure why this is coming up now ?

prashantgupta24:
@waleedqk no idea, does it happen locally as well?

prashantgupta24:
@waleedqk seems like that error is happening with other PRs as well

tests/runtime/model_management/test_model_manager.py
```diff
@@ -1054,3 +1055,91 @@ def test_lazy_load_handles_temporary_errors():
...
+            merge_strategy="merge",
+        ):
+            MODEL_MANAGER = ModelManager()
+            assert len(MODEL_MANAGER.loaded_models) == 0
+            assert MODEL_MANAGER._local_models_dir == cache_dir
+
+
+def test_lazy_load_true_local_models_dir_invalid():
+    """When lazy_load_local_models is True and local_models_dir does not exist.
+    Raise ValueError with an appropriate message
```
prashantgupta24, line 1084: nit: if you mention message in the description, then probably better to assert that the exception has the message too

prashantgupta24 (changes requested):
Looking good! Some more commends

caikit/runtime/model_management/model_manager.py
```diff
@@ -114,8 +114,12 @@ def __init__(self):
         self._lazy_load_local_models = runtime_cfg.lazy_load_local_models
         error.value_check(
             "<RUN44773514E>",
-            not self._lazy_load_local_models or self._local_models_dir,
-            "Must set runtime.local_models_dir with runtime.lazy_load_local_models",
+            # _lazy_load_local_models is set, and _local_models_dir is empty
+            not (self._lazy_load_local_models and (not self._local_models_dir)),
+            (
```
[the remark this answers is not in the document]
waleedqk: It is the same condition as before. was modified for ease of understanding. will revert it

tests/runtime/model_management/test_model_manager.py
```diff
@@ -1054,3 +1055,99 @@ def test_lazy_load_handles_temporary_errors():
...
+    """
+
+    with TemporaryDirectory() as cache_dir:
+
+        with pytest.raises(
+            ValueError,
+            match=(
+                "runtime.local_models_dir must be a valid path if set with"
+                " runtime.lazy_load_local_models"
```
prashantgupta24, line 1093: nit: can we check `Provided path` too?

waleedqk: Do you mean update the patch to have the path: As part of this test, it will be an empty string to nothing will really change in the test match string

caikit/runtime/model_management/model_manager.py
```diff
@@ -114,8 +114,12 @@ def __init__(self):
...
+                "runtime.local_models_dir must be a valid path"
+                " if set with runtime.lazy_load_local_models"
+            ),
```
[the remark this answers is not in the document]
prashantgupta24: Also can this whole message not be a single string?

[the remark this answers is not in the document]
prashantgupta24: Also based on the code [before this](https://github.com/caikit/caikit/blob/main/caikit/runtime/model_management/model_manager.py#L105-L111), `self._local_models_dir` will always be an empty string?

tests/runtime/model_management/test_model_manager.py
```diff
@@ -1054,3 +1055,99 @@ def test_lazy_load_handles_temporary_errors():
...
+            ),
+        ) as exec_info:
+
+            ModelManager._ModelManager__instance = None
+            with temp_config(
+                {
+                    "runtime": {
+                        "local_models_dir": "",
```
prashantgupta24, line 1101: Going back on my earlier suggestion, an empty string might not be the best test case here. Probably better to have it something like
```
"local_models_dir": "invalid_dir"
```
and then assert that the exception has `invalid_dir`

prashantgupta24 (changes requested):
Thanks for the changes! Couple more comments

tests/runtime/model_management/test_model_manager.py
waleedqk: Yes, if the path is invalid, the code you linked to will update the config var to empty.
But possible to get around that with the latest changes

caikit/runtime/model_management/model_manager.py
```diff
@@ -114,8 +114,13 @@ def __init__(self):
         self._lazy_load_local_models = runtime_cfg.lazy_load_local_models
         error.value_check(
             "<RUN44773514E>",
+            # _lazy_load_local_models is set, and _local_models_dir is empty
             not self._lazy_load_local_models or self._local_models_dir,
-            "Must set runtime.local_models_dir with runtime.lazy_load_local_models",
+            (
+                "runtime.local_models_dir must be a valid path"
+                " if set with runtime.lazy_load_local_models. "
```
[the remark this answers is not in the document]
prashantgupta24: But a test case for the same would still be helpful!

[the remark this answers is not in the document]
prashantgupta24: The original error statement was handling that use case - 
"Must set runtime.local_models_dir with runtime.lazy_load_local_models" -> we just need to make sure that we include that too. So it needs to both be set AND be valid if `lazy_load_local_models` is True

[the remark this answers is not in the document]
prashantgupta24: I can totally see 2 different error conditions for the 2 scenarios? 🤔 Let me know what you think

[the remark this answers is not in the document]
waleedqk: In my test cases, when I do not set `local_models_dir` it is defaulting to `models` in the logic. 
So its picking it up from [somewhere](https://github.com/caikit/caikit/blob/6677470dd6b1896a7705e744277ae1fd5bf3a990/caikit/config/config.yml#L95)  

Which makes it difficult to assume, if the user did not set it or its default. I can delete that entry and then have logic, that checks if `local_models_dir` was not set. 
Is that the scenario we want, then I can make that change. But it will affect the config in the code base which does not make sense. 
Is there an alternative config I can add to the repo to load as part of the test?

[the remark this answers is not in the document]
prashantgupta24: Hmm, the test config should be isolated from the base config and it should not be defaulting to `models`. Try to see where that's happening?

tests/runtime/model_management/test_model_manager.py
```diff
@@ -1054,3 +1055,131 @@ def test_lazy_load_handles_temporary_errors():
...
+    Check that the local_models_dir is False / Empty
+    """
+
+    with TemporaryDirectory() as cache_dir:
+
+        ModelManager._ModelManager__instance = None
+        with temp_config(
+            {
+                "runtime": {
+                    # "local_models_dir": "",
```
prashantgupta24, line 1177: Worth removing the line?

tests/runtime/model_management/test_model_manager.py
```diff
@@ -1054,3 +1055,131 @@ def test_lazy_load_handles_temporary_errors():
...
+                    "runtime": {
+                        "local_models_dir": "invalid",
+                        "lazy_load_local_models": True,
+                    }
+                },
+                merge_strategy="merge",
+            ):
+                MODEL_MANAGER = ModelManager()
+
+        assert "runtime.local_models_dir must be a valid path" in str(exec_info.value)
```
prashantgupta24, line 1110: any reason for duplicating the `match` value as well as asserting here?

caikit/runtime/model_management/model_manager.py
```diff
@@ -112,10 +112,20 @@ def __init__(self):
...
         self._lazy_load_local_models = runtime_cfg.lazy_load_local_models
+        error_message = (
+            "runtime.local_models_dir must be a valid path"
+            " if set with runtime.lazy_load_local_models. "
+        )
+        if runtime_cfg.local_models_dir is None:
+            error_message += "runtime.lazy_load_local_models is unset in config file"
+        elif not self._local_models_dir:
+            error_message += f"Provided path: {runtime_cfg.local_models_dir}"
+
```
prashantgupta24, line 123: This might be neater if we removed the `value_check` entirely and raised exceptions based on these conditions itself?

tests/runtime/model_management/test_model_manager.py
prashantgupta24, line 1110: any reason for duplicating the `match` value as well as asserting here?

tests/runtime/model_management/test_model_manager.py
prashantgupta24, line 1177: Any reason for keeping the line but commented out?

prashantgupta24 (changes requested):
I see a bit of duplicated conditions

caikit/runtime/model_management/model_manager.py
waleedqk: I can see it making sense having two cases, but since the root is still the same variable value, having it compact like this seems reasonable. 
The combination of the two env vars will raise the error, and the fix is to set the env var. The message error will reflect what type of change to make

tests/runtime/model_management/test_model_manager.py
waleedqk: I may have misunderstood, but I thought this was what you wanted based on previous feedback.
https://github.com/caikit/caikit/pull/591#discussion_r1416161055

prashantgupta24: Yes but from what I understand but doesn't the `match` value at the beginning of the test do that already?

caikit/runtime/model_management/model_manager.py
prashantgupta24: My only concern with this is that if in the future, if the condition to throw an error changes or another condition needs to be added, then it has to be changed in 2 places, one in the block above and one within the `error.value_check`?

prashantgupta24:
```
if self._lazy_load_local_models:
  if runtime_cfg.local_models_dir is None:
     # unset - throw value error with unset
   elif self._local_models_dir is None:
     # set but incorrect - throw value error with invalid path
 ```
or
```
if self._lazy_load_local_models:
  error.value_check(runtime_cfg.local_models_dir is not None)
     # unset - throw value error with unset
   error.value_check(not self._local_models_dir):
     # set but incorrect - throw value error with invalid path
```

File: caikit/runtime/model_management/model_manager.py [hunk 15]
```diff
@@ -112,11 +112,27 @@ def __init__(self):
...
+                (
+                    "runtime.local_models_dir must be set"
+                    " if using runtime.lazy_load_local_models. "
+                    "local_models_dir is unset in config file"
+                ),
+            )
+
+            error.value_check(
+                "<RUN44773514E>",
+                not self._local_models_dir == "",
```
prashantgupta24, line 129: should be the same as `not self._local_models_dir`? Python treats empty strings as false

caikit/runtime/model_management/model_manager.py
```diff
@@ -112,11 +112,27 @@ def __init__(self):
...
-        )
+
+        if self._lazy_load_local_models:
+            error.value_check(
+                "<RUN44773525E>",
+                runtime_cfg.local_models_dir is not None,
+                (
+                    "runtime.local_models_dir must be set"
+                    " if using runtime.lazy_load_local_models. "
+                    "local_models_dir is unset in config file"
```
prashantgupta24, line 123: Let's not mention this line since the user can set this value through env variables as well

caikit/runtime/model_management/model_manager.py
```diff
@@ -112,11 +112,27 @@ def __init__(self):
...
+                ),
+            )
+
+            error.value_check(
+                "<RUN44773514E>",
+                self._local_models_dir != "",
```
prashantgupta24, line 129: nit: I liked `not self._local_models_dir` better, python treats empty strings as false

tests/runtime/model_management/test_model_manager.py
```diff
@@ -15,6 +15,7 @@
 from concurrent.futures import Future, ThreadPoolExecutor
 from contextlib import contextmanager
 from functools import partial
+from os import path
```
prashantgupta24, line 18: nit: unnecessary import?

gabe-l-hart:
@prashantgupta24 @waleedqk It looks like this PR was approved but now has conflicts. Can you guys get those resolved and get this merged?

waleedqk:
@gabe-l-hart good to go now I think

tests/runtime/model_management/test_model_manager.py
gabe-l-hart:
```suggestion
```

gabe-l-hart (changes requested):

gabe-l-hart (approved):

PR #593, opened 2023-12-04
Better aborts

joerunde:

This PR updates caikit's runtime concurrency management to allow canceling the work threads in the server's pool instead of spinning up new "DestroyableThreads" for each request. Spinning up a new thread for every request is especially problematic when running "shim models" because:
1. There are often many requests in flight on one caikit instance (100s or even 1000s?) so creating new threads is actually slow and it is possible to run out of threads
2. We often use shims to take advantage of third party runtimes that are very well optimized for compute to maximize either throughput or latency, so we really don't want to add extra overhead in our proxy layer

Draft state as I'm working on some testing and naming, the config will be cleaned up before merge.

The gist in sequence view:

old stuff
new hotness

The two important questions to ask here are:

1. In the case where contexts are not timing out or canceling, does this perform as good as a plain old thread pool?
2. In the case where request contexts are timing out and canceling, does this perform as good as our current abortable actions?

Quick stats I can elaborate on / check in the test code for later:

Test 1: No cancels or aborts, 2k VUs, joe's laptop:
- Throughput (no aborters):              1991 req/s
- Throughput (abortable contexts):  1987 req/s
- Throughput (abortable actions):    1411 req/s

Test 2: 3% requests canceled, 200VUs, joe's laptop:
- Throughput (no aborters):              161 req/s
- Throughput (abortable contexts):  183 req/s
- Throughput (abortable actions):    182 req/s

Test 3: 30% requests canceled, 200 VUs, joe's laptop:
- Throughput (no aborters):              12 req/s
- Throughput (abortable contexts):  88 req/s
- Throughput (abortable actions):    90 req/s

I think those results show the answer to both of those questions is yes.
Caveat: synthetic models run locally, no real world results, ymmv

**If applicable**:


caikit/runtime/work_management/abortable_action.py
```diff
@@ -116,3 +133,128 @@ def do(self):
...
+    ActionAborters and raise exceptions in the working threads if they need to be aborted.
+    
+    This offers a performance advantage over using `AbortableActions`, since only one extra
+    listener thread is created that lives for the whole lifetime of the program. The caveat
+    is that all work to be aborted must be running in a thread that is safe to kill: this would
+    not be safe to use with asyncio tasks running in an event loop.
+    """
+
+    def __init__(self):
+        self._queue = Queue()
```
joerunde, line 149: A queue was the easiest thing to use (and most suggested on stack overflow lol) to allow one interrupter to wait for many contexts to potentially cancel. In practice though it does seem slower (~5ms ish?) to wait on a queue pull than to wait on a threading event (<1ms).

Since this isn't go and you can't `select` over a large set of events, I don't know if there is a more performant way to do this. I had some slightly buggy thing going to wait on a single condition variable that threads could set, but each context also had to stick itself onto a locked stack so that the interrupter would know which contexts to kill when it woke up, and that lock contention was way slower than a queue.

joerunde: oooh, but a custom `Event` implementation using sockets does sound like a great December project!

caikit/runtime/work_management/abortable_context.py
```diff
@@ -0,0 +1,178 @@
...
+    """
+
+    def __init__(self):
+        # Using a SimpleQueue because we don't need the Queue's task api
+        self._queue = SimpleQueue()
+        self._thread = None
+
+        self._context_thread_map: Dict[uuid.UUID, int] = {}
+
+        self._shutdown = True
```
gabe-l-hart, line 73: NIT: The name `_shutdown` makes me think this would start `False`. The usage below makes me think `_running` might be a clearer semantic.

caikit/runtime/work_management/abortable_context.py
```diff
@@ -0,0 +1,178 @@
...
+
+    def start(self):
+        if not self._shutdown:
```
gabe-l-hart, line 76: I think it's safe to assume that `start` will always be called from a single parent thread, but as written, this is not thread safe unless you put a lock around all of the start/stop logic. I'm fine to clarify this with a comment that indicates these functions are NOT thread safe.

caikit/runtime/work_management/abortable_context.py
```diff
@@ -0,0 +1,178 @@
...
+        self._thread.start()
+
+    def stop(self):
+        if self._shutdown:
+            log.debug("ThreadInterrupter already shut down")
+            return
+
+        log.info("Stopping ThreadInterrupter")
+        self._shutdown = True
+        self._queue.put(0)
```
gabe-l-hart, line 91: Why do we need to put this sentinel `0` on the queue? Is this to avoid getting stuck in `_watch_loop`'s call to `self._queue.get()`? This feels like a special sentinel value that should be defined as `self.__class__._SHUTDOWN` or something, but I don't think it's actually ever used on the receiving end. Either way, a comment explaining this `self._queue.put(0)` would be nice.

caikit/runtime/work_management/abortable_context.py
```diff
@@ -0,0 +1,178 @@
...
+
+    def kill(self, context_id: uuid) -> None:
+        # Put this context onto the queue for abortion and immediately return
+        self._queue.put(context_id, block=False)
+
+    def _watch_loop(self):
+        while True:
+            try:
+                log.debug4("Waiting on any work to abort")
+                context_id = self._queue.get()
```
gabe-l-hart, line 108: Might be nice to throw in a log here showing the thread that will be aborted. Since the `get()` call will block, I would not expect it to be too noisy.

caikit/runtime/work_management/abortable_context.py
```diff
@@ -0,0 +1,178 @@
...
+
+                if self._shutdown:
+                    log.debug4("Ending abort watch loop")
```
gabe-l-hart, line 111: NIT: This and the line above _could_ be on a more verbose log level. If this loop were non-blocking, I would _absolutely_ say put them on `debug4` (or omit them entirely), but as written, they should not be overly verbose on a lower level (and might be useful for debugging strange cancellation behavior).

caikit/runtime/work_management/abortable_context.py
```diff
@@ -0,0 +1,178 @@
...
+                    return
+
+                self._kill_thread(context_id)
+
+                # Ensure this context/thread pair is unregistered
+                self.unregister(context_id)
+
+            except Exception:
+                log.warning("Caught exception while running abort loop", exc_info=True)
```
gabe-l-hart, line 120: Is this something we expect to happen, or is it just guarding against errors in the implementation of the loop? I would expect that the loop itself is _explicitly_ exception-free, so you wouldn't need this `try/except` and if you did happen to catch an exception, it would be fatal since it's not anticipated.

caikit/runtime/work_management/abortable_context.py
```diff
@@ -0,0 +1,178 @@
...
+
+    def _kill_thread(self, context_id: uuid.UUID) -> bool:
+        thread_id = self._context_thread_map.get(context_id, None)
+
+        if thread_id:
+            async_exception_result = ctypes.pythonapi.PyThreadState_SetAsyncExc(
```
gabe-l-hart, line 126: I remember there being a really nice big comment about this magic in `AbortableAction`. Can we port that over here?

caikit/runtime/model_management/model_loader.py
```diff
@@ -94,14 +87,9 @@ def load_model(
...
         log.debug2("Loading model %s async", model_id)
-        if aborter is not None:
-            log.debug3("Using abortable action to load %s", model_id)
-            action = AbortableAction(aborter, self._load_module, *args)
-            future_factory = partial(self._load_thread_pool.submit, action.do)
-        else:
-            future_factory = partial(
-                self._load_thread_pool.submit, self._load_module, *args
-            )
+        future_factory = partial(
```
gabe-l-hart, line 90: Ok, I think I understand what's going on here: This class will always be used _inside_ an `AbortableContext` (that may or may not actually be abortable, depending on whether the `aborter`/`interrupter` are real instances).

tests/runtime/conftest.py
```diff
@@ -98,13 +100,18 @@ def sample_inference_service(render_protos) -> ServicePackage:
 
 @pytest.fixture(scope="session")
 def sample_predict_servicer(sample_inference_service) -> GlobalPredictServicer:
-    servicer = GlobalPredictServicer(inference_service=sample_inference_service)
+    interrupter = ThreadInterrupter()
+    interrupter.start()
```
gabe-l-hart, line 104: NIT: Would it make sense for the GPS to take ownership over the `start/stop` logic of the `interrupter`? I suppose the lifecycle of an `interrupter` may be shared between multiple servicers, though.

tests/runtime/work_management/test_abortable_context.py
```diff
@@ -0,0 +1,162 @@
...
+def wait_for_interrupter_to_run(interrupter, timeout=1):
+    """Helper to wait until the interrupter's queue is empty.
+    This should only deadlock if the interrupter's polling thread exits.
+    """
+    start = datetime.datetime.now()
+    while (datetime.datetime.now() - start).total_seconds() < timeout:
+        if interrupter._queue.empty():
+            return
+        time.sleep(0.001)
+
```
gabe-l-hart, line 64: NIT: In other tests, I tend to like putting a comment bar separating `## Helpers #####...` from `## Tests #####...`. Personal preference, but I think it makes the test files more readable and organized.

gabe-l-hart (changes requested):
This is some _very_ nice work Joe. I think I've parsed it all and I have a few places that I think would benefit from some clarification comments since this is likely to be both very 🌶️ code and also quite hard for others to parse without all the context.

caikit/runtime/servicers/global_train_servicer.py
```diff
@@ -297,12 +297,3 @@ def rpc_termination_callback():
             model_name=request.model_name,
             training_id=model_future.id,
         )
-
-    def _load_trained_model(self, model_name: str, model_path: str):
```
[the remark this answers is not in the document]
joerunde: yup, just finding dead code in dark corners!

tests/runtime/conftest.py
joerunde: Yeah, I thought about that but 
- the current billing metric start/stop is a bit awkward
- the interrupting only makes sense in the context of a running server 
- the servers each own their own GPS
- and there may be multiple servicers like the ModelRuntimeServicer, like you said

caikit/runtime/work_management/abortable_context.py
joerunde: ah true

caikit/runtime/work_management/abortable_context.py
```diff
@@ -0,0 +1,178 @@
...
+# First Party
+import alog
+
+# Local
+from caikit.runtime.types.aborted_exception import AbortedException
+
+log = alog.use_channel("ABORT-ACTION")
+
+
+class AbortableContextBase(abc.ABC):
```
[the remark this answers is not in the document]
joerunde: The day has come where @gabe-l-hart says _not_ to abstract!

(I extended this in a test case, but I agree it doesn't need a base just for a test)

caikit/runtime/servicers/global_predict_servicer.py
```diff
@@ -94,6 +97,7 @@ def __init__(
         self,
         inference_service: ServicePackage,
         use_abortable_threads: bool = get_config().runtime.use_abortable_threads,
+        interrupter: ThreadInterrupter = None,
```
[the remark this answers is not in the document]
joerunde: yeah, I could certainly simplify things more and only do the "abort-y" things if `interrupter` is set

caikit/runtime/model_management/model_loader.py
joerunde: Yeah, from some tracing I did it seemed like `aborter` was never even set here, and if it was then it would have been double-wrapping with destroyable threads

caikit/runtime/work_management/abortable_context.py
joerunde: > it would be fatal since it's not anticipated

Yeah... true. I was trying to be more defensive and not crash anything if something fails in the abort process, but I guess that if we crash the interrupter here then the server could still keep chugging along 🤷

gabe-l-hart: I think that if the interrupter crashes, we _want_ the server to go down, right? If the server is booted with the interrupter and something in our code happens to be borked and raise an exception, that's a fatal coding problem that should crash the server and trigger a bug report.

joerunde: maybe, but that could also mean very sad users!

I think you're right though that nothing in here should raise, there's really just a queue get, a dict get, and a ctypes call that should all be safe

caikit/runtime/servicers/global_predict_servicer.py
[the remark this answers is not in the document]
joerunde: done- now the only spot the `use_abortable_threads` config is used is to initialize an interrupter

caikit/runtime/servicers/global_predict_servicer.py
```diff
@@ -250,7 +253,7 @@ def predict_model(
                 inference request
         """
 
-        with self._handle_predict_exceptions(model_id, request_name):
+        with (self._handle_predict_exceptions(model_id, request_name)):
```
gabe-l-hart, line 256: Curious why the extra parens are needed?

tests/runtime/http_server/test_http_server.py
```diff
@@ -486,20 +486,6 @@ def test_inference_other_task(other_task_model_id, client):
     assert json_response["farewell"] == "goodbye: world 42 times"
 
 
-def test_inference_no_aborter(other_task_model_id, client):
```
gabe-l-hart, line 489: Why is this removed?

gabe-l-hart (approved):
Ship it! Two little questions left, but nothing to stop the train

caikit/runtime/servicers/global_predict_servicer.py
joerunde: idk man I just ran `black` lol

tests/runtime/http_server/test_http_server.py
joerunde: There's another test that does the same thing: `test_inference_other_task`
🤷

caikit/runtime/servicers/global_predict_servicer.py
joerunde: aaaand if I remove + rerun, they don't reappear. Odd

PR #594, opened 2023-12-04
Add MultiModelInitializer Implementation

HonakerM:
**What this PR does / why we need it**:
This PR adds a `MultiModelInitializer` implementation that is very similar to the existing MultiModelFinder except that it works with initializers. This will support future initializers that might need a backup/default override. 

**Special notes for your reviewer**:

**If applicable**:
- [ ] this PR contains documentation
- [ ] this PR contains unit tests
- [ ] this PR has been tested for backwards compatibility

PR #597, opened 2023-12-06
Propose RemoteModelInvocation ADR

HonakerM:
This PR creates an ADR for remote model invocation. For a full list of reasoning and description of the proposed solution take a look at the ADR; however, the first paragraph of the description is copied here for convenience:

> We propose implementing a system for remote model invocation directly in the Caikit core library. 
This system will handle discovering and loading models from a remote runtime. These remote modules 
will look and function like local modules, except any task or train invocation is forwarded to the 
remote. The core components of this system will be a new ModelFinder named RemoteModelFinder and a 
new ModelInitializer called RemoteModelInitializer.

This ADR assumes the functionality provided by [PR590](https://github.com/caikit/caikit/pull/590) and [PR594](https://github.com/caikit/caikit/pull/594)

**If applicable**:

### DIFF
--- a/docs/adrs/024-remote-module-invocation.md
+++ b/docs/adrs/024-remote-module-invocation.md
@@ -44,13 +44,14 @@ model_management:
             type: REMOTE
             config:
                 connection:
-                    host: str <remote host>
+                    hostname: str <remote host>
                     port: int <remote port>
                     protocol: Optional[str]="grpc" <protocol the remote server is using (grpc or http)>
                     tls:
                         enabled: Optional[bool]=False <if ssl is enabled on the remote server>
-                        ca: Optional[str]=None <path to remote ca file>
-                        cert: Optional[str]=None <path to MTLS cert>
+                        ca_file: Optional[str]=None <path to remote ca file>
+                        cert_file: Optional[str]=None <path to MTLS cert>
+                        key_file: Optional[str]=None <path to MTLS key>
                     options:  Optional[Dict[str,str]]={} <optional dict of grpc or http configuration options>
                 discover_models: Optional[bool]=True <bool to automatically discover remote models via the /info/models endpoint> 
                 supported_models: Optional[Dict[str, str]]={} <mapping of model names to module_ids that this remote supports>
@@ -59,14 +60,15 @@ model_management:
 
 The proposed configuration for the RemoteModelFinder is above. The only required field is the 
 generic `connection` dictionary that supports a secure channel, mutual TLS, and custom GRPC/HTTP 
-options. The `connection.host` setting contains the remote's hostname, while `connection.port` determines the 
+options. The `connection.hostname` setting contains the remote's hostname, while `connection.port` determines the 
 runtime port. The optional `connection.protocol` config is used to select which protocol to send
 requests over, with the default being `grpc`. The `connection.tls` dictionary contains all information 
-related to TLS with `tls.enabled` controlling if the server is running SSL, `tls.ca` is the path to
-the CA file that the remote's certificate is signed by, and finally, `tls.cert` is the path to the
-MTLS client certificate, which will be sent with the request. The final connection config is `connection.options` 
-which defines a list of options to pass to either the HTTP or GRPC request; for an example of options,
-take a look at the [GRPC Channel options](https://grpc.github.io/grpc/core/group__grpc__arg__keys.html#details)
+related to TLS with `tls.enabled` controlling if the server is running SSL, `tls.ca_file` is the path to
+the CA file that the remote's certificate is signed by, `tls.cert_file` is the path to the
+MTLS client certificate to be sent with the request, and finally, `tls.key_file` which is the file 
+containing the MTLS client key. The final connection config is `connection.options` which defines a 
+list of options to pass to either the HTTP or GRPC request; for an example of options, take a look 
+at the [GRPC Channel options](https://grpc.github.io/grpc/core/group__grpc__arg__keys.html#details)
 
 
 Two additional optional fields help control what models this remote supports. The 
@@ -260,7 +262,7 @@ def generate_config_for_module(module: ModuleBase, connection_info: Dict[str, An
 
 ### Diagram
 
-<img width="1439" alt="image" src="https://github.com/HonakerM/MTConnect-Python-Agent/assets/37811263/70563a23-8439-4221-b81d-c242f638a579">
+<img width="1439" alt="image" src="https://github.com/caikit/caikit/assets/37811263/4e0d0573-04fa-42fa-bdce-fcd989b9bbd6">
 This is an updated block diagram of the various model loading components and their relationships. 
 
 
--- a/docs/adrs/024-remote-module-invocation.md
+++ b/docs/adrs/024-remote-module-invocation.md
@@ -268,7 +268,7 @@ This is an updated block diagram of the various model loading components and the
 
 ## Status
 
-choose one: Proposed
+choose one: Accepted
 
 if deprecated, include a rationale.
 
--- a/docs/adrs/024-remote-module-invocation.md
+++ b/docs/adrs/024-remote-module-invocation.md
@@ -1,0 +1,280 @@
+# ADR 024 Remote Module Invocation
+
+Currently, caikit only supports locally loaded modules and does not understand a remote runtime 
+server. Remote servers can allow for a better distribution of resources and more complex runtime 
+architectures. Remote modules could also form the basis of a ["thin client"](https://github.com/caikit/caikit/issues/255) 
+where a service can run remote invocations without installing library dependencies.
+Library users have started creating custom initializers to solve these problems; however, 
+those implementations depend on specific assumptions about the caikit runtime that might change 
+from version to version.
+
+## Decision
+
+We propose implementing a system for remote model invocation directly in the Caikit core library. 
+This system will handle discovering and describing models from a remote runtime. These remote modules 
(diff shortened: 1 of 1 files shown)


docs/adrs/024-remote-module-invocation.md
```diff
@@ -0,0 +1,280 @@
...
+any global configuration settings, as the remote information will be passed in via the 
+RemoteModelFinder.  If a system is expected to have both local and remote models, consider using a 
+MultiModelInitializer to handle both use cases.
+
+To help illustrate how the RemoteModelInitializer would initialize a Module, we provided a snippet
+of pseudo Python code:
+```python3
+def init(model_config: RemoteModuleConfig)->ModuleBase:
+    # Construct empty RemoteModule Instance
+    @module(
```
gabe-l-hart, line 120: I think this cause problems if the IDs conflict with local modules that _do_ exist. There's nothing that requires the `Module` instance returned by a `ModelInitializer` to return an instance whose `MODULE_ID` matches the id of the config that got passed in.

docs/adrs/024-remote-module-invocation.md
```diff
@@ -0,0 +1,280 @@
...
+
+    # Only used for infer RPC types
+    input_streaming: bool
+    output_streaming: bool
+```
+
+The RemoteModuleConfig is a custom subclass of ModuleConfig, which contains a description of a Module's
+tasks, inference and train methods, and version information, as well as the connection information of 
+the remote runtime. The combination of the two allows the RemoteModelInitializer to construct a new 
+RemoteModule without having to import `caikit.runtime` or the source model. To help simplify the
```
gabe-l-hart, line 194: Ah! I forgot that we had moved `CaikitMethodSignature` to `core`. This makes it really nice and clean.

docs/adrs/024-remote-module-invocation.md
```diff
@@ -0,0 +1,280 @@
...
+config definition and access, we decided to create a helper dataclass, `RemoteMethodRpc`, which contains
+information about a specific method and includes things like the CaikitMethodSignature, 
+request&response DataModel names, and the remote RPC name. The `RemoteMethodRpc` dataclass contains 
+all the runtime-specific assumptions  and is the main point where overlap happens. The 
+`RemoteModelFinder`, which constructs `RemoteMethodRpc`s, should not import `caikit.runtime`; 
+however, it will re-use a lot of code around the different naming schemes.
```
gabe-l-hart, line 200: It would be nice if that could be refactored so that these naming schemes can live in one place (and eventually be made less brittle)

docs/adrs/024-remote-module-invocation.md
```diff
@@ -0,0 +1,280 @@
...
+            "module_id": module.MODULE_ID,
+            "module_name": module.MODULE_NAME,
+        }
+    )
+        
+```
+
+### Diagram
+
+<img width="1439" alt="image" src="https://github.com/HonakerM/MTConnect-Python-Agent/assets/37811263/70563a23-8439-4221-b81d-c242f638a579">
```
gabe-l-hart, line 263: This is a really nice diagram! I think the one thing that feels a little confusing about it is how the `remote` versions seem like they're their own set of sequential things. I know that's not the intent, but it would be good to clarify that the `finder`/`initializer` boundaries are still encapsulated, so it would be critical for a given model to only be found by a single `finder` and in the case of the `RemoteModelFinder`, that would need to happen IFF the `RemoteModelInitializer` would be guaranteed to actually be able to initialize it.

gabe-l-hart (commented):
Some questions/comments. Overall, this is a really nice proposal and I think we should get to work getting it in soon!

docs/adrs/024-remote-module-invocation.md
```diff
@@ -0,0 +1,280 @@
...
+    # Gather a description of all tasks and their associated methods
+    task_methods = []
+    for task_class in module.tasks:
+        task_functions = []
+        for input, output, signature in module_class.get_inference_signatures(task_class):
+            # Construct request_dm_name,  task_request_name, and rpc_name. This makes assumptions
+            # about caikit.runtime service generation
+            request_dm_name ~= "TaskRequest"
+            task_request_name ~= "TaskPredict"
+            rpc_name = "/api/v1/TaskPredict"
```
joerunde, line 216: Can we avoid the coupling-by-naming-scheme by having the remote model finder lookup this information as well?

We could add the supported tasks and relevant names to the new `ModelInfoResponse`, and/or we could return the supported rpc names and use the grpc reflection api to go look up the names and schemas of the request and response.

docs/adrs/024-remote-module-invocation.md
```diff
@@ -0,0 +1,280 @@
...
+
+## Decision
+
+We propose implementing a system for remote model invocation directly in the Caikit core library. 
+This system will handle discovering and describing models from a remote runtime. These remote modules 
+will look and function like local modules, except any task or train invocation is forwarded to the 
+remote. The core components of this system will be a new ModelFinder named RemoteModelFinder and a 
+new ModelInitializer called RemoteModelInitializer.
+
+The RemoteModelFinder will gather the remote server's connection and model information. In the first
```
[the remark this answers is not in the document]
HonakerM: > Is the idea here that the RemoteModelFinder will need to explicitly be given the connection information

Correct each finder will have it's own connection information. The idea being that models could be split between multiple remote runtimes.


> Specifically, there's no way to couple a finder and an initializer
Yeah, that was one of my main issues when trying to design this. My solution was to create a new ModuleConfig subclass  named `RemoteModuleConfig` and the `RemoteModelInitializer` will only accept that type of config, similarly the `LocalModelInitializer` will only accept the traditional ModuleConfigs.

docs/adrs/024-remote-module-invocation.md
```diff
@@ -0,0 +1,280 @@
...
+
+
+```yaml
+model_management:
+    finders:
+        <finder name>:
+            type: REMOTE
+            config:
+                connection:
+                    host: str <remote host>
```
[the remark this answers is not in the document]
HonakerM: @gabe-l-hart good call! I'll update the ADR

docs/adrs/024-remote-module-invocation.md
```diff
@@ -0,0 +1,280 @@
...
+
+Two additional optional fields help control what models this remote supports. The 
+`discover_models` setting is a boolean that controls if the finder should query the remote runtime
+to dynamically discover what models are loaded and their corresponding `module_id`'s. The
+`supported_models` config is a dictionary that contains a static mapping of model_paths to module_ids 
+that the remote supports. The `supported_models`  setting is required to add support for remotes that
+don't have a reflection api or ones that lazily load their models (like ModelMesh). 
+
+
+To help illustrate the above config, we included some pseudo python code to illustrate what happens
```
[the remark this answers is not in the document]
HonakerM: >  I think it would be possible to include these configurations with a TGISAutoFinder
What do you mean by this? Like move the TGIS's connection info from the Finder to the backend?

> the RemoteModelFinder only returned non-None for models that are actually supported in the remote 
Yep that was the idea! The `RemoteModelFinder` will only return a `RemoteModuleConfig` if it's l>ocated both the remote model and local ModuleBase.

docs/adrs/024-remote-module-invocation.md
HonakerM: @gabe-l-hart so my idea for this was actually to use a modified module_id and not the exact same one e.g. `f"{original_module_id}-remote"`. That way you could have both Module's coexist

docs/adrs/024-remote-module-invocation.md
HonakerM: Yeah... it would be great if all of this could live in interfaces or something

docs/adrs/024-remote-module-invocation.md
joerunde: 🌶️ , yeah that's still probably the easiest way to get this done. If these schemes changed then it would be a headache, but that's just a "what if"

HonakerM:
@gabe-l-hart I'm planning to start working on the implementation of this ADR next week. Is this good to go? If so should this PR be merged or wait for implementation?

docs/adrs/024-remote-module-invocation.md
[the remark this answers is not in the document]
gabe-l-hart: I was just trying to figure out in my head if `TGISAutoFinder` and `RemoteModelFinder` could co-exist. I think they can!

docs/adrs/024-remote-module-invocation.md
gabe-l-hart: Interesting, that would be nice. At some point, we had talked about enforcing a module ID to be a `uuid`, but that never got implemented, and personally I kind of like having some meaningful text in a module ID

docs/adrs/024-remote-module-invocation.md
```diff
@@ -0,0 +1,282 @@
...
+
+### Diagram
+
+<img width="1439" alt="image" src="https://github.com/caikit/caikit/assets/37811263/4e0d0573-04fa-42fa-bdce-fcd989b9bbd6">
+This is an updated block diagram of the various model loading components and their relationships. 
+
+
+## Status
+
+choose one: Proposed
```
gabe-l-hart, line 271:
```suggestion
choose one: Accepted
```

gabe-l-hart (approved):
Looks good! Commit that change to `Accepted` an let's merge it

PR #601, opened 2023-12-07
Configurable load wait 600

### LINKED ISSUE #600: Allow initial model loading to happen in the background in runtime

For some usecases, an instance of the runtime may boot up with a large number of models in `local_models_dir`. The current boot up logic will sequentially load each model before booting the server which can take a very long time, leading to health/readiness probe failures and `CrashLoopBackoff` if the probes are not given a very long initial delay. Furthermore, the number of models may grow over time, so a valid initial delay may become invalid as the number of models grows.

The requested feature here is to allow [the initial local_models_dir_sync](https://github.com/caikit/caikit/blob/main/caikit/runtime/model_management/model_manager.py#L152) to not wait for model loads to complete. This should be configurable so that different usecases can set the behavior to suit.

* Add a boolean config argument for `runtime.wait_for_initial_model_loads`
* Use the config argument when performing the initial `sync`

We could update the sync logic to parallelize the loads to make the initial loads happen in parallel. This could cause the periodic syncs to suck up too much CPU and degrade the performance of inference requests when the sync runs.

## Additional context

This is particularly relevant when the models are held in a mounted volume in `kubernetes` where the volume driver is slow to initialize individual download streams (e.g. `s3fs`).

gabe-l-hart:

Closes #600 

**If applicable**:

### DIFF
--- a/caikit/config/config.yml
+++ b/caikit/config/config.yml
@@ -93,8 +93,8 @@ runtime:
     # TODO: replace with libraries below when runtime can support multiple libraries
     library: sample_lib
     local_models_dir: models
-    # When loading models at boot, wait for the load to complete before moving
-    # on to the next model
+    # When loading models at boot, wait for the loads to complete before booting
+    # the server
     wait_for_initial_model_loads: true
 
     # If enabled, the models in local_models_dir will be periodically sync'ed
--- a/caikit/config/config.yml
+++ b/caikit/config/config.yml
@@ -87,10 +87,15 @@ data_streams:
 
 ### Runtime configurations
 runtime:
-    # The runtime library (or libraries) whose models we want to serve using Caikit Runtime. This should
-    # be a snake case string, e.g., caikit_nlp or caikit_cv.
-    library: sample_lib # TODO: replace with libraries below when runtime can support multiple libraries
+    # The runtime library (or libraries) whose models we want to serve using
+    # Caikit Runtime. This should be a snake case string, e.g., caikit_nlp or
+    # caikit_cv.
+    # TODO: replace with libraries below when runtime can support multiple libraries
+    library: sample_lib
     local_models_dir: models
+    # When loading models at boot, wait for the load to complete before moving
+    # on to the next model
+    wait_for_initial_model_loads: true
 
     # If enabled, the models in local_models_dir will be periodically sync'ed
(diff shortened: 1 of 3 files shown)


caikit/config/config.yml
```diff
@@ -87,10 +87,15 @@ data_streams:
...
-    # be a snake case string, e.g., caikit_nlp or caikit_cv.
-    library: sample_lib # TODO: replace with libraries below when runtime can support multiple libraries
+    # The runtime library (or libraries) whose models we want to serve using
+    # Caikit Runtime. This should be a snake case string, e.g., caikit_nlp or
+    # caikit_cv.
+    # TODO: replace with libraries below when runtime can support multiple libraries
+    library: sample_lib
     local_models_dir: models
+    # When loading models at boot, wait for the load to complete before moving
+    # on to the next model
```
joerunde, line 97: should this say `wait for the load to complete before starting the server`?

joerunde: ah, jk if I'm reading correctly then the loads are currently serial if `wait=True`?

joerunde: oh or not! The `wait` we pass to `self.load_model` is always false, then this `wait` waits for all of them to load

🤯 !

caikit/runtime/model_management/model_manager.py
```diff
@@ -148,8 +148,13 @@ def __init__(self):
...
         if self._local_models_dir:
-            log.info("<RUN44739400I>", "Loading local models into Caikit Runtime...")
-            self.sync_local_models(wait=True)
+            wait = runtime_cfg.wait_for_initial_model_loads
+            log.info(
+                "<RUN44739400I>",
+                "Loading local models into Caikit Runtime. Wait: %s",
+                wait,
+            )
+            self.sync_local_models(wait=wait)
```
joerunde, line 157: should the `wait` parameter here be renamed to not be confused with the `wait` on `load_model` that waits for a single load, where this one waits for all loads to finish?

I'd lean no, but just bringing it up. We probably could update the docstring on `sync_local_models` to elaborate a bit that the model loads happen in parallel and the `wait` waits on all of them after queueing them all up

gabe-l-hart: Yeah, I think elaborating on the docstring is the right thing. I'll do that

PR #602, opened 2023-12-07
:zap: Share thread pool across servers

joerunde:

This PR both:
- Explicitly sets a thread pool to use for `loop.run_in_executor` calls in the http server that map sync model code to async event loops. This is required to give users the flexibility to specify large thread pools for IO bound models (e.g. remote shims)
- Shares a single thread pool across all servers, so that an http and grpc server running in the same process would share the same access to threads

The new `runtime.server_thread_pool_size` config option sets the size of this pool, but if the older `runtime.grpc.server_thread_pool_size` is set, it will take precedence.

**If applicable**:

### DIFF
--- a/caikit/config/config.yml
+++ b/caikit/config/config.yml
@@ -111,6 +111,9 @@ runtime:
     # Number of threads to make available for load jobs (null => all)
     load_threads: null
 
+    # Number of threads for both the http and grpc server to share for servicing requests
+    server_thread_pool_size: 100
+
     # TLS configs
     tls:
         server:
@@ -123,8 +126,6 @@ runtime:
     grpc:
         enabled: true
         port: 8085
-        # Number of workers with which we will run the gRPC server
-        server_thread_pool_size: 5
         # gRPC Server shutdown grace period
         # the server shuts down immediately when stopped,
         # and all RPCs active at the end of the grace period are aborted
--- a/caikit/runtime/server_base.py
+++ b/caikit/runtime/server_base.py
@@ -12,8 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """Base class with common functionality across all caikit servers"""
-
 # Standard
+from concurrent.futures import ThreadPoolExecutor
 from typing import Optional
 import abc
 import signal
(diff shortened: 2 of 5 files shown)


caikit/runtime/server_base.py
```diff
@@ -27,12 +27,38 @@
...
+
+class ServerThreadPool:
+    """Simple wrapper for all servers to share a single thread pool"""
+
+    @staticmethod
+    def _build_pool() -> ThreadPoolExecutor:
+        config = caikit.get_config()
+        # Leave in backwards compatibility for the old runtime.grpc.server_thread_pool_size
+        # parameter, which many users may have deployed with.
+        if pool_size := config.runtime.grpc.get("server_thread_pool_size", None):
```
gabe-l-hart, line 48: Two NITs:

1. `config.runtime.grpc.server_thread_pool_size` is exactly the same as `config.runtime.grpc.get("server_thread_pool_size", None)` because `aconfig` defines `getattr` as `get(key, None)`
2. Since this field is removed from `config.yml`, the most likely way that this backwards-compatible flag would be set (via an env var) won't work. The env vars are only read for keys that exist in the config looking for overrides.

caikit/config/config.yml
```diff
@@ -123,8 +126,6 @@ runtime:
     grpc:
         enabled: true
         port: 8085
-        # Number of workers with which we will run the gRPC server
-        server_thread_pool_size: 5
```
gabe-l-hart, line 127: I think we need to keep this in so that env var overrides work. We can default it to `null` or `0` though.

gabe-l-hart (changes requested):
I think we need to keep the config value around so that env overrides work

caikit/config/config.yml
joerunde: 🌶️

caikit/runtime/server_base.py
joerunde: (Been doing too much defensive coding with @mynhardtburger these days!)

gabe-l-hart (approved):
Looks great, thanks Joe!

PR #607, opened 2023-12-11
:bug: ensure correct shutdown handlers called

joerunde:

This was a tangent I went on today. This PR both:
- allows multiple servers to cleanly register shutdown handlers for interruption
- Updates the http server hackery to ensure that previously registered signal handlers are actually invoked when uvicorn overrides all the signal handlers

This is a bit of a departure from how we were explicitly setting signal handlers in `main` methods before, but this allows us to easily handle uvicorn's insistence on overriding all signal handlers.

**If applicable**:

### DIFF
--- a/caikit/runtime/__main__.py
+++ b/caikit/runtime/__main__.py
@@ -1,6 +1,16 @@
-# Standard
-import signal
-
+# Copyright The Caikit Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
 # First Party
 import alog
 
@@ -13,25 +23,6 @@ log = alog.use_channel("RUNTIME-MAIN")
 def main():
     _grpc_server = None
     _http_server = None
-
-    def interrupt(signal_, _stack_frame):
-        log.info(
-            "<RUN87630120I>",
-            "Caikit Runtime received interrupt signal %s, shutting down",
-            signal_,
-        )
-        if _grpc_server:
-            _grpc_server.stop()
-        if _http_server:
-            _http_server.stop()
-
-    # NOTE: signal function can only be called from main thread of the main
-    # interpreter. If this function is called from a thread (like in tests)
-    # then signal handler cannot be used. Thus, we will only have real
-    # termination_handler when this is called from the __main__.
-
-    signal.signal(signal.SIGINT, interrupt)
-    signal.signal(signal.SIGTERM, interrupt)
 
     #####################
     # Start the servers
--- a/caikit/runtime/http_server/http_server.py
+++ b/caikit/runtime/http_server/http_server.py
@@ -28,6 +28,7 @@ import io
 import json
 import os
 import re
+import signal
 import ssl
 import tempfile
 import threading
@@ -237,17 +238,8 @@ class RuntimeHTTPServer(RuntimeServerBase):
             # deleted if they're ephemeral
             config.load()
 
-        # Start the server with the loaded config
+        # Build the server with the loaded config
         self.server = uvicorn.Server(config=config)
-
-        # Patch the exit handler to call this server's stop
-        original_handler = self.server.handle_exit
-
-        def shutdown_wrapper(*args, **kwargs):
-            original_handler(*args, **kwargs)
-            self.stop()
-
-        self.server.handle_exit = shutdown_wrapper
 
         # Placeholder for thread when running without blocking
         self._uvicorn_server_thread = None
@@ -270,10 +262,13 @@ class RuntimeHTTPServer(RuntimeServerBase):
             self.thread_pool._max_workers,
         )
 
+        # Patch the exit handler to retain correct signal handling behavior
+        self._patch_exit_handler()
+
         if blocking:
             self.server.run()
         else:
-            self.run_in_thread()
+            self._run_in_thread()
 
     def stop(self):
         """Stop the server, with an optional grace period.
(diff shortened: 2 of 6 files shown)


caikit/runtime/__main__.py
```diff
@@ -1,6 +1,16 @@
-# Standard
-import signal
-
+# Copyright The Caikit Authors
```
gabe-l-hart, line 1: Whoops! Good catch

caikit/runtime/http_server/http_server.py
```diff
@@ -282,11 +277,14 @@ def stop(self):
             grace_period_seconds (Union[float, int]): Grace period for service shutdown.
                 Defaults to application config
         """
-        self.server.should_exit = True
```
gabe-l-hart, line 285: Why was this moved? I think it makes sense to put it under the `if`, I'm just curious if having it outside was causing some specific problem

caikit/runtime/server_base.py
```diff
@@ -137,6 +137,43 @@ def interrupt(self, signal_, _stack_frame):
...
+    def _add_signal_handler(sig, handler):
+        def nested_interrupt_builder(*handlers):
+            """Build and return an interrupt handler that calls all of *handlers"""
+
+            log.debug("Building interrupt handler: %s", handlers)
+
+            def interrupt(signal_, _stack_frame):
+                for handler in handlers:
+                    # Only call the handler if it is a callable fn that is _not_ a default handler
+                    log.debug("Running interrupt handler: %s", handler)
```
gabe-l-hart, line 159: Do we have to worry about writing to `stdout` in a signal handler? I know I've seen issues with writing to closed pipes before when logging in `__del__` functions

caikit/runtime/server_base.py
```diff
@@ -137,6 +137,43 @@ def interrupt(self, signal_, _stack_frame):
...
+                        and handler is not signal.default_int_handler
+                    ):
+                        handler(signal_, _stack_frame)
+
+            return interrupt
+
+        try:
+            signal.signal(sig, nested_interrupt_builder(handler, signal.getsignal(sig)))
+        except ValueError:
+            log.warning(
```
gabe-l-hart, line 173: Do we expect this to indicate a problem? Seems like putting this on `warning` would likely cause confusion if this is a "working as expected" usecase when running either of the servers in a child thread.

gabe-l-hart (commented):
A couple of questions about some of the logging. Also, I don't see any unit tests. I expect that this would be quite hard to actually test since it only works with the primary process, but we probably could do something like [this test](https://github.com/caikit/caikit/blob/main/tests/runtime/test_grpc_server.py#L1331) where we launch the server in a subprocess so that we can send it a signal

caikit/runtime/http_server/http_server.py
joerunde: Nah I was just being a bit more specific- it's only needed to signal the server running in another thread to stop

caikit/runtime/server_base.py
joerunde: true, I guess this should be info level

caikit/runtime/server_base.py
joerunde: uhhh, well we were logging before so I kept logging here? But that's a good question.

FWIW the uvicorn handler definitely spits out some info messages about shutting down 🤷 

Think this is relevant? https://docs.python.org/3/library/logging.html#thread-safety

> If you are implementing asynchronous signal handlers using the [signal](https://docs.python.org/3/library/signal.html#module-signal) module, you may not be able to use logging from within such handlers. This is because lock implementations in the [threading](https://docs.python.org/3/library/threading.html#module-threading) module are not always re-entrant, and so cannot be invoked from such signal handlers.

I don't exactly know if by "asynchronous" here they mean `async def ...` or just "asynchronous" in general 🤔

joerunde:
> Also, I don't see any unit tests

Yup! Didn't really have the bandwidth here and I couldn't find the gumption to try to patch in mock handlers that I could assert were called, but have them running in a subprocess...

tests/runtime/test_grpc_server.py
```diff
@@ -1350,6 +1351,63 @@ def test_grpc_sever_shutdown_with_model_poll(open_port):
...
+            RUNTIME_METRICS_ENABLED="false",
+            RUNTIME_GRPC_ENABLED="true",
+            RUNTIME_HTTP_ENABLED="true",
+        )
+        with server_proc as proc:
+            # Wait for the grpc server to be up:
+            _assert_connection(
+                grpc.insecure_channel(f"localhost:{open_port}"), max_failures=500
+            )
+            print("grpc server is up!")
```
gabe-l-hart, line 1378:
```suggestion
```

tests/runtime/test_grpc_server.py
```diff
@@ -1350,6 +1351,63 @@ def test_grpc_sever_shutdown_with_model_poll(open_port):
...
+    """Test that a SIGINT successfully shuts down all running servers"""
+
+    # whoops, need 2 ports. Cannot use `_open_port()` again because it'll return the same port
+    other_open_port = open_port + 37
+
+    with tempfile.TemporaryDirectory() as workdir:
+        server_proc = ModuleSubproc(
+            "caikit.runtime",
+            kill_timeout=30.0,
+            RUNTIME_GRPC_PORT=str(open_port),
```
gabe-l-hart, line 1364: Since this is testing with logs (meh), throw a `LOG_LEVEL` in there so that we guarantee the logs will be configured to print in case the parent invokes the test with logging turned all the way off

gabe-l-hart (changes requested):
Two tiny nits!

tests/runtime/test_grpc_server.py
joerunde: hot

tests/runtime/test_grpc_server.py
joerunde: ahhhhh my debugs lol

tests/runtime/conftest.py
```diff
@@ -69,10 +69,9 @@ def http_session_scoped_open_port():
     return _open_port()
 
 
-def _open_port():
+def _open_port(start=8888):
```
gabe-l-hart, line 72: heh, we should probably move to [tls_test_tools.open_port()](https://github.com/IBM/tls-test-tools/blob/main/tls_test_tools/__init__.py#L39) for consistency, but certainly not now

gabe-l-hart (approved):
Looks good!

PR #612, opened 2023-12-14
Move Request processing under Exception Handling

HonakerM:
This PR moves the pydantic/multi-part form request processing to inside of the `try/catch` runtime section. This allows end-users to raise `CaikitRuntimeExceptions` and have them properly captured and returned.

I'm not sure this is the correct solution for the problem. Please let me know if there is a better alternative 

**If applicable**:

### DIFF
--- a/caikit/runtime/http_server/http_server.py
+++ b/caikit/runtime/http_server/http_server.py
@@ -391,11 +391,11 @@ class RuntimeHTTPServer(RuntimeServerBase):
             log.debug("In unary handler for %s", rpc.name)
             loop = asyncio.get_running_loop()
 
-            # build request DM object
-            request = await pydantic_from_request(pydantic_request, context)
-            http_request_dm_object = pydantic_to_dataobject(request)
-
             try:
+                # build request DM object
+                request = await pydantic_from_request(pydantic_request, context)
+                http_request_dm_object = pydantic_to_dataobject(request)
+
                 call = partial(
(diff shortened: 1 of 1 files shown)


gabe-l-hart (changes requested):
It would be great to have a unit test showing how validation can raise! Otherwise, this looks good to go

caikit/core/exceptions/caikit_core_exception.py
```diff
@@ -36,3 +37,4 @@ class CaikitCoreException(Exception):
     def __init__(self, status_code: CaikitCoreStatusCode, message: str) -> None:
         self.status_code = status_code
         self.message = message
+        self.id = uuid.uuid4().hex
```
gabe-l-hart, line 40: Why is this needed?

gabe-l-hart (commented):
Just one question about the introduction of the `uuid` on the exception instance

caikit/core/exceptions/caikit_core_exception.py
HonakerM: I wanted to match the format of [CaikitRuntimeException](https://github.com/caikit/caikit/blob/70685013943c63b15f57b3526eb1148675185705/caikit/runtime/types/caikit_runtime_exception.py#L44) so they could reuse the same error handlers

gabe-l-hart (approved):
Looks good! Thanks Casey

PR #615, opened 2023-12-14
ModelFetchOutOfAsyncio

gabe-l-hart:

Move the model lookup for inference signature into predict_model

The `retrieve_model` operation can potentially be expensive. With the `async` HTTP server, doing this outside of the `run_in_executor` call meant that the work to look up the model (and potentially wait for it to finish loading) was blocking the event loop. That work to fetch the model was _only_ being done so that the inference function could be deduced, so moving that logic inside of `predict_model` allows the lookup (and load) logic to not block the event loop.

**If applicable**:

### DIFF
--- a/caikit/runtime/servicers/global_predict_servicer.py
+++ b/caikit/runtime/servicers/global_predict_servicer.py
@@ -29,7 +29,7 @@ import alog
 
 # Local
 from caikit import get_config
-from caikit.core import ModuleBase
+from caikit.core import ModuleBase, TaskBase
 from caikit.core.data_model import DataBase, DataStream
 from caikit.core.signature_parsing import CaikitMethodSignature
 from caikit.runtime.metrics.rpc_meter import RPCMeter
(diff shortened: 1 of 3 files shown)


prashantgupta24:
Nice!

caikit/runtime/servicers/global_predict_servicer.py
```diff
@@ -227,8 +232,15 @@ def predict_model(
                 The name of the request message to validate the model's task
             model_id (str):
                 The ID of the loaded model
+            input_streaming (bool):
```
prashantgupta24, line 235: extremely minor nit: `input_streaming` is below `inference_func_name` in the list of arguments

prashantgupta24 (approved):
Looks good! One minor nit

caikit/runtime/servicers/global_predict_servicer.py
gabe-l-hart: Good catch!

PR #618, opened 2023-12-18
fix exception f-string

dtrifiro:
- model sizer: fix exception f-string


gabe-l-hart (approved):
Good catch! Thanks for the fix

PR #610, opened 2023-12-13
Uvicorn config 609

gabe-l-hart:
**What this PR does / why we need it**:

Closes #609 

This PR makes it possible to set arbitrary `uvicorn` server config overrides as long as they don't overlap with the ones that `caikit` manages explicitly.

**If applicable**:
- [ ] this PR contains documentation
- [x] this PR contains unit tests
- [x] this PR has been tested for backwards compatibility

PR #623, opened 2023-12-19
re-expose http server

dtrifiro:
Refactors (#610) removed the host argument to the uvicorn config, causing the http server to only bind to 127.0.0.1.


gabe-l-hart (approved):
Thanks for catching this!

PR #627, opened 2023-12-29
Add Finder/Initializer options to Runtime Model Manager load

HonakerM:
This PR adds two optional parameters `finder` and `initializer` to the `RuntimeModelManager.load_model` function. This allows end users to customize which finder/initializer is being used similarly to the core `MODEL_MANAGER.load` function.

**If applicable**:

### DIFF
--- a/caikit/runtime/model_management/model_loader.py
+++ b/caikit/runtime/model_management/model_loader.py
@@ -26,6 +26,7 @@ import alog
 # Local
 from caikit.config import get_config
 from caikit.core import MODEL_MANAGER, ModuleBase
+from caikit.core.model_management import ModelFinderBase, ModelInitializerBase
 from caikit.runtime.model_management.batcher import Batcher
 from caikit.runtime.model_management.loaded_model import LoadedModel
 from caikit.runtime.types.caikit_runtime_exception import CaikitRuntimeException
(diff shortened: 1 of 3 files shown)


HonakerM:
The py3.8 job is failing on a `caikit_health_probe` test. I didn't touch that file so I think it's just a transient failure

caikit/runtime/model_management/model_loader.py
```diff
@@ -96,14 +99,26 @@ def load_model(
         return model_builder.build()
 
     def _load_module(
-        self, model_path: str, model_id: str, model_type: str
+        self,
+        model_path: str,
+        model_id: str,
+        model_type: str,
+        finder: Union[None, str, ModelFinderBase],
+        initializer: Union[None, str, ModelInitializerBase],
```
gabe-l-hart, line 107: I think these semantics would be a little cleaner for the type hint (same as above)

```suggestion
        finder: Optional[Union[str, ModelFinderBase]],
        initializer: Optional[Union[str, ModelInitializerBase]],
```

caikit/runtime/model_management/model_loader.py
```diff
@@ -96,14 +99,26 @@ def load_model(
...
     ) -> LoadedModel:
         try:
             log.info("<RUN89711114I>", "Loading model '%s'", model_id)
 
+            # Only pass finder/initializer if they have values
+            load_kwargs = {}
+            if finder:
```
gabe-l-hart, line 114: Should this be an `is None` check to allow empty strings? I think probably not, but I could _maybe_ imagine a case where someone uses an empty key. Probably not worth supporting

gabe-l-hart (changes requested):
One tiny change request on the type hint

caikit/runtime/model_management/model_loader.py
HonakerM: Good call! I also updated them to be kwargs

gabe-l-hart (approved):
Looks good!

PR #636, opened 2024-01-05
Liveness probe 635

### LINKED ISSUE #635: Lightweight Liveness Probe

The current implementation of `caikit_health_probe` implements a complete `readiness` probe which will not pass unless/until the server is running and able to service requests. This is correct behavior to determine if the `caikit.runtime` server(s) are ready to take traffic, but it is _not_ correct behavior to determine if they are still alive. When a server becomes overloaded with a backlog of requests, it is not able to handle more requests, but it _is_ still able to process the in-flight requests and will eventually become unblocked if new requests stop arriving.

A `caikit.runtime` server should be considered alive if it has not crashed and is able to make progress on outstanding requests. The exact semantics of how that should be measured are part of solving this issue.

gabe-l-hart:
Closes #635 

This PR expands the `caikit_health_probe` tool to allow for separate `liveness`/`readiness` semantics. The former `health_probe` is converted to `readiness_probe` and a new `liveness_probe` is added. To select a probe type, add the type (`liveness` or `readiness`) to the `python -m caikit_health_probe` command.

In addition to the new `liveness` probe, this PR adds the `runtime.grpc.probe_timeout` config argument which can be used to set a timeout for the grpc `readiness` probe. It defaults to `null` so the default behavior will not change during startup.

The change is non-breaking since the default behavior of `python -m caikit_health_probe` is equivalent to `python -m caikit_health_probe readiness`.

This change adds a new dependency on `psutil` to the base dependency set. The justification for this is the friendly license (BSD 3-clause) and lack of transitive dependencies. The open question remains whether this should be added _only_ for the `runtime-*` extras or be part of the core dependency set. Since  there is no common `runtime-base` extra set, and because it is a lightweight addition, I opted for the core instead of either creating a "hidden" `runtime-base` set or adding it to both `runtime-http` and `runtime-grpc` independently which would require keeping the versions in sync.

**If applicable**:

### DIFF
--- a/tests/runtime/test_caikit_health_probe.py
+++ b/tests/runtime/test_caikit_health_probe.py
@@ -25,6 +25,9 @@ from dataclasses import dataclass
 from enum import Enum
 from unittest import mock
 import os
+import shlex
+import subprocess
+import sys
 
 # Third Party
 import pytest
@@ -146,7 +149,7 @@ class ProbeTestConfig:
         ),
     ],
 )
-def test_health_probe(test_config: ProbeTestConfig):
+def test_readiness_probe(test_config: ProbeTestConfig):
     """Test all of the different ways that the servers could be running"""
     with alog.ContextLog(log.info, "---LOG CONFIG: %s---", test_config):
         # Get ports for both servers
@@ -192,11 +195,11 @@ def test_health_probe(test_config: ProbeTestConfig):
                 "merge",
             ):
                 # Health probe fails with no servers booted
-                assert not caikit_health_probe.health_probe()
+                assert not caikit_health_probe.readiness_probe()
                 # If booting the gRPC server, do so
                 with maybe_runtime_grpc_test_server(grpc_port):
                     # If only running gRPC, health probe should pass
-                    assert caikit_health_probe.health_probe() == (
+                    assert caikit_health_probe.readiness_probe() == (
                         test_config.should_become_healthy
                         and test_config.server_mode == ServerMode.GRPC
                     )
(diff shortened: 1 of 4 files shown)


tests/runtime/test_caikit_health_probe.py
```diff
@@ -208,6 +211,32 @@ def test_health_probe(test_config: ProbeTestConfig):
...
                         assert (
-                            caikit_health_probe.health_probe()
+                            caikit_health_probe.readiness_probe()
                             == test_config.should_become_healthy
                         )
+
+
+@pytest.mark.parametrize(
+    ["proc_identifier", "expected"],
+    [(None, True), ("caikit.runt", True), ("foobar", False)],
```
evaline-ju, line 221: A bit confused with the `caikit.runt` example here - with the `in` check here: https://github.com/caikit/caikit/pull/636/files#diff-e8e3913c46ffe6feeadbaa8ce888d10c282f7b2e38e75d656932755be682295cR104 would this then imply this is just checking for running other (`caikit` vs. `caikit.runtime`?) processes in this particular test case?

gabe-l-hart: heh, yeah, sorry, this one is gorpy. `caikit.runt` _should_ correctly satisfy `proc_identifier in arg`, and thus pass the liveness test with a custom override that _does_ match. This is all meant to support the case where the actual runtime process is actually running with some other executable (e.g. `wrapper_server` or something). The third test with `"foobar"` is there to make sure that if there _happens_ to be some other `caikit.runtime` running that is _not_ the server being probed, it doesn't get found by accident.

evaline-ju (approved):
LGTM!

PR #638, opened 2024-01-11
Add dataclass attribute to DataBases

HonakerM:
This PR adds the `@dataclass` attribute to various DataBases include `File` and the `PrimitiveSequences`. This enables better type hinting and debugging

**If applicable**:

### DIFF
--- a/caikit/interfaces/common/data_model/file.py
+++ b/caikit/interfaces/common/data_model/file.py
@@ -16,6 +16,7 @@ This file contains interfaces required to handle File data
 """
 
 # Standard
+from dataclasses import dataclass
 from typing import Optional
 
 # First Party
@@ -26,6 +27,7 @@ from caikit.core.data_model import PACKAGE_COMMON, DataObjectBase, dataobject
 
 
 @dataobject(PACKAGE_COMMON)
+@dataclass
 class File(DataObjectBase):
     data: Annotated[bytes, FieldNumber(1)]
     filename: Annotated[Optional[str], FieldNumber(2)]
--- a/caikit/interfaces/common/data_model/primitive_sequences.py
+++ b/caikit/interfaces/common/data_model/primitive_sequences.py
@@ -16,6 +16,7 @@ This file contains interfaces required for unions of lists
 """
 
 # Standard
+from dataclasses import dataclass
 from typing import List
 
 # First Party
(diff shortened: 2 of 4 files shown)


gabe-l-hart (commented):
Hm, this is an interesting change. Is the purpose here largely around IDE integration and type hinting? My concerns with doing this are:

1. It's redundant since `@dataobject` will do this automatically (more lines of code to maintain)
2. It may confuse readers who will assume that `@dataobject` does _not_ do this for you

HonakerM:
@gabe-l-hart you're correct that this is largely an IDE/type-hint thing. I'm not sure how to properly solve it because while `Dataobject` effectively does `@dataclass`, static type linters are unable to detect it.

The other idea I had was to add a `pyi` file but that felt a little to much

gabe-l-hart:
Hm, interesting. Are we thinking that this is because those IDEs/type checkers have special logic for `@dataclass` since it's a builtin? It seems odd to me that they would behave differently there unless they've special-cased `@dataclass`.

HonakerM:
@gabe-l-hart 
> Are we thinking that this is because those IDEs/type checkers have special logic for @dataclass since it's a builtin?
I believe so. I think part of the problem is that we use it like a dataclass but it's created dynamically which confuses static type checkers.

sboagibm:
@gabe-l-hart said:
> It's redundant since @dataobject will do this automatically (more lines of code to maintain)

It's a decorator.  For what it gives, not egregious, very high return on investment.

Using [@typing.dataclass_transform()](https://docs.python.org/3.11/whatsnew/3.11.html#pep-681-data-class-transforms) may be another option, that may feel less funky, though I haven't played with that myself.

> It may confuse readers who will assume that @dataobject does not do this for you

Explicit is more confusing than Implicit magic?  The dataclass decorator just makes explicit what is hidden behind dynamic construction.  dataclass is built into python now, widely used, and a very good thing.

> Are we thinking that this is because those IDEs/type checkers have special logic for @dataclass since it's a builtin? It seems odd to me that they would behave differently there unless they've special-cased @dataclass.

Most of what dataclass does is create an implicit __init__ function.  All the type inference does with them is infer the arguments to the __init__ function.  Both PyCharm and VSCode know about dataclasses, they are really core python now.  I've been using these dataclass decorators in WDU for quite a while now, and, as far as I can tell, they add no harm, and make it so using the datamodel objects make sense to the developer-user.

This developer experience is uber important for any open source or inner source offering.  I'll keep saying that until I'm blue in the face.

sboagibm:
What is the status with this?

gabe-l-hart:
Thanks for re-reminding me Scott. I'll look into this today

gabe-l-hart (approved):
I've attempted a more robust solution in #647, but it seems like there's more work to get this fully working with `pylance`/`pyright`. Let's merge this stopgap PR in the meantime and continue to look at full type correctness in `@dataobject` going forward.

PR #641, opened 2024-01-15
Fix Initialization bug in MultiModelInitializer

HonakerM:
This PR fixes a small bug which would break the MultiModelInitializer when loading from config. This also updates the test cases to hopefully avoid this issue in the future

**If applicable**:


gabe-l-hart (approved):
Yikes! Thanks for catching this

PR #643, opened 2024-01-19
Fix union validation

alex-jw-brooks:
Currently there's a bug in the way that we validate module signatures against task definitions when both contain union typed args. The main inconsistency is as follows:
- Say that a module has param named `inputs`, which is of type `Union[int, str, bytes]`
- Now say that this module implements task `Foo`. If `Foo` defines:
     - `inputs` as a an `int`, `str`, OR, `bytes`, everything is happy
     - `inputs` as a `Union` containing some subset of these types, everything explodes with a TypeError

This behavior should be allowed, because in some cases, the module may have args that aren't directly compatible with the base data model, e.g., types supported by the Image data model backend, namely PIL image. We should be able to pass these in at the library level, but ensure that any of the valid serializable arg types can be used in the task definition for support in runtime.

This PR fixes validation of union validation by ensuring that union typed parameters are further inspected, and are allowed if they are subsets of the corresponding union typed signature def.

Currently Caikit Computer Vision has `bytes` as the input type for its tasks, but we need this change to be able to support a wider variety of inputs that are supported by the image backend, e.g., paths to images, bytes, decoded images, etc.

**If applicable**:

### DIFF
--- a/caikit/core/task.py
+++ b/caikit/core/task.py
@@ -169,9 +169,17 @@ class TaskBase:
         ).items():
             signature_type = signature.parameters[parameter_name]
             if parameter_type != signature_type:
-                if typing.get_origin(
-                    signature_type
-                ) == typing.Union and parameter_type in typing.get_args(signature_type):
+                if typing.get_origin(signature_type) == typing.Union and (
+                    # Either our parameter type is not a union & is part of the union signature
+                    parameter_type in typing.get_args(signature_type)
+                    # Or our parameter type is a union that's a subset of the union signature
+                    or (
+                        typing.get_origin(parameter_type) == typing.Union
+                        and set(typing.get_args(parameter_type)).issubset(
+                            set(typing.get_args(signature_type))
+                        )
+                    )
+                ):
                     continue
                 if input_streaming and cls._is_iterable_type(parameter_type):
                     streaming_type = typing.get_args(parameter_type)[0]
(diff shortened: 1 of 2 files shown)


caikit/core/task.py
```diff
@@ -169,9 +169,17 @@ def validate_run_signature(
...
             if parameter_type != signature_type:
-                if typing.get_origin(
-                    signature_type
-                ) == typing.Union and parameter_type in typing.get_args(signature_type):
+                if typing.get_origin(signature_type) == typing.Union and (
+                    # Either our parameter type is not a union & is part of the union signature
+                    parameter_type in typing.get_args(signature_type)
+                    # Or our parameter type is a union that's a subset of the union signature
+                    or (
+                        typing.get_origin(parameter_type) == typing.Union
```
gabe-l-hart, line 177: Do we need to handle `Annotated` here at all? I guess we probably need to unwrap that before this loop? Probably not part of this PR though.

gabe-l-hart: NOTE for future: The `validate_run_signature` is only called at decoration time, so while this additional logic may _slightly_ slow down `import` time, it will have no impact at runtime.

gabe-l-hart (approved):
Thank you for this Alex! Very nice improvement to unblock vision and other multimodal cases

caikit/core/task.py
alex-jw-brooks: Good question - probably, although currently we don't us annotated explicitly in our task definitions and module definitions, right? When are we expected to have annotations in this validation if the annotations are mostly defining things on data model objects?

PR #647, opened 2024-01-25
Use dataclass transform for dataobject

### LINKED ISSUE #608: 🐛 Cannot use type annotations for custom `__init__` args in `@dataobject` classes
## Describe the bug

When declaring a `dataobject` with a custom `__init__`, the decorator [looks for the presence of `__annotations__`](https://github.com/caikit/caikit/blob/main/caikit/core/data_model/dataobject.py#L474) on the `__init__` function to determine whether it was created by `@dataclass` or not. This is not a good enough test since the `__annotations__` key is added to the `__init__` function whenever type annotations are added to the arguments.

## Platform

- Interpreter version: `3.8.16`
- Library version: `0.25.6` (`main`)

## Sample Code

```py
from caikit.core.data_model import DataObjectBase, dataobject

@dataobject
class Doubler(DataObjectBase):
    val: int
    def __init__(self, val: int): # <-- Without the type annotation this works as expected
        self.val = val * 2

x = Doubler(4)
print(x.val)
assert x.val == 8
```

## Expected behavior

The above custom `__init__` will be used and `x.val` will be set to `8`

## Observed behavior

The above custom `__init__` is ignored and replaced with the auto-generated `__init__`, so `x.val` is set to `4`

gabe-l-hart:

This PR solves two related problems:

1. It uses [@dataclass_transform](https://docs.python.org/3/library/typing.html#typing.dataclass_transform) on the definition of the `@dataobject` decorator to tell static type checkers that it behaves like `@dataclass`.
2. It resolves #608 allowing custom `__init__` functions to use type hints

This is an attempt at a global solution to the problems that prompted #638. Without being able to replicate the exact VS Code failures, I'm not sure that `@dataclass_transform` will fix the issue.

**If applicable**:


gabe-l-hart:
@sboagibm I'm going to close this one as I don't think it actually solved your type hinting issue well. We can always revisit

PR #651, opened 2024-01-31
Single module server

gabe-l-hart:

This PR enables the usage of `__main__` as the `runtime.library` so that an end-to-end example of `caikit.runtime` can be contained in a single python script.

While not strictly necessary for standard operation, this PR is worth merging because (a) it helps with concise examples and (b) users _may_ want to write this kind of single-module server in the future.

**If applicable**:

### DIFF
--- a/caikit/core/toolkit/name_tools.py
+++ b/caikit/core/toolkit/name_tools.py
@@ -19,4 +19,4 @@ and other Protobuf names
 
 def snake_to_upper_camel(string: str) -> str:
     """Simple snake -> upper camel conversion for descriptors"""
-    return "".join([part[0].upper() + part[1:] for part in string.split("_")])
+    return "".join([part[0].upper() + part[1:] for part in string.split("_") if part])
--- a/caikit/runtime/grpc_server.py
+++ b/caikit/runtime/grpc_server.py
@@ -120,14 +120,14 @@ class RuntimeGRPCServer(RuntimeServerBase):
             )
 
             # Add training management servicer to the gRPC server
-            training_management_service: ServicePackage = (
+            self.training_management_service: ServicePackage = (
                 ServicePackageFactory.get_service_package(
                     ServicePackageFactory.ServiceType.TRAINING_MANAGEMENT,
                 )
             )
-            service_names.append(training_management_service.descriptor.full_name)
+            service_names.append(self.training_management_service.descriptor.full_name)
 
-            training_management_service.registration_function(
+            self.training_management_service.registration_function(
                 TrainingManagementServicerImpl(), self.server
             )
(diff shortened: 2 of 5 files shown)


caikit/core/toolkit/name_tools.py
```diff
@@ -19,4 +19,4 @@
 
 def snake_to_upper_camel(string: str) -> str:
     """Simple snake -> upper camel conversion for descriptors"""
-    return "".join([part[0].upper() + part[1:] for part in string.split("_")])
+    return "".join([part[0].upper() + part[1:] for part in string.split("_") if part])
```
gabe-l-hart, line 22: When using `__main__` as the library, the `string.split("_")` results in `('', '', 'main', '', '')` and therefore `part[0]` raises an `IndexError` for the empty parts.

caikit/runtime/grpc_server.py
```diff
@@ -120,14 +120,14 @@ def __init__(
             )
 
             # Add training management servicer to the gRPC server
-            training_management_service: ServicePackage = (
+            self.training_management_service: ServicePackage = (
```
gabe-l-hart, line 123: I added this to enable referencing the `training_management_service` off of an in-memory server handle. This is very likely not a production usage, but it mirrors how `self.training_service` and `self.inference_service` are stored.

caikit/runtime/servicers/global_train_servicer.py
```diff
@@ -73,7 +73,7 @@ def __init__(self, training_service: ServicePackage):
             lib_version = "unknown"
 
         log.info(
-            "<RUN76884779I>",
+            "<RUN76884778I>",
```
gabe-l-hart, line 76: This was a copy-paste duplicate!

caikit/runtime/utils/import_util.py
```diff
@@ -161,6 +162,8 @@ def get_dynamic_module(module_name: str, module_dir: str = None) -> ModuleType:
     Returns:
         (module): Handle to the module after dynamic import
     """
+    if module := sys.modules.get(module_name):
```
gabe-l-hart, line 165: This is needed to short-circuit for the "special" module `__main__` that is already (always) in `sys.modules`, but will fail on `importlib.util.find_spec` since there is not spec to be found for it.

gabe-l-hart: It also provides a nice quick-escape for modules that have already been imported some other way

alex-jw-brooks: Nice!

alex-jw-brooks (approved):
LGTM

PR #652, opened 2024-01-31
added id to error dict, and removed extra message in details

swith004:
This issue should address inconsistent messaging for unary and streaming endpoints (with and without guardrails), reported in [issue 170 ](https://github.ibm.com/ai-foundation/watsonx_platform/issues/170#issuecomment-66594673).

**If applicable**:
-  this PR has been tested for compatibility with guardrails


caikit/runtime/servicers/global_predict_servicer.py
```diff
@@ -329,7 +329,7 @@ def _handle_predict_exceptions(self, model_id: str, request_name: str):
             ).inc()
             raise CaikitRuntimeException(
                 StatusCode.INVALID_ARGUMENT,
-                f"Exception raised during inference. This may be a problem with your input: {e}",
+                f"{e}",
```
prashantgupta24, line 332: don't really need the f-string?

caikit/runtime/servicers/global_predict_servicer.py
```diff
@@ -359,7 +359,7 @@ def _handle_predict_exceptions(self, model_id: str, request_name: str):
                 model_id=model_id,
             ).inc()
             raise CaikitRuntimeException(
-                StatusCode.INTERNAL, "Unhandled exception during prediction"
+                StatusCode.INTERNAL, f"{e}",
```
prashantgupta24, line 362: don't really need the f-string?

caikit/runtime/servicers/global_predict_servicer.py
swith004: It seems we do, as removing it causes "Internal server error" to display instead of the exception

caikit/runtime/servicers/global_predict_servicer.py
swith004: It seems we do, as removing it causes "Internal server error" instead of the exception

gabe-l-hart (approved):
Looks good, thanks Shonda!

prashantgupta24 (approved):

swith004:
@prashantgupta24 - I've updated the description of this PR

PR #654, opened 2024-01-31
Model info loaded 653

### LINKED ISSUE #653: Expose loading status in `/info/models`

The [current code](https://github.com/caikit/caikit/blob/main/caikit/runtime/servicers/info_servicer.py#L75) for the `InfoServicer` will report all models that are currently in `model_manager.loaded_models` without checking the status of whether or not those models have completed loading.

In some cases, a user will need to know that a model has been fully loaded in addition to knowing that it has _started_ loading. The request in this PR is to add a new field to the [ModelInfo](https://github.com/caikit/caikit/blob/main/caikit/interfaces/runtime/data_model/info.py#L57) object to represent the load state.

* Add a new `loaded: bool` field to `ModelInfo`
* Add a `loaded() -> bool` function to `LoadedModel`
* In `InfoServicer`, check the `LoadedModel.loaded()` value to populate `loaded`

* Just try an inference and see if it fails. This could work, but could also cause unnecessary work since the inference would be discarded.

## Additional context

This is particularly relevant when using `caikit-nlp` + TGIS since some inference requests may go directly to TGIS and skip the `caikit.runtime` wrapper layer. If the request to TGIS occurs after `caikit.runtime` has started loading a given prompt model, but before it has finished copying that model to TGIS, the user can experience an error where TGIS attempts to load the model and fails because the file is partially copied. With this fix in place, a user could poll `/info/models` until the load completes before hitting TGIS directly.

gabe-l-hart:

Closes #653 

This PR adds a new `loaded` field to the output of each model info block behind the `InfoServicer`'s model info endpoint/rpc

**If applicable**:

### DIFF
--- a/caikit/interfaces/runtime/data_model/info.py
+++ b/caikit/interfaces/runtime/data_model/info.py
@@ -62,6 +62,7 @@ class ModelInfo(DataObjectBase):
     name: Annotated[str, FieldNumber(2)]
     size: Annotated[int, FieldNumber(3)]
     metadata: Annotated[Dict[str, str], FieldNumber(4)]
+    loaded: Annotated[bool, FieldNumber(7)]
 
     # Module Information
     module_id: Annotated[str, FieldNumber(5)]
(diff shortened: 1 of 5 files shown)


evaline-ju (approved):
LGTM!

Issue #646, opened 2024-01-22
Flakey tests `Address already in use`


gabe-l-hart:
## Describe the bug

We're seeing pretty regular CI failures (e.g. [this one](https://github.com/caikit/caikit/actions/runs/7607018941/job/20713665701#step:5:426)) with `Address already in use`. Those need to go away!

## Expected behavior

Our tests should use port randomization to avoid port collisions between tests. They already do this, but it's not a perfect science since the check for a free port may occur when the port is free and it may be taken by another thread before the thread that asked attempts to bind it.

## Observed behavior

[What you see happening (error messages, stack traces, etc...)](https://github.com/caikit/caikit/actions/runs/7607018941/job/20713665701#step:5:426)

PR #655, opened 2024-01-31
FlakeyPortsInTests

gabe-l-hart:

Use tls_test_tools.open_port. This randomizes the port selection so it's far less likely to hit conflicts

Addresses https://github.com/caikit/caikit/issues/646

**If applicable**:


gabe-l-hart:
This seems to not actually make the problem better and can make for _more_ hanging!

PR #660, opened 2024-02-06
FlakeyTestsRandomPorts: Use internal open-port function

### LINKED ISSUE #646: Flakey tests `Address already in use`

gabe-l-hart:

Closes #646 

Second attempt to solve the flakey port collision tests!

**If applicable**:

### DIFF
--- a/tests/runtime/conftest.py
+++ b/tests/runtime/conftest.py
@@ -3,7 +3,7 @@ This sets up global test configs when pytest starts
 """
 
 # Standard
-from contextlib import contextmanager
+from contextlib import closing, contextmanager
 from functools import partial
 from typing import Dict, List, Optional, Type, Union
 import os
@@ -52,7 +52,11 @@ log = alog.use_channel("TEST-CONFTEST")
 
 def get_open_port():
     """Non-fixture function to get an open port"""
-    return _open_port()
+    with closing(socket.socket(socket.AF_INET, socket.SOCK_STREAM)) as s:
+        s.bind(("", 0))
+        s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+        port = s.getsockname()[1]
+        return port
 
 
 @pytest.fixture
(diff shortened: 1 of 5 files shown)


tests/runtime/conftest.py
```diff
@@ -50,6 +50,11 @@
 log = alog.use_channel("TEST-CONFTEST")
 
 
+def get_open_port():
+    """Non-fixture function to get an open port"""
+    return _open_port()
+
+
 @pytest.fixture
 def open_port():
```
z103cb, line 59: @gabe-l-hart, there's a slightly better approach to getting a next available port.  
In the caikit-nlp-client library that @dtrifiro and I have developed we execute this function:

```python
def get_random_port():
    with closing(socket.socket(socket.AF_INET, socket.SOCK_STREAM)) as s:
        s.bind(("", 0))
        s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
        port = s.getsockname()[1]
        return port
```
Basically it returns a random open port from the whole range of ports on the host. We have not seen any problems with it running into any collisions and I have seen this technique work very well in other languages. Here's the [link to the original](https://github.com/z103cb/caikit-nlp-client/blob/c797414e68ee3723636ee84375eda01d8c1ee318/tests/fixtures/utils.py#L13C1-L18C20) to the code. We used this [SO Post](https://stackoverflow.com/questions/1365265/on-localhost-how-do-i-pick-a-free-port-number) as the basis of the implementation.

PR #662, opened 2024-02-09
Proactive load toggle 661

### LINKED ISSUE #661: Runtime lazy load only mode

When managing large numbers of models, it can be very expensive to perform the proactive loading in [sync_local_models](https://github.com/caikit/caikit/blob/main/caikit/runtime/model_management/model_manager.py#L263). This can result in slow boots and starved server thread pools.

The request here is to add a feature to enable the periodic sync to only handle deletion and not perform proactive loading. This would defer all loading to lazy-loading on inference while maintaining the ability to unload models that have been removed from `local_models_dir`.

* Add a new config argument `runtime.proactive_load_local_models`
* In [_local_models_dir_sync](https://github.com/caikit/caikit/blob/main/caikit/runtime/model_management/model_manager.py#L490), only perform the loading if `proactive_load_local_models` is `true`

This issue is related to [an issue in `caikit-tgis-runtime`](https://github.com/caikit/caikit-tgis-backend/issues/51) that is coming up due to massive repetition of copying prompt artifacts from `local_models_dir` to TGIS' prompt cache. We've considered much more sophisticated solutions and may implement them in the future, but this will act as a simple mitigation to avoid ongoing problems.

## Additional context

The main downside of this proposal is that the `info` service loses the ability to guarantee eventual consistency between replicas of a given server.

gabe-l-hart:

Closes #661 

This PR adds the `runtime.load_new_local_models` config option which can disable loading new models during synchronization with `local_models_dir`.

When loading new local models is disabled, the model info endpoints are not guaranteed to be consistent across replicas of `caikit.runtime` running behind a loadbalancer (e.g. k8s `Deployment` / `Service`) since loading only happens on inference calls which are distributed across the replicas.

**If applicable**:
- [x] this PR contains documentation (comment in `config.yml`)

### DIFF
--- a/caikit/config/config.yml
+++ b/caikit/config/config.yml
@@ -97,6 +97,11 @@ runtime:
     # the server
     wait_for_initial_model_loads: true
 
+    # If true, on each local_models_dir sync (include the initial one), any new
+    # models will start loading. If false, new models will be ignored and will
+    # only lazy load on inference.
+    load_new_local_models: true
+
     # If enabled, the models in local_models_dir will be periodically sync'ed
     # with the in-memory models. New models that are not in-memory that are
     # found in local_models_dir will be loaded and existing models that were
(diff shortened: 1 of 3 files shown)


alex-jw-brooks (approved):
Looks good! LGTM

PR #675, opened 2024-03-04
Add input token count to embedding, reranker, sentence similarity

mynhardtburger:
A requirement for https://github.com/caikit/caikit-nlp/pull/334
Adds `input_token_count` to the results data models. This allows for metrics / billing on the size of each task to be collected.

**If applicable**:

### DIFF
--- a/caikit/interfaces/nlp/data_model/embedding_vectors.py
+++ b/caikit/interfaces/nlp/data_model/embedding_vectors.py
@@ -15,6 +15,7 @@
 
 # Standard
 from dataclasses import dataclass
+from typing import Optional
 
 # First Party
 from py_to_proto.dataclass_to_proto import Annotated, FieldNumber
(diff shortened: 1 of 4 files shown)


mynhardtburger:
cc: @markstur

markstur (approved):
LGTM

Curious if @caikit/caikit-maintainers would want Optional on the new fields.

evaline-ju (commented):
> Curious if @caikit/caikit-maintainers would want Optional on the new fields.

I think the general consensus has been that while this extra information is helpful for particular use cases, it is not fundamental to understanding/parsing an embeddings result - so `Optional` may be better here.

mynhardtburger:
I've made the `input_token_count` fields `Optional[int]`.

Note that I don't think that the `Optional` label has any impact on the actual GRPC messages, other than allowing to test for the [presence of the field](https://protobuf.dev/programming-guides/field_presence/) and more clearly communicating intent to a reader.

evaline-ju:
> Note that I don't think that the `Optional` label has any impact on the actual GRPC messages, other than allowing to test for the [presence of the field](https://protobuf.dev/programming-guides/field_presence/) and more clearly communicating intent to a reader.

Yes I don't believe the actual GRPC messages will be impacted either

evaline-ju (approved):
LGTM!

PR #681, opened 2024-03-14
Multi connection remote model finder 680

### LINKED ISSUE #680: Support multiple remote servers in `RemoteModelFinder`

The `RemoteModelFinder` allows a single `caikit.runtime` to proxy requests to a single remote `caikit.runtime` server for the collection of models loaded there. This feature request is to extend that functionality to support multiple remote servers behind a single instance of `RemoteModelFinder`.

* Allow the `connection.hostname` key in the config to include a template that can be filled in with the `model_name`/`model_id`
    * This allows the set of connections to change dynamically without rolling over the server to reload config
* Add support for a `remote_connections` config to pre-populate connections to multiple known servers.

The desired feature can be accomplished with the current implementation by having one configured `RemoteModelFinder` for each remote server, however this set of servers would be immutable after boot, so the server acting as the proxy would need to be rolled over to update the mapping.

gabe-l-hart:

Closes #680 

This PR adds support to the `caikit.runtime.client.RemoteModelFinder` so that it can find models across multiple remote `caikit.runtime` servers. There are two new ways to enable multi-connection finding:

1. The `config.connection.hostname` string can contain a single `{}` template that will be replaced with `model_id`.
2. The `config.remote_connections` entry can be a list of connection blobs, each one with the same set of arguments as `config.connection` (`hostname`, `port`, `tls`, `options`).

Here's an example of a full configuration with `RemoteModelFinder`, `LocalFinder`, and [TGISAutoFinder](https://github.com/caikit/caikit-nlp/blob/main/caikit_nlp/model_management/tgis_auto_finder.py):

```yaml
model_management:
  initializers:
    default:
      type: LOCAL
      config:
        backend_priority:
          - type: LOCAL
          - type: TGIS
            config:
              connection:
                hostname: tgis.server:8080
  finders:
    default:
      type: LOCAL
      config:
        load_path: null
    remote:
      type: REMOTE
      config:
        connection:
          hostname: foo.bar.{}
          port: 1234
        remote_connection:
          - hostname: bar.baz
            port: 54321
          - hostname: biz.buz
            port: 12121
    tgis:
      type: TGIS-AUTO
      config:
        local_initializer_name: default
        tgis_backend_priority: 1
```

PR still needs more unit tests!

**If applicable**:

### DIFF
--- a/caikit/runtime/client/remote_model_finder.py
+++ b/caikit/runtime/client/remote_model_finder.py
@@ -95,14 +95,14 @@ class RemoteModelFinder(ModelFinderBase):
 
         # If a remote_models key is found, it's a mapping from model name to
         # connection info
-        for remote_conn in config.get("remote_connections", {}):
+        for remote_conn in config.get("remote_connections", []):
             conn = ConnectionInfo(**remote_conn)
-            self._connections[conn.hostname] = conn
+            self._connections[f"{conn.hostname}:{conn.port}"] = conn
 
         # If a single "global" connection given, initialize with model_name None
         if config.connection:
             default_conn = ConnectionInfo(**config.connection)
-            if default_conn.hostname not in self._connections:
+            if f"{default_conn.hostname}:{default_conn.port}" not in self._connections:
                 self._connection_template = default_conn
                 self._connections[default_conn.hostname] = default_conn
 
@@ -266,27 +266,29 @@ class RemoteModelFinder(ModelFinderBase):
                     model_info_proto = info_service_rpc(
                         ModelInfoRequest().to_proto(), timeout=conn.timeout
                     )
+
+                    model_info_response = ModelInfoResponse.from_proto(model_info_proto)
+
+                    # Parse response into dictionary of name->conn
+                    for model_info in model_info_response.models:
+                        model_name = model_info.name
+                        module_id = model_info.module_id
+
+                        log.debug(
+                            "Discovered model %s with module_id %s from remote runtime %s",
+                            model_name,
+                            module_id,
+                            target,
+                        )
+                        supported_modules[model_name] = ModuleConnectionInfo(
+                            conn, module_id
+                        )
                 except grpc.RpcError as exc:
                     log.warning(
                         "Unable to discover modules from remote: %s. Error: %s",
-                        conn.hostname,
+                        target,
                         str(exc),
                     )
-                    return {}
-
-            model_info_response = ModelInfoResponse.from_proto(model_info_proto)
-
-            # Parse response into dictionary of name->conn
-            for model_info in model_info_response.models:
-                model_name = model_info.name
-                module_id = model_info.module_id
-
-                log.debug(
-                    "Discovered model %s with module_id %s from remote runtime",
-                    model_name,
-                    module_id,
-                )
-                supported_modules[model_name] = ModuleConnectionInfo(conn, module_id)
 
         return supported_modules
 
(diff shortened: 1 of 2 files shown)


caikit/runtime/client/remote_model_finder.py
```diff
@@ -195,110 +238,139 @@ def _safe_discover(self) -> Dict[str, str]:
...
+
+    def _get_conn_candidates(self, model_name: Optional[str]) -> List[ConnectionInfo]:
+        """Common utility to get all connections to try"""
+        candidate_conns = list(self._connections.values())
+        if (
+            model_name is not None
+            and self._connection_template is not None
+            and (model_conn := self._render_conn_template(model_name))
+        ):
+            candidate_conns.append(model_conn)
```
gkumbhat, line 372: this will put the connection with template at the back of the list so when trying to make connection request to check for model, for example [here](https://github.com/caikit/caikit/pull/681/files#diff-b2d06d99d6692ffc8747bc7bc99c53f4ae9122c17e27971babf99052c82f9141R307), we would end up making a call to incorrect server first (assuming it satisfies the template) right?  Should we prepend instead?

caikit/runtime/client/remote_model_finder.py
```diff
@@ -195,110 +238,139 @@ def _safe_discover(self) -> Dict[str, str]:
...
+                        model_name = model_info.name
+                        module_id = model_info.module_id
+
+                        log.debug(
+                            "Discovered model %s with module_id %s from remote runtime %s",
+                            model_name,
+                            module_id,
+                            target,
+                        )
+                        supported_modules[model_name] = ModuleConnectionInfo(
```
gkumbhat, line 283: since we are saving / caching the connection here, would this create problem for the cases where the pods gets refreshed and so the old connection object is now invalid?

gkumbhat (commented):
Thanks Gabe. Left a few questions, mainly concerned if connection caching and discoverability would cause problem in case of pod rollover for servers running the models.

caikit/runtime/client/remote_model_finder.py
gabe-l-hart: Hmm, good call. Probably best to prepend. This does touch on one open question I have: how to handle multiple servers supporting the same model? Currently, it's [last one wins](https://github.com/caikit/caikit/pull/681/files#diff-b2d06d99d6692ffc8747bc7bc99c53f4ae9122c17e27971babf99052c82f9141R341), so moving to prepending would not actually save any effort since all servers would be checked anyway. We could certainly make this more efficient, though! That said, this is a load-time operation, so once the connection is established, it should have no effect on the inference latency.

caikit/runtime/client/remote_model_finder.py
gabe-l-hart: Yeah, this is definitely a key observation: There are several assumptions about the service behind the connection:

1. It is persistent. This is most easily accomplished by having the connection connect to a `Service` instead of `Pod`, but that then leads to (2)
2. It is homogeneous such that the model info endpoint will respond with the same set of loaded models, regardless of which replica is hit behind a service router.

caikit/runtime/client/remote_model_finder.py
gkumbhat: true. But:
1. Wouldn't it be unintuitive for the last one to win, because one would expect first one to get discovered would win and have first come first serve 
2. Making call to incorrect backend would also create noise, so from the perspective of debugging a routing / model loading issue, we would see more failures from servers that are not responsible.

gabe-l-hart: Updated so that the first-in wins. The catch is the the connection calls are not per-model since models are `1:N` with runtimes, so we can't really limit the discovery calls by model.

gkumbhat: true. 

Thanks for updating

gkumbhat (approved):
LGTM. Thanks for putting this together and iterations.

PR #686, opened 2024-03-20
Consolidated dump 685

### LINKED ISSUE #685: Consolidate dumped protobufs by package

When running `python -m caikit.runtime.dump_services`, the output for the `grpc` services produces a single `.proto` file for every individual object (`Service`, `Message`, or `Enum`). This quickly becomes unmanageable for clients and is not best-practice for `protobuf`.

The request here is to (optionally for compatibility) aggregate all objects by their package name and create a single `.proto` file for each `package`. This follows standard `protobuf` conventions and will make the dumped services much more usable.

Add a new flag to `dum_services` to dump the protobufs concolidated by package.

It would be nice to make this feature on by default, but that would be a breaking change, so for now it should be opt-in.

gabe-l-hart:

Closes #685 

This PR adds two new options to `caikit.runtime.dump_services`:

* `-c | --clean`: Clean out existing content in output dir
* `-p | --consolidate-packages`: Consolidate protobufs by package

The second is the primary point of this PR. It _drastically_ reduces the clutter in the dumped services directory and aligns more closely with standard protobuf practices to consolidate multiple objects into individual `.proto` files that are scoped by the `package`.

**If applicable**:

### DIFF
--- a/caikit/runtime/dump_services.py
+++ b/caikit/runtime/dump_services.py
@@ -13,51 +13,66 @@
 # limitations under the License.
 
 # Standard
+from typing import Dict, List, Optional, Union
 import argparse
 import json
 import os
+import shutil
 import sys
 
+# Third Party
+from google.protobuf import descriptor as _descriptor
+from google.protobuf import descriptor_pb2, descriptor_pool
+
 # First Party
+from py_to_proto import descriptor_to_file
+from py_to_proto.utils import safe_add_fd_to_pool
 import alog
 
 # Local
+from ..config.config import get_config
 from ..core.data_model import render_dataobject_protos
-from .service_factory import ServicePackageFactory
-from caikit.config.config import get_config
+from ..core.data_model.dataobject import _AUTO_GEN_PROTO_CLASSES
+from ..core.exceptions import error_handler
+from .service_factory import ServicePackage, ServicePackageFactory
 import caikit
 
 log = alog.use_channel("RUNTIME-DUMP-SVC")
+error = error_handler.get(log)
+
+## Public ######################################################################
 
 
-def dump_grpc_services(output_dir: str, write_modules_file):
-    """Utility for rendering the all generated interfaces to proto files"""
-    inf_enabled = get_config().runtime.service_generation.enable_inference
-    train_enabled = get_config().runtime.service_generation.enable_training
+def dump_grpc_services(
+    output_dir: str,
+    write_modules_file: bool,
+    consolidate: bool = False,
+):
+    """Utility for rendering the all generated interfaces to proto files
 
-    if inf_enabled:
-        inf_svc = ServicePackageFactory.get_service_package(
-            ServicePackageFactory.ServiceType.INFERENCE,
-            write_modules_file=write_modules_file,
+    Args:
+        output_dir (str): The directory where the generated services should be
+            placed
+        write_modules_file (bool): Whether or not to write out the compatibility
+            file for supported modules
+        consolidate (bool): Whether or not to consolidate the generated protos
+            by package
+    """
+    service_packages = _get_grpc_service_packages()
+    if not consolidate:
+        log.info(
+            "Dumping raw service and data model protos without package consolidation"
         )
-    if train_enabled:
-        train_svc = ServicePackageFactory.get_service_package(
-            ServicePackageFactory.ServiceType.TRAINING,
-        )
-        train_mgt_svc = ServicePackageFactory.get_service_package(
-            ServicePackageFactory.ServiceType.TRAINING_MANAGEMENT,
-        )
-    info_svc = ServicePackageFactory.get_service_package(
-        ServicePackageFactory.ServiceType.INFO,
-    )
-
-    render_dataobject_protos(output_dir)
-    if inf_enabled:
-        inf_svc.service.write_proto_file(output_dir)
-    if train_enabled:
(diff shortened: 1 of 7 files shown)


caikit/runtime/dump_services.py
```diff
@@ -13,51 +13,69 @@
...
+from py_to_proto import descriptor_to_file
+from py_to_proto.utils import safe_add_fd_to_pool
 import alog
 
 # Local
+from ..config.config import get_config
 from ..core.data_model import render_dataobject_protos
-from .service_factory import ServicePackageFactory
-from caikit.config.config import get_config
+from ..core.data_model.dataobject import _AUTO_GEN_PROTO_CLASSES
```
joerunde, line 35: big picture q: Where should all this .proto writing logic live?

It looks like we're trying to consolidate it here, so maybe we should kill `data_model.render_dataobject_protos()` and instead have a public `data_model.get_generated_proto_classes()` for this to use.

Ditto on killing `service_package.write_proto_file()` too?

gabe-l-hart: Yeah, I was fairly torn about this. The reason I _didn't_ go down this rabbit hole is that fundamentally `data_model` and "dump services" are serving two different purposes that overlap:

* `data_model`: Python structs that can be serialized to protobuf and managed just like any other python objects (NB: dynamic class declaration should be just fine!)
* "dump services": Client-usable protobuf interfaces

The difference comes around how the dynamic nature of the `data_model` classes maps down to `FileDescriptor`s. A `FileDescriptor` is meant to be static (you can easily mutate a `FileDescriptorProto`, but not a `FileDescriptor`), so having the `data_model` classes all live in package-consolidated `FileDescriptor`s would be really challenging since you'd need to know them all apriori.

Because of this gorpiness, I decided to do all of the logic to munch all the existing `Descriptor`/`EnumDescriptor`/`ServiceDescriptor` objects into a consolidated `FileDescriptorProto` -> `FileDescriptor` over here in the `dump_services` module for isolation. This should _not_ be how these protos are created in "normal" operations because it would badly break the extensibility of the data model.

So, this finally comes to the question of whether we kill off `data_model.get_generated_proto_classes` and `service_package.write_proto_file`. I think the former should be kept as it's not strictly about file rendering. I could go either way on the later, but would probably lean towards removing it (although that would be an API breaking change).

joerunde: > This should not be how these protos are created in "normal" operations because it would badly break the extensibility of the data model

Is there a use case for generating .proto files for the data model outside of generating a client for the runtime? I'm just wondering if it's worth keeping the data model's current proto rendering around at all.

> So, this finally comes to the question of whether we kill off `data_model.get_generated_proto_classes` and `service_package.write_proto_file.` I think the former should be kept as it's not strictly about file rendering

I'm confused, I was proposing creating a `data_model.get_generated_proto_classes` method to use instead of accessing a private variable from here. I don't think it exists yet.

> ... although that would be an API breaking change

how far has the ship sailed on caikit ever having a stable api? 😉 
I'd lean towards deleting both of the existing rendering methods as well

tests/runtime/test_dump_services.py
```diff
@@ -58,6 +58,33 @@ def test_dump_grpc_services_dir_does_not_exist():
...
+        assert all(
+            fname in dumped_files
+            for fname in {
+                "caikit_runtime_SampleLib.proto",
+                "caikit_runtime_info.proto",
+                "caikit_runtime_training.proto",
+                "caikit_data_model_common.proto",
+                "caikit_data_model_common_runtime.proto",
+                "caikit_data_model_runtime.proto",
+                "caikit_data_model_sample_lib.proto",
```
joerunde, line 83: Should we do some checks to see that at least one of these files contains all of the expected data models?

caikit/runtime/dump_services.py
gabe-l-hart: >I'm confused, I was proposing creating a data_model.get_generated_proto_classes method to use instead of accessing a private variable from here. I don't think it exists yet.

Sorry, bad reading comprehension. You're right that `data_model.render_dataobject_protos()` is basically the same thing as what I have here. I thought about doing all of this as an extra arg in that function, but opted against it because of all the temporary pool muckery.

joerunde: I think the simpler refactor I'm trying to suggest here is just
```
from ..core.data_model.dataobject import _AUTO_GEN_PROTO_CLASSES
```
becomes
```
from  ..core.data_model.dataobject import get_generated_proto_classes
```

But I suppose you could also just rename `_AUTO_GEN_PROTO_CLASSES` -> `AUTO_GEN_PROTO_CLASSES` if we want to access it outside the data model

joerunde: anyway- yeah my care level on this is pretty low. I like the output and consolidating the proto rendering into this file where possible is good. Let's get this merged?

gabe-l-hart: Yeah, I agree that importing the private singleton is not great. Will try to fix shortly

gabe-l-hart: Done

tests/runtime/test_dump_services.py
gabe-l-hart: Done

joerunde (approved):

PR #689, opened 2024-03-27
LimitHttpConcurrency

gabe-l-hart:

This PR adds new logic to set the `limit_concurrency` config value for the HTTP server. The default behavior is to set it to 2x the server thread pool size. This is indicated by using a sentinel value of `-1` by default (any negative number). If set to `0`, no concurrency limiting will be applied. If set to a positive number, that number will be used directly.

This _could_ be considered a breaking change since it changes the default behavior to enable concurrency limiting

**If applicable**:


evaline-ju (approved):
LGTM!

PR #690, opened 2024-03-28
TrainEntrypoint: Add train main entrypoint script

gabe-l-hart:

This PR adds a new module `caikit.runtime.train` which will execute a single training job. It is useful for running module train operations inside of a job orchestration framework.

```sh
LOG_LEVEL=debug3 \
LOG_FORMATTER=pretty \
PYTHONPATH=$PWD/tests/fixtures \
python -m caikit.runtime.train \
    -m sample_lib.modules.sample_task.sample_implementation.SampleModule \
    -n "my-model" \
    -k '{"training_data": {"jsondata": {"data": [{"number": 1, "label": "foo"}, {"number": 2, "label": "bar"}]}}}'
```

**If applicable**:

### DIFF
--- a/caikit/runtime/train.py
+++ b/caikit/runtime/train.py
@@ -1,0 +1,197 @@
+# Copyright The Caikit Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+This module is a central entrypoint for running a single synchronous training
+job using caikit.core.train
+"""
+
+# Standard
+from typing import Type
+import argparse
+import importlib
+import json
+import os
+import sys
+
+# Third Party
+from google.protobuf import json_format
+
+# First Party
+import alog
+
+# Local
+from ..core import ModuleBase, train
+from ..core.data_model import TrainingStatus
+from ..core.exceptions import error_handler
+from ..core.registries import module_registry
+from ..core.toolkit.logging import configure as config_logging
+from .names import get_service_package_name
+from .service_factory import ServicePackageFactory
+from .utils.servicer_util import build_caikit_library_request_dict
+
+log = alog.use_channel("TRAIN")
+error = error_handler.get(log)
+
+
+def main():
+    """Main entrypoint for running training jobs"""
+    parser = argparse.ArgumentParser(description=__doc__)
+
+    # Required Args
+    parser.add_argument(
+        "--training-kwargs",
+        "-k",
+        required=True,
+        help="Json string or json file pointer with keyword args for the training job",
+    )
+    parser.add_argument(
+        "--module",
+        "-m",
+        required=True,
+        help="Module name (package.Class) or UID to train",
+    )
+    parser.add_argument(
+        "--model-name",
+        "-n",
+        required=True,
+        help="Name to save the model under",
+    )
+
+    # Optional args
+    parser.add_argument(
+        "--save-path",
+        "-s",
+        default=".",
+        help="Path to save the output model to",
+    )
+    parser.add_argument(
+        "--library",
+        "-l",
+        nargs="*",
+        help="Libraries that need to be imported to register the module to train",
+    )
+    parser.add_argument(
+        "--trainer",
+        "-t",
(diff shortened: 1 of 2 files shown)


caikit/runtime/train.py
```diff
@@ -0,0 +1,198 @@
...
+        "--save-path",
+        "-s",
+        default=".",
+        help="Path to save the output model to",
+    )
+    parser.add_argument(
+        "--library",
+        "-l",
+        nargs="*",
+        help="Libraries that need to be imported to register the module to train",
```
gkumbhat, line 83: nit: `Caikit libraries` ?

caikit/runtime/train.py
```diff
@@ -0,0 +1,198 @@
...
+            with open(args.training_kwargs, encoding="utf-8") as handle:
+                training_kwargs = json.load(handle)
+        else:
+            training_kwargs = json.loads(args.training_kwargs)
+    except json.decoder.JSONDecodeError:
+        log.error(
+            "<COR65834760E>",
+            "--training-kwargs must be valid json or point to a valid json file",
+        )
+        raise
```
gkumbhat, line 149: not sure how the error handling works in the new system, but do we need to take above message and throw it back here with a `raise`?

caikit/runtime/train.py
```diff
@@ -0,0 +1,198 @@
...
+        future = train(module, wait=True, **train_kwargs)
+        info = future.get_info()
+        if info.status == TrainingStatus.COMPLETED:
+            log.info("Training finished successfully")
+            return 0
+        else:
+            log.error("Training finished unsuccessfully")
+            for err in info.errors or []:
+                log.error(err)
+            return 1
```
gkumbhat, line 194: concerned about the error handling here. In case of validation or runtime errors, currently the model's train function with throw error, which then will get rethrown ?

gkumbhat: Is there a mechanism to surface the actual error to user in the script consuming this script?

caikit/runtime/train.py
gabe-l-hart: Not sure that needs to be specified? I think the most correct term would be `Python libraries` since it is importing arbitrary python libs.

caikit/runtime/train.py
gabe-l-hart: `raise` with no argument will automatically raise the top exception on the exception stack, so it will be the `jsoin.decoder.JSONDecodeError`. I figured it was best to just let that pass through for ease of debugging

caikit/runtime/train.py
gabe-l-hart: Good question. Right now, no, there is no better mechanism beyond looking at the logs. Assuming that this is the `command` to the `Job` that eventually gets run, I'm not sure there's any more sophisticated mechanism to report errors beyond logging them, right? I suppose we could/should see if it's possible to get the full stack trace out of that `log.error(err)` line, though

gabe-l-hart:
I can confirm that the stack trace for training exceptions is logged:

```sh
# Run with poison pill batch_size
LOG_FORMATTER=pretty \
PYTHONPATH=$PWD/tests/fixtures \
python \
    -m caikit.runtime.train \
    -m sample_lib.modules.sample_task.sample_implementation.SampleModule \
    -n "my-model" \
    -k '{"training_data": {"jsondata": {"data": [{"number": 1, "label": "foo"}, {"number": 2, "label": "bar"}]}}, "batch_size": 999}'
```

```
2024-04-04T18:24:37.999248 [DESTROY-THRD:WARN:6151974912] <COR16788843W> Work for <bound method LocalModelTrainer.LocalModelFuture._train_and_save of <caikit.core.model_management.local_model_trainer.LocalModelTrainer.LocalModelFuture object at 0x11ab35d00>> threw exception: Batch size of 999 is not allowed! {"stack_trace": "Traceback (most recent call last):\n  File \"/Users/ghart/Projects/github/caikit/caikit/caikit/core/toolkit/concurrency/destroyable_thread.py\", line 124, in run\n    self.__runnable_result = self.runnable_func(\n  File \"/Users/ghart/Projects/github/caikit/caikit/caikit/core/model_management/local_model_trainer.py\", line 222, in _train_and_save\n    trained_model = self._module_class.train(*args, **kwargs)\n  File \"/Users/ghart/Projects/github/caikit/caikit/tests/fixtures/sample_lib/modules/sample_task/sample_implementation.py\", line 200, in train\n    raise ValueError(\nValueError: Batch size of 999 is not allowed!\n"}
2024-04-04T18:24:37.999389 [TRAIN       :ERRR:8721611456] Training finished unsuccessfully
2024-04-04T18:24:37.999420 [TRAIN       :ERRR:8721611456] Batch size of 999 is not allowed!
2024-04-04T18:24:37.999445 [TRAIN       :INFO:8721611456] Finished training my-model in: 0:00:00.000679
```

gkumbhat (approved):
Looks good to me!!

Issue #695, opened 2024-04-15
Add support for alchemy-logging's protocols as alternative to a `logger` object


mynhardtburger:
## Describe the bug

Caikit has `alchemy-logging` as a requirement, but doesn't support the recently added protocols which `alchemy-logging` implements. Eg. `alog.use_channel()` returns an object which implements `ALogLoggerProtocol`. `ALogLoggerProtocol` describes the methods which alog injected into a `logger` object. 

However [`caikit.core.exceptions.error_handler.get()`](https://github.com/caikit/caikit/blob/main/caikit/core/exceptions/error_handler.py#L41) doesn't recognise an object implementing `ALogLoggerProtocol` as a valid argument:
![image](https://github.com/caikit/caikit/assets/24638547/1c2cbb0c-a3ef-4b22-af3b-c7616e26ed19)

## Expected behavior

`caikit.core.exceptions.error_handler.get()` should additionally recognise an object which implements the [alog protocols](https://github.com/IBM/alchemy-logging/blob/main/src/python/alog/protocols.py) as a valid argument.

PR #696, opened 2024-04-15
Add alog protocol support

mynhardtburger:
closes https://github.com/caikit/caikit/issues/695

Update typing for `error_handler` to include [`LoggerProtocol`](https://github.com/IBM/alchemy-logging/blob/2bc2d8127ac878c55d5faa6c65c8fba31c4f1cf9/src/python/alog/protocols.py#L135). 

`LoggerProtocol` represents the interface for the stdlib `logger` class. `ALogLoggerProtocol`, which inherits from `LoggerProtocol`, will also be an acceptable argument to `error_handler.get()` and `ErrorHandler`.

**If applicable**:

### DIFF
--- a/caikit/core/exceptions/error_handler.py
+++ b/caikit/core/exceptions/error_handler.py
@@ -19,9 +19,12 @@
 # Standard
 from collections.abc import Iterable
 from types import GeneratorType
-from typing import TYPE_CHECKING, Dict, NoReturn, Optional, Type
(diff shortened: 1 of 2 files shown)


gabe-l-hart (changes requested):
Thanks Mynhardt! I think we also need to pin `alchemy-logging` >= `1.3.2` [here](https://github.com/caikit/caikit/blob/main/pyproject.toml#L16).

gabe-l-hart (approved):
Looks good!

Issue #700, opened 2024-04-22
Consolidate the package naming for `runtime` services


gabe-l-hart:
## Is your feature request related to a problem? Please describe.

There is currently a discrepancy in how the `package` is named the `caikit.interfaces.runtime.data_model` objects:

* The `info` objects use [`caikit_data_model.common.runtime` here](https://github.com/caikit/caikit/blob/main/caikit/interfaces/runtime/data_model/info.py#L31)
* The `training_management` objects use [`caikit_data_model.runtime` here](https://github.com/caikit/caikit/blob/main/caikit/interfaces/runtime/data_model/training_management.py#L29)

## Describe the solution you'd like

* All `runtime` data objects should use `caikit_data_model.runtime`
* The `caikit_data_model` portion should be populated with a reference to the [base constant `CAIKIT_DATA_MODEL`](https://github.com/caikit/caikit/blob/main/caikit/core/data_model/dataobject.py#L73)

## Describe alternatives you've considered

We could just leave it!

## Additional context

This would be an API breaking change, so I've opted to not make it now. It would be good to include this the next time an API breaking change is needed.

PR #702, opened 2024-04-25
Model service

### LINKED ISSUE #486: REST support for /get_training_status

We need REST support for `get_training_status`

REST APIs to query `get_training_status`

Don't have support for REST, instead rely on gRPC, which is bad.

## Additional context

### LINKED ISSUE #579: Consolidate http errors
## Description

As a Caikit developer, I would like to consolidate all `http_errors` so that they are consistent across all APIs provided by the REST server.

## Acceptance Criteria

- [ ] Unit tests cover new/changed code
- [ ] Examples build against new/changed code
- [ ] READMEs are updated
- [ ] Type of [semantic version](https://semver.org/) change is identified

gabe-l-hart:

Closes #579 
Closes #486 

This PR adds a new runtime `servicer` for model management with two new endpoints: `DeployModel` and `UndeployModel`. Additionally, it extends support for this service and the training management service to the `http_server`.

The logic of the `deploy`/`undeploy` is located in `caikit/runtime/model_management/model_manager.py` and centers around `local_models_dir`. These endpoints will raise errors if `local_models_dir` is not enabled.

This is a first-pass at a deploy mechanism that had known limitations, specifically for medium and large models. The immediate need for this feature is driven by Prompt Tune models from `caikit-nlp` which are extremely light and can easily be transmitted in a single unary message. In the future, the `model_files` field of `DeployModelRequest` may be extended to support other types of file sources, possibly up to the point of something equivalent to `DataStreamSource` with extensible plugins.

The other main limitation of this implementation is the reliance on `local_models_dir` for replica synchronization. For this to work with replicas, it is assumed that disk operations to a shared volume mounted as `local_models_dir` are atomic and therefore once the model files are written, they will be immediately visible to all other replicas besides the one that received the deploy request. When run with lazy loading enabled, this will effectively synchronize the new models to be loaded on the first inference call (or periodic sync) in each replica.

A more robust solution to the reliance on `local_models_dir` would be to hoist all logic for `local_models_dir` into a state management abstraction with local disk being the default implementation. This is a much bigger abstraction task than we want to undertake at the moment, however.

**If applicable**:

### DIFF
--- a/caikit/core/toolkit/name_tools.py
+++ b/caikit/core/toolkit/name_tools.py
@@ -16,7 +16,19 @@
 and other Protobuf names
 """
 
+# Standard
+import re
+
 
 def snake_to_upper_camel(string: str) -> str:
     """Simple snake -> upper camel conversion for descriptors"""
     return "".join([part[0].upper() + part[1:] for part in string.split("_") if part])
+
+
+def camel_to_snake_case(string: str, kebab_case: bool = False) -> str:
+    """Convert from CamelCase (or camelCase) to snake_case or kebab-case"""
+    return re.sub(
+        r"(?<!^)(?=[A-Z])",
+        "-" if kebab_case else "_",
+        string,
+    ).lower()
--- a/caikit/interfaces/runtime/data_model/package.py
+++ b/caikit/interfaces/runtime/data_model/package.py
@@ -11,21 +11,11 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+"""
+Package constant for all runtime service data model objects
+"""
 
 # Local
-from . import training_management
-from .info import (
-    ModelInfo,
-    ModelInfoRequest,
-    ModelInfoResponse,
-    RuntimeInfoRequest,
-    RuntimeInfoResponse,
-)
-from .model_management import DeployModelRequest, UndeployModelRequest
-from .training_management import (
-    ModelPointer,
-    TrainingInfoRequest,
-    TrainingJob,
-    TrainingStatus,
-    TrainingStatusResponse,
-)
+from caikit.core.data_model import CAIKIT_DATA_MODEL
+
+RUNTIME_PACKAGE = f"{CAIKIT_DATA_MODEL}.runtime"
--- a/caikit/runtime/http_server/http_server.py
+++ b/caikit/runtime/http_server/http_server.py
@@ -363,6 +363,76 @@ class RuntimeHTTPServer(RuntimeServerBase):
         log.debug4("Server healthy")
         return "OK"
 
+    async def _deploy_model(self, context: Request) -> Response:
+        """POST handler for deploying a model"""
+        assert hasattr(
+            self, "_deploy_pydantic_request"
+        ), "Cannot call _deploy_model without _bind_model_management_routes"
+        try:
+            request = await pydantic_from_request(
+                self._deploy_pydantic_request, context
+            )
+            result = self.model_management_servicer.deploy_model(
+                request.model_id,
+                {f.filename: f.data for f in request.model_files},
+            )
+            return Response(
+                content=result.to_json(),
+                media_type="application/json",
+            )
+        except Exception as err:
+            if error_content := self._handle_exception(err):
+                return Response(
+                    content=json.dumps(error_content),
+                    status_code=error_content["code"],
+                )
+            raise
+
+    async def _undeploy_model(self, model_id: Annotated[str, Query]) -> Response:
+        """DELETE handler for undeploying a model"""
+        try:
+            result = self.model_management_servicer.undeploy_model(model_id)
+            return Response(content=result.to_json(), media_type="application/json")
+        except Exception as err:
+            if error_content := self._handle_exception(err):
+                return Response(
+                    content=json.dumps(error_content),
+                    status_code=error_content["code"],
+                )
+            raise
+
+    def _get_training_status(self, training_id: Annotated[str, Query]) -> Response:
+        """GET handler for fetching a training"""
+        try:
+            result = self.training_management_servicer.get_training_status(training_id)
+            return Response(
+                content=result.to_json(),
+                media_type="application/json",
+            )
+        except Exception as err:
+            if error_content := self._handle_exception(err):
+                return Response(
+                    content=json.dumps(error_content),
+                    status_code=error_content["code"],
+                )
+            raise
+
+    def _cancel_training(self, training_id: Annotated[str, Query]) -> Response:
+        """DELETE handler for undeploying a model"""
+        try:
+            result = self.training_management_servicer.cancel_training(training_id)
+            return Response(
+                content=result.to_json(),
+                media_type="application/json",
+            )
+        except Exception as err:
+            if error_content := self._handle_exception(err):
+                return Response(
+                    content=json.dumps(error_content),
+                    status_code=error_content["code"],
+                )
+            raise
+
     #####################
     ## Request Binding ##
     #####################
@@ -403,37 +473,23 @@ class RuntimeHTTPServer(RuntimeServerBase):
         )
         deploy_pydantic_response = dataobject_to_pydantic(deploy_dataobject_response)
 
-        @self.app.post(
+        # Bind deploy_model
+        # NOTE: The deploy_pydantic_request must be bound to `self` so that it
+        #   it does not need to be bound to the `_deploy_model` function which
+        #   is hard since its async.
+        self._deploy_pydantic_request = deploy_pydantic_request
+        self.app.post(
             MODEL_MANAGEMENT_ENDPOINT,
             responses=self._get_response_openapi(
                 deploy_dataobject_response, deploy_pydantic_response
             ),
             openapi_extra=self._get_request_openapi(deploy_pydantic_request),
             response_class=Response,
-        )
-        async def _deploy_model(context: Request) -> Response:
-            """POST handler for deploying a model"""
-            try:
-                request = await pydantic_from_request(deploy_pydantic_request, context)
-                result = self.model_management_servicer.DeployModel(request, None)
-                return Response(
-                    content=deploy_dataobject_response.from_proto(result).to_json(),
-                    media_type="application/json",
-                )
-            except Exception as err:
-                if error_content := self._handle_exception(err):
-                    return Response(
-                        content=json.dumps(error_content),
-                        status_code=error_content["code"],
-                    )
-                raise
+        )(self._deploy_model)
 
         # Bind DELETE to undeploy a model
         undeploy_spec = MODEL_MANAGEMENT_SERVICE_SPEC["service"]["rpcs"][1]
         assert undeploy_spec["name"] == "UndeployModel"
-        undeploy_dataobject_request = DataBase.get_class_for_name(
-            undeploy_spec["input_type"]
-        )
         undeploy_dataobject_response = DataBase.get_class_for_name(
             undeploy_spec["output_type"]
         )
@@ -441,30 +497,13 @@ class RuntimeHTTPServer(RuntimeServerBase):
             undeploy_dataobject_response
         )
 
-        @self.app.delete(
+        self.app.delete(
             MODEL_MANAGEMENT_ENDPOINT,
             responses=self._get_response_openapi(
                 undeploy_dataobject_response, undeploy_pydantic_response
             ),
             response_class=Response,
-        )
-        def _undeploy_model(model_id: Annotated[str, Query]) -> Response:
-            """DELETE handler for undeploying a model"""
-            try:
-                result = self.model_management_servicer.UndeployModel(
-                    undeploy_dataobject_request(model_id).to_proto(), None
-                )
-                return Response(
-                    content=undeploy_dataobject_response.from_proto(result).to_json(),
-                    media_type="application/json",
-                )
-            except Exception as err:
-                if error_content := self._handle_exception(err):
-                    return Response(
-                        content=json.dumps(error_content),
-                        status_code=error_content["code"],
-                    )
-                raise
+        )(self._undeploy_model)
 
     def _bind_training_management_routes(self):
         """Bind the routes for get/cancel trainings"""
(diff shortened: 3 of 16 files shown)


caikit/core/toolkit/name_tools.py
```diff
@@ -16,7 +16,19 @@
...
+# Standard
+import re
+
 
 def snake_to_upper_camel(string: str) -> str:
     """Simple snake -> upper camel conversion for descriptors"""
     return "".join([part[0].upper() + part[1:] for part in string.split("_") if part])
+
+
+def camel_to_snake_case(string: str, kebab_case: bool = False) -> str:
```
gabe-l-hart, line 28: Moved here for reuse out of `names.py`

caikit/interfaces/runtime/data_model/package.py
```diff
@@ -0,0 +1,21 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+Package constant for all runtime service data model objects
+"""
+
+# Local
+from caikit.core.data_model import CAIKIT_DATA_MODEL
+
+RUNTIME_PACKAGE = f"{CAIKIT_DATA_MODEL}.runtime"
```
gabe-l-hart, line 21: We were missing a common constant for the `interfaces.runtime` package. As such, there's [inconsistency in the packages](https://github.com/caikit/caikit/issues/700) which I opted not to fix at the moment to avoid an API breaking change.

caikit/interfaces/runtime/data_model/training_management.py
```diff
@@ -18,37 +18,28 @@
...
-from caikit.core.toolkit.wip_decorator import Action, WipCategory, work_in_progress
+from ....core.data_model import DataObjectBase, TrainingStatus, dataobject
+from .package import RUNTIME_PACKAGE
 
-log = alog.use_channel("MDLOPS")
 
-RUNTIME_PACKAGE = "caikit_data_model.runtime"
-
-
-@work_in_progress(action=Action.WARNING, category=WipCategory.BETA)
```
gabe-l-hart, line 32: This isn't really doing anything at this point beyond spitting out warnings. It's in real usage, so it's no longer WIP.

caikit/runtime/http_server/http_server.py
```diff
@@ -318,19 +341,34 @@ def stop(self):
         if self.interrupter:
             self.interrupter.stop()
 
-    ##########
-    ## Impl ##
```
gabe-l-hart, line 322: I did a lot of rearranging here. Sorry! The `Impl` section had too many functions of different flavors, so I broke it up into `Static Endpoints`, `Request Binding`, and `Helpers`.

caikit/runtime/http_server/http_server.py
```diff
@@ -350,49 +388,154 @@ def _bind_routes(self, service: ServicePackage):
             elif isinstance(rpc, ModuleClassTrainRPC):
                 self._train_add_unary_input_unary_output_handler(rpc)
 
-    def _get_model_id(self, request: Type[pydantic.BaseModel]) -> str:
```
gabe-l-hart, line 353: Moved to the `Helpers` section

caikit/runtime/http_server/http_server.py
```diff
@@ -350,49 +388,154 @@ def _bind_routes(self, service: ServicePackage):
...
-        if model_id is None:
-            raise CaikitRuntimeException(
-                status_code=StatusCode.INVALID_ARGUMENT,
-                message="Please provide model_id in payload",
-            )
-        return model_id
+    def _bind_model_management_routes(self):
+        """Bind the routes for deploy/undeploy"""
 
-    def _get_request_params(
```
gabe-l-hart, line 364: Moved to the `Helpers` section

caikit/runtime/http_server/http_server.py
```diff
@@ -607,30 +714,105 @@ async def _generator() -> pydantic_response:
...
                 )
 
             return EventSourceResponse(_generator())
 
+    #############
+    ## Helpers ##
+    #############
+
+    @staticmethod
+    def _handle_exception(err: Exception) -> Optional[dict]:
```
gabe-l-hart, line 731: This new function consolidates all error handling to a single location for consistency

caikit/runtime/http_server/http_server.py
```diff
@@ -774,40 +956,8 @@ def _get_response_openapi(
         output = {200: {"content": response_schema}}
         return output
 
-    def _model_info(
```
gabe-l-hart, line 777: Moved to `Static Endpoints`

caikit/runtime/http_server/http_server.py
```diff
@@ -774,40 +956,8 @@ def _get_response_openapi(
...
-            return error_content
-
-    @staticmethod
-    def _health_check() -> str:
-        log.debug4("Server healthy")
-        return "OK"
-
     @contextmanager
-    def _tls_files(self) -> _TlsFiles:
+    def _tls_files(self) -> Iterable[_TlsFiles]:
```
gabe-l-hart, line 960: Fixed type hint

caikit/runtime/model_management/model_manager.py
```diff
@@ -325,18 +326,19 @@ def unload_model(self, model_id) -> int:
             Model_size (int) : Size of the loaded model in bytes
         """
         log.debug("List of loaded models: %s", str(self.loaded_models))
-        # If the model failed to load, just return 0; no need to throw an error here.
-        if model_id not in self.loaded_models:
-            log.debug("Model '%s' is not loaded, so it cannot be unloaded!", model_id)
-            return 0
+        try:
```
gabe-l-hart, line 329: Rearranged here for a small efficiency to avoid a potential race around looking up in the map twice (once to check presence and once again to remove)

caikit/runtime/service_factory.py
```diff
@@ -59,42 +62,6 @@
 log = alog.use_channel("SVC-FACTORY")
 error = error_handler.get(log)
 
-TRAINING_MANAGEMENT_SERVICE_NAME = "TrainingManagement"
```
gabe-l-hart, line 62: It turns out, all of this had been half-way moved to `names.py`, but never removed from here. I consolidated it all over there.

caikit/runtime/servicers/model_management_servicer.py
```diff
@@ -0,0 +1,94 @@
...
+
+class ModelManagementServicerImpl:
+    __doc__ = __doc__
+
+    def __init__(self):
+        self._model_manager = ModelManager.get_instance()
+
+    def DeployModel(
+        self,
+        request: DeployModelRequestProto,  # type: ignore
```
gabe-l-hart, line 50: The need to type ignore here is sad, but I could not coerce the type checker to allow the return value of `get_proto_class` to be a valid type hint, no matter how much I tried to tell it what the class was.

tests/runtime/http_server/test_http_server.py
```diff
@@ -265,14 +335,73 @@ def test_docs_with_models(
...
+
+
 def test_inference_sample_task(sample_task_model_id, client):
     """Simple check that we can ping a model"""
     json_input = {"inputs": {"name": "world"}, "model_id": sample_task_model_id}
     response = client.post(
         f"/api/v1/task/sample",
         json=json_input,
     )
-    json_response = json.loads(response.content.decode(response.default_encoding))
```
gabe-l-hart, line 275: This was a rampant anti-pattern in this file, so I fixed it everywhere to avoid folks wondering down the line why some used `response.json()` and others used this complex parsing

gabe-l-hart (commented):
Left manual comments to help with reviewer parsing

caikit/runtime/http_server/http_server.py
joerunde: 🌶️🌶️🌶️

I like the organization! While we're at it, can we take this to the logical conclusion of splitting out http servicers for separate tasks? This file is getting suuuuper chonky with everything still piled into one class.

Since the new endpoint is static, I actually expected to see something more like this:

```
class HttpModelManagermentServicer:

    @app.post(MODEL_MANAGEMENT_ENDPOINT, ...)
    async def deploy_model(self, stuff...)
        # do the deploy things
```

Related question: Why are this and the training management endpoints not in the `Static Endpoints` section here as is? I thought they were static, as opposed to the dynamically registered inference and training endpoints that depend on the caikit library implementation.

tests/runtime/model_management/test_model_manager.py
```diff
@@ -773,6 +774,164 @@ def test_lazy_load_ephemeral_model():
...
+                },
+                "merge",
+            ) as managers:
+                manager = managers[0]
+                model_name = "my-model"
+
+                # Make sure the deploy fails with a permission error
+                with pytest.raises(CaikitRuntimeException) as excinfo:
+                    manager.deploy_model(model_name, deploy_good_model_files, wait=True)
+                    assert excinfo.status_code == grpc.StatusCode.PERMISSION_DENIED
```
joerunde, line 908: I think PERMISSION_DENIED implies that the caller doesn't have permission to perform the action, but this error case is that the user id running the server itself does not have permission to write the files. Is this a `FAILED_PRECONDITION`?


https://grpc.github.io/grpc/core/md_doc_statuscodes.html
>  The operation was rejected because the system is not in a state required for the operation's execution. For example, the directory to be deleted is non-empty, an rmdir operation is applied to a non-directory, etc. Service implementors can use the following guidelines to decide between FAILED_PRECONDITION, ABORTED, and UNAVAILABLE: (a) Use UNAVAILABLE if the client can retry just the failing call. (b) Use ABORTED if the client should retry at a higher level (e.g., when a client-specified test-and-set fails, indicating the client should restart a read-modify-write sequence). (c) Use FAILED_PRECONDITION if the client should not retry until the system state has been explicitly fixed. E.g., if an "rmdir" fails because the directory is non-empty, FAILED_PRECONDITION should be returned since the client should not retry unless the files are deleted from the directory.

tests/runtime/model_management/test_model_manager.py
```diff
@@ -773,6 +774,164 @@ def test_lazy_load_ephemeral_model():
...
+            },
+            "merge",
+        ) as managers:
+            manager = managers[0]
+            model_name = "my-model"
+
+            # Make sure model is not currently loaded
+            with pytest.raises(CaikitRuntimeException) as excinfo:
+                manager.retrieve_model(model_name)
+                assert excinfo.value.status_code == grpc.StatusCode.NOT_FOUND
```
joerunde, line 797: 🤔  I thought that `excinfo` had to be accessed after the scope of the context manager because raising the exception will exit from it. But if that's not the case, then TIL!

tests/runtime/http_server/test_http_server.py
```diff
@@ -1063,57 +1139,222 @@ def test_train_long_running_sample_task(client, runtime_http_server):
...
+    # Make sure missing param returns 422
+    assert client.get(http_server.TRAINING_MANAGEMENT_ENDPOINT).status_code == 422
+
+    # Make sure unknown training DELETE returns 404
+    assert (
+        client.delete(
+            http_server.TRAINING_MANAGEMENT_ENDPOINT, params={"training_id": "bad-id"}
+        ).status_code
+        == 404
+    )
```
joerunde, line 1360: image

Not a total blocker and very related to my other comment about splitting up http servicers, but a few minutes splitting this file up on some logical boundaries would pay off a lot in readability

caikit/runtime/http_server/http_server.py
```diff
@@ -350,49 +388,154 @@ def _bind_routes(self, service: ServicePackage):
...
+            response_class=Response,
+        )
+        def _undeploy_model(model_id: Annotated[str, Query]) -> Response:
+            """DELETE handler for undeploying a model"""
+            try:
+                result = self.model_management_servicer.UndeployModel(
+                    undeploy_dataobject_request(model_id).to_proto(), None
+                )
+                return Response(
+                    content=undeploy_dataobject_response.from_proto(result).to_json(),
```
joerunde, line 458: It looks like we're doing a lot of extra translation? The request is going:
`json -> data model -> protobuf ` and then the response is going
`data model -> protobuf -> data model -> json`

Probably not a performance concern since this is a deploy-time API, but certainly adds complexity and inflates the bug surface.

Any chance we can refactor that out? Maybe just call the model manager directly here and not involve protobuf at all?

tests/runtime/model_management/test_model_manager.py
```diff
@@ -773,6 +774,164 @@ def test_lazy_load_ephemeral_model():
...
+            # local models dir
+            with pytest.raises(CaikitRuntimeException) as excinfo:
+                manager.retrieve_model(model_name)
+                assert excinfo.value.status_code == grpc.StatusCode.NOT_FOUND
+            assert not os.path.exists(os.path.join(cache_dir, model_name))
+
+
+@pytest.mark.parametrize(
+    "invalid_fname",
+    ["", "\t\n  ", "/foo/bar.txt"],
```
joerunde, line 828: nit: can we add an expected reason match as well for these? Like `".*Cannot use absolute paths for model files.*"`

and then `with pytest.raises(CaikitRuntimeException, match=expected_reason)`

caikit/runtime/names.py
```diff
@@ -215,7 +221,7 @@ def get_task_predict_request_name(
 ##  Service Definitions
 
 TRAINING_MANAGEMENT_SERVICE_NAME = "TrainingManagement"
-TRAINING_MANAGEMENT_PACKAGE = "caikit.runtime.training"
+TRAINING_MANAGEMENT_SERVICE_PACKAGE = "caikit.runtime.training"
```
joerunde, line 224: nit: Is it worth renaming these to `*_PACKAGE_NAME` to make it clear that they're string constants?

tests/runtime/model_management/test_model_manager.py
gabe-l-hart: Ah, very good call. Will switch to `FAILED_PRECONDITION`

tests/runtime/model_management/test_model_manager.py
gabe-l-hart: Now that you say it, I think you're right. That feels wrong to me since it's a contextmanager, but 🤷

caikit/runtime/http_server/http_server.py
gabe-l-hart: Yeah, this does kind of suck. I don't think there's a simpler way to do this without a _big_ refactor that puts `pydantic` at the heart of `dataobject` unfortunately. There's no guarantee that `json -> protobuf` or `protobuf -> json` will work since individual `data model` objects may have custom logic for to/from json/proto. In this specific case, we know they don't but I would be very unsurprised if it were added down the line to support special upload capabilities and I wouldn't want this to suddenly break.

tests/runtime/model_management/test_model_manager.py
gabe-l-hart: Good call

caikit/runtime/names.py
gabe-l-hart: We've used just `package` in this context in a lot of places (e.g. `package.py`). I can see that being confusing for those not in the guts, but I don't really want to go find all the places this is under-specified in this PR. Ok to skip for now?

tests/runtime/http_server/test_http_server.py
gabe-l-hart: On the one hand, I _totally_ agree that there should be a full overhaul of the modularity in `http_server`, and I've taken a first step in this PR, so I'm already ankles-deep. I'd prefer not to further bloat this feature PR with the full refactor though. Good to skip and revisit?

caikit/runtime/http_server/http_server.py
gabe-l-hart: >Related question: Why are this and the training management endpoints not in the Static Endpoints section here as is? I thought they were static, as opposed to the dynamically registered inference and training endpoints that depend on the caikit library implementation.

Very valid question. The reason I did them as dynamic methods was the need to compute the input/output types from data objects. Technically, I could have just hard-coded the data objects because I _do_ know them, but this would create multiple places where the definition of the service is explicitly coded (service definition and here). I preferred to keep the service definition as the single source-of-truth so that any changes down the line do not require mucking with these handlers.

>While we're at it, can we take this to the logical conclusion of splitting out http servicers for separate tasks? This file is getting suuuuper chonky with everything still piled into one class.

Yeah, we probably _should_ do this, but I'm afraid it would make this PR even more of a "general purpose refactor" which was not its intent. I'd prefer to open a ticket and revisit in a separate PR.

caikit/runtime/model_management/model_manager.py
```diff
@@ -450,6 +452,107 @@ def retrieve_model(self, model_id: str) -> ModuleBase:
...
+        artifacts from the local models dir.
+        """
+        error.value_check(
+            "<RUN05068606E>",
+            self._local_models_dir,
+            "runtime.local_models_dir must be a valid path to undeploy models directly.",
+        )
+
+        # Get a handle to the LoadedModel
+        model = self.loaded_models.get(model_id)
```
gkumbhat, line 542: In case when we have multiple caikit pods, and only 1 of them have this model in memory, because we don't have a global shared state of models among caikit pods, this will not be able to "undeploy" the model right?

caikit/runtime/http_server/http_server.py
joerunde: What I'm getting at here is a smaller refactor, not overhauling dataobjects. Instead of having this complexity around here to wrap the grpc servicers (which is generic and very cool and all), can't we just write the logic for the requests and responses inline since they're gonna be reasonably static?

```
def _undeploy_model(model_id: Annotated[str, Query]) -> Response:    
    self.model_manager.undeploy(model_id)
    return Response(content={"model_id": model_id})

def _deploy_model(context: Request) -> Response:
    request = await pydantic_from_request(deploy_pydantic_request, context)
    loaded_model = self._model_manager.deploy_model(
        model_id=request.model_id,
        model_files={f.filename: f.data for f in request.model_files},
        wait=False,
    )

    model_info = ModelInfo(
        model_path=loaded_model.path(),
        name=loaded_model.id(),
        size=loaded_model.size(),
        loaded=loaded_model.loaded(),
    )
    
    return Response(content=model_info.to_json())
```

caikit/runtime/model_management/model_manager.py
gabe-l-hart: These APIs will _only_ work if we have a shared volume for `local_models_dir`. That said, I think you found a race where the other replica has not yet loaded the deployed model (no inference, periodic poll hasn't triggered yet). In this case, the undeploy would fail. I'll fix that!

caikit/runtime/names.py
joerunde: sure 👍

caikit/runtime/http_server/http_server.py
gabe-l-hart: Ok, I dug in deeper and you were totally right! Much simpler now

caikit/runtime/http_server/http_server.py
```diff
@@ -318,19 +341,104 @@ def stop(self):
...
+                    content=json.dumps(error_content), status_code=error_content["code"]
+                )
+            raise
+
+    @staticmethod
+    def _health_check() -> str:
+        log.debug4("Server healthy")
+        return "OK"
+
+    async def _deploy_model(self, context: Request) -> Response:
```
joerunde, line 366: hell yeah this is sick now

joerunde (approved):
get it

PR #708, opened 2024-05-03
Multitask train modules 707

### LINKED ISSUE #707: Training endpoints change with multitask module
## Describe the bug

In an attempt to fix [a bug on the caikit-nlp side](https://github.com/caikit/caikit-nlp/issues/350) where tasks added to modules did not seem invokable, an unintentional change was triggered for the training side. The multiple tasks were relevant for inference, but the training endpoint also changed (from text generation to tokenization task).

## Expected behavior

Training endpoints should not be unexpectedly changed with task additions.

## Observed behavior

Prompt tuning changed from `TextGenerationTaskPeftPromptTuningTrain` to `TokenizationTaskPeftPromptTuningTrain`

gabe-l-hart:

Closes #707 

This PR updates the logic for how the `ModuleClass.tasks` property is implemented. The previous use of a `set` was there to ensure uniqueness. It also had the subtle implication that all tasks are equally important to a given `ModuleClass` since a `set` does not guarantee order.

This second implication, however, was not respected in the train API generation. Since a given `ModuleClass` can have only a single `train` method, the mapping from `ModuleClass` to `task` was nondeterministic for multitask module training. With this PR, the order of `tasks` is now strictly defined to be the unique concatenation of the list of tasks in the module itself with the task lists of all parent modules in MRO.

**If applicable**:

### DIFF
--- a/caikit/core/modules/decorator.py
+++ b/caikit/core/modules/decorator.py
@@ -188,15 +188,14 @@ def module(
         cls_.MODULE_CLASS = classname
         cls_.PRODUCER_ID = ProducerId(cls_.MODULE_NAME, cls_.MODULE_VERSION)
 
-        cls_._TASK_CLASSES = tasks
-
         # Parse the `train` and `run` signatures
         cls_.RUN_SIGNATURE = CaikitMethodSignature(cls_, "run")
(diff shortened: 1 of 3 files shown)


evaline-ju (approved):
LGTM!

gkumbhat (approved):
Looks good to me!

PR #709, opened 2024-05-03
Add exception catching for everything in train.py

kellyaa:
Because train.py is being launched via the command line, we need to convert all thrown exceptions to exit codes. When running in a k8s container environment, the error message must also be written to the termination log.

### DIFF
--- a/caikit/runtime/train.py
+++ b/caikit/runtime/train.py
@@ -158,7 +158,7 @@ def main() -> int:
         for library in args.library or []:
             log.info("<COR88091092I>", "Importing library %s", library)
             importlib.import_module(library)
-    except Exception as e:
+    except Exception:
         message = "Unable to import module {}".format(library)
         log.warning(
             {
@@ -193,7 +193,7 @@ def main() -> int:
             "Unable to find module {} to train",
             args.module,
         )
-    except (ValueError, Exception) as e:
+    except (ValueError, Exception):
         message = "Unable to find module {} to train".format(args.module)
         log.warning(
             {
@@ -262,7 +262,7 @@ def main() -> int:
                 "stack_trace": traceback.format_exc(),
             }
         )
-    except Exception as e:
+    except Exception:
         message = "Exception encountered when attempting to parse input parameters"
         log.warning(
             {
@@ -292,7 +292,7 @@ def main() -> int:
                     log.error(err)
                 write_termination_log("Training finished unsuccessfully")
                 return INTERNAL_ERROR_EXIT_CODE
-    except MemoryError as e:
+    except MemoryError:
         message = "OOM error during training"
         log.warning(
             {
@@ -303,7 +303,7 @@ def main() -> int:
         )
         write_termination_log(message)
         exit(INTERNAL_ERROR_EXIT_CODE)
-    except Exception as e:
+    except Exception:
         message = "Unhandled exception during training"
         log.warning(
             {
--- a/caikit/runtime/train.py
+++ b/caikit/runtime/train.py
@@ -155,7 +155,8 @@ def main() -> int:
                 "log_code": "<COR39662029E>",
                 "message": message,
                 "stack_trace": traceback.format_exc(),
-            }
+            },
+            exc_info=True,
         )
         write_termination_log(message)
         exit(USER_ERROR_EXIT_CODE)
@@ -172,7 +173,8 @@ def main() -> int:
                 "log_code": "<COR17776539E>",
                 "message": message,
                 "stack_trace": traceback.format_exc(),
-            }
+            },
+            exc_info=True,
         )
         write_termination_log(message)
         exit(USER_ERROR_EXIT_CODE)
@@ -269,7 +271,8 @@ def main() -> int:
                 "log_code": "<COR65474760E>",
                 "message": message,
                 "stack_trace": traceback.format_exc(),
-            }
+            },
+            exc_info=True,
         )
     except Exception:
         message = "Exception encountered when attempting to parse input parameters"
@@ -278,7 +281,8 @@ def main() -> int:
                 "log_code": "<COR17776549E>",
                 "message": message,
                 "stack_trace": traceback.format_exc(),
-            }
+            },
+            exc_info=True,
         )
         write_termination_log(message)
         exit(USER_ERROR_EXIT_CODE)
@@ -308,7 +312,8 @@ def main() -> int:
                 "log_code": "<COR04280062E>",
                 "message": message,
                 "stack_trace": traceback.format_exc(),
-            }
+            },
+            exc_info=True,
         )
         write_termination_log(message)
         exit(INTERNAL_ERROR_EXIT_CODE)
@@ -319,7 +324,8 @@ def main() -> int:
                 "log_code": "<COR04280062E>",
                 "message": message,
                 "stack_trace": traceback.format_exc(),
-            }
+            },
+            exc_info=True,
         )
         write_termination_log(message)
         exit(INTERNAL_ERROR_EXIT_CODE)
--- a/caikit/runtime/train.py
+++ b/caikit/runtime/train.py
@@ -51,6 +51,7 @@ INTERNAL_ERROR_EXIT_CODE = 200
 
 class ArgumentParserError(Exception):
     """Custom exception class for ArgumentParser errors."""
+
     pass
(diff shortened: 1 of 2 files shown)


caikit/runtime/train.py
```diff
@@ -44,10 +45,37 @@
...
+    def error(self, message):
+        """Error handler that raises an exception instead of exiting."""
+        raise ArgumentParserError(f"{self.prog}: error: {message}")
+
+
+def write_termination_log(text: str, log_file="/dev/termination-log"):
+    try:
+        f = open(log_file, "a")
+        f.write(text)
+        f.close()
```
gabe-l-hart, line 68:
```suggestion
        with open(log_file, "a") as handle
            handle.write(text)
```

caikit/runtime/train.py
```diff
@@ -44,10 +45,37 @@
...
 error = error_handler.get(log)
 
+USER_ERROR_EXIT_CODE = 1
+INTERNAL_ERROR_EXIT_CODE = 200
+
+
+class ArgumentParserError(Exception):
+    """Custom exception class for ArgumentParser errors."""
+
+    pass
```
gabe-l-hart, line 55:
```suggestion
```

caikit/runtime/train.py
```diff
@@ -95,44 +123,86 @@ def main() -> int:
...
+            "save_with_id": args.save_with_id,
+            "model_name": args.model_name,
+        }
+        if args.trainer is not None:
+            train_kwargs["trainer"] = args.trainer
+
+    except Exception as e:
+        message = "Exception raised during training. This may be a problem with your input: {}".format(
+            e
+        )
```
gabe-l-hart, line 150:
```suggestion
        message = f"Exception raised during training. This may be a problem with your input: {e}"
```

caikit/runtime/train.py
```diff
@@ -95,44 +123,86 @@ def main() -> int:
...
+        log.warning(
+            {
+                "log_code": "<COR39662029E>",
+                "message": message,
+                "stacK_trace": traceback.format_exc(),
```
gabe-l-hart, line 155:
```suggestion
                "stack_trace": traceback.format_exc(),
```

caikit/runtime/train.py
```diff
@@ -95,44 +123,86 @@ def main() -> int:
...
+            args.module, mod_pkg_to_mod.get(args.module)
+        )
+        error.value_check(
+            "<COR03876205E>",
+            module is not None,
+            "Unable to find module {} to train",
+            args.module,
+        )
+    except Exception as e:
+        message = "Unable to import module {}".format(args.module)
```
gabe-l-hart, line 196: We need different log messages and unique `log_code`s on all of these `except` clauses

gabe-l-hart (changes requested):
A few NITs and tightness suggestions. It would also be good to rebase on the latest `main` since the diff is showing up funny with all of the other changes since you branched.

caikit/runtime/train.py
kellyaa: thanks, fixed

caikit/runtime/train.py
kellyaa: thanks, fixed

caikit/runtime/train.py
kellyaa: ooh, good eyes. Thanks!

caikit/runtime/train.py
kellyaa: should be resolved now

caikit/runtime/train.py
```diff
@@ -44,10 +45,34 @@
 log = alog.use_channel("TRAIN")
 error = error_handler.get(log)
 
+USER_ERROR_EXIT_CODE = 1
+INTERNAL_ERROR_EXIT_CODE = 200
```
gabe-l-hart, line 49: NIT: Was there a reason you chose 200? Seems like that might confuse some folks w.r.t. HTTP 200

caikit/runtime/train.py
```diff
@@ -44,10 +45,34 @@
...
+
+def write_termination_log(text: str, log_file="/dev/termination-log"):
+    try:
+        with open(log_file, "a") as handle:
+            handle.write(text)
+    except Exception as e:
+        log.warning(
+            "<COR96300323W>",
+            "Unable to write termination log due to error {}".format(e),
+        )
```
gabe-l-hart, line 70:
```suggestion
            "Unable to write termination log due to error %s",
            e,
        )
```

gabe-l-hart: This is actually pretty important as a general practice. In this instance, it doesn't really matter because it's behind an `except`, but in general we need to make sure that all log messages use lazy interpolation (note the `,` instead of `%` between the format string and the args). One of the key tenants of `alog` is that log messages should cost nothing more than an `if` statement when disabled. If we do proactive interpolation, the log message must be created before the `if` statement is evaluated which is extremely wasteful. In the worst example I've seen, a single `%` -> `,` saved almost 25% performance of a specific `watson_nlp` module because it was buried down in a deeply nested loop.

caikit/runtime/train.py
```diff
@@ -95,44 +120,90 @@ def main() -> int:
...
+        module: Type[ModuleBase] = mod_reg.get(
+            args.module, mod_pkg_to_mod.get(args.module)
+        )
+        error.value_check(
+            "<COR03876205E>",
+            module is not None,
+            "Unable to find module {} to train",
+            args.module,
+        )
+    except (ValueError, Exception):
```
gabe-l-hart, line 196: `ValueError` is a subclass of `Exception`, so this is unnecessary

```suggestion
    except Exception:
```

caikit/runtime/train.py
```diff
@@ -141,57 +212,108 @@ def main() -> int:
...
+            info = future.get_info()
+            if info.status == TrainingStatus.COMPLETED:
+                log.info("Training finished successfully")
+                return 0
+            else:
+                log.error("Training finished unsuccessfully")
+                for err in info.errors or []:
+                    log.error(err)
+                write_termination_log("Training finished unsuccessfully")
+                return INTERNAL_ERROR_EXIT_CODE
```
gabe-l-hart, line 294: Should this be `exit`?

gabe-l-hart (changes requested):
A few more NITs to fix

caikit/runtime/train.py
kellyaa: This is true. It just needed to be a number >= 128. (generally considered a system/retryable error by projects like Kubeflow Training Operator). I can see the confusion around 200 = HTTP OK! I can change it to something else

gabe-l-hart: Ah, makes sense. It would be good to leave a note about that in the code!

caikit/runtime/train.py
kellyaa: Yeah I was debating that. I think you're right, the `exit` makes more sense here.

caikit/runtime/train.py
```diff
@@ -95,44 +127,90 @@ def main() -> int:
...
+            args.module,
+        )
+    except Exception:
+        message = "Unable to find module {} to train".format(args.module)
+        log.warning(
+            {
+                "log_code": "<COR17476539E>",
+                "message": message,
+                "stack_trace": traceback.format_exc,
+            }
```
gabe-l-hart, line 210: One more NIT! This will print the stack trace in the log message

```suggestion
            },
            exc_info=True,
```

caikit/runtime/train.py
```diff
@@ -141,57 +219,108 @@ def main() -> int:
...
-        log.error(
-            "<COR65834760E>",
-            "--training-kwargs must be valid json or point to a valid json file",
+        message = "training-kwargs must be valid json or point to a valid json file"
+        log.warning(
+            {
+                "log_code": "<COR65834760E>",
+                "message": message,
+                "stack_trace": traceback.format_exc(),
+            }
```
gabe-l-hart, line 259:
```suggestion
            },
            exc_info=True,
```

gabe-l-hart (approved):
Looks great! One more little group of NITs to include `exc_info=True` in the `log.warning` calls so that the stack trace prints. This is not a blocker, though, so I'm approving as is. If you don't have bandwidth to add these, we can merge as is.

PR #711, opened 2024-05-06
Add option to pass Context argument to Predict Servicer

HonakerM:
This PR adds a new task option to allow passing the runtime context to a module run method. This allows developers to change module logic based off the request context. For example a module could look for a HTTP header and change its internal processing.

**If applicable**:

### DIFF
--- a/caikit/runtime/http_server/http_server.py
+++ b/caikit/runtime/http_server/http_server.py
@@ -635,6 +635,7 @@ class RuntimeHTTPServer(RuntimeServerBase):
                         output_streaming=False,
                         task=rpc.task,
                         aborter=aborter,
+                        context=context,
                         **request_params,
                     )
                     result = await loop.run_in_executor(self.thread_pool, call)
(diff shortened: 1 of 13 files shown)


caikit/runtime/http_server/http_server.py
```diff
@@ -635,6 +635,7 @@ async def _handler(
                         output_streaming=False,
                         task=rpc.task,
                         aborter=aborter,
+                        context=context,
```
gabe-l-hart, line 638: The main question I have on this PR is whether we need a `caikit.core` abstract definition of a `Context` so that modules have a contractual API signature when choosing to query the context. It would mean implementing some kind of common shim between HTTP and gRPC for common operations. This would also be particularly useful for tracing so that once we've got [trace instrumentation](https://github.com/caikit/caikit/pull/719) in place, the body of a module could fetch a request ID and forward it to downstream calls.

gabe-l-hart (approved):
Looks great, thanks Casey!

PR #714, opened 2024-05-17
:goal_net: Handle caikit core exceptions in servicers

evaline-ju:

After rpc errors were interpreted as caikit core exceptions in https://github.com/caikit/caikit-nlp/pull/355, when the servicers get called, these always get interpreted as `500` e.g. `{"details": "(<CaikitCoreStatusCode.INVALID_ARGUMENT: 2>, 'input tokens (3222) plus prefix length (0) must be < 2048')", "code": 500, "id": "875b83810ee04c058357eb5124c7cea8"`

Since the mapping of status codes is already present in caikit, this can be handled here.

**If applicable**:


gabe-l-hart (approved):
Thanks Evaline! Good catch on this

PR #499, opened 2023-09-29
WIP: Replace Prometheus instrumentation with OpenTelemetry

hickeyma:
**What this PR does / why we need it**:

Replaces Prometheus instrumentation with OpenTelemetry instrumentation.

**TODO: Add more information.**

Fixes #11 

**Special notes for your reviewer**:
- TODO:
  - Docs
  - Tests
  - Remove Prometheus parts

**If applicable**:
- [ ] this PR contains documentation
- [ ] this PR contains unit tests
- [ ] this PR has been tested for backwards compatibility

PR #719, opened 2024-05-21
first version showing how otel compliant tracing can be enabled

PRATIBHA-Moogi:
**Scope of Change**:
caikit/runtime/servicers/global_predict_servicer.py inserted with global tracer and span with attributes

**Changes requires otel sdk and other dependent pkgs to be installed**:
- [ ] this PR has been tested by locally standing up embedding model server and also deployed stand-alone jaeger to visualise traces/spans with applied attributes

CC @gabe-l-hart @gkumbhat

### DIFF
--- a/caikit/config/config.yml
+++ b/caikit/config/config.yml
@@ -191,6 +191,8 @@ runtime:
         endpoint: "localhost:4317"
         protocol: grpc
         service_name: caikit.runtime
+        # Flush the trace on exit
+        flush_on_exit: true
         # TLS config for the otel server. If CA is falsy, insecure. If client
         # key and cert provided with CA, mTLS, otherwise TLS.
         tls:
--- a/caikit/config/config.yml
+++ b/caikit/config/config.yml
@@ -187,7 +187,8 @@ runtime:
     # Configuration for the trace push client
     trace:
         enabled: false
-        endpoint: "localhost:14250"
+        # For HTTP default, use "http(s)://localhost:4318/v1/traces"
+        endpoint: "localhost:4317"
         protocol: grpc
         service_name: caikit.runtime
         # TLS config for the otel server. If CA is falsy, insecure. If client
--- a/caikit/config/config.yml
+++ b/caikit/config/config.yml
@@ -184,6 +184,19 @@ runtime:
         enabled: true
         port: 8086
 
+    # Configuration for the trace push client
+    trace:
+        enabled: false
+        endpoint: "localhost:14250"
+        protocol: grpc
+        service_name: caikit.runtime
+        # TLS config for the otel server. If CA is falsy, insecure. If client
+        # key and cert provided with CA, mTLS, otherwise TLS.
+        tls:
+            ca: ""
+            client_key: ""
+            client_cert: ""
+
     # Whether or not to abort work on client cancellation
     use_abortable_threads: true
 
--- a/caikit/runtime/servicers/global_predict_servicer.py
+++ b/caikit/runtime/servicers/global_predict_servicer.py
@@ -135,7 +135,6 @@ class GlobalPredictServicer:
             library,
             lib_version,
         )
-        super()
 
     def Predict(
         self,
--- a/caikit/runtime/servicers/global_predict_servicer.py
+++ b/caikit/runtime/servicers/global_predict_servicer.py
@@ -27,6 +27,32 @@ from prometheus_client import Counter, Summary
 # First Party
 import alog
 
+#OTEL Tracing SDK and pkgs
+########################
+
+from opentelemetry import trace
+from opentelemetry.sdk.trace import TracerProvider
+from opentelemetry.sdk.trace.export import (
+    BatchSpanProcessor,
+    ConsoleSpanExporter,
+)
+
+from opentelemetry.exporter.otlp.proto.grpc.trace_exporter import OTLPSpanExporter
+from opentelemetry.sdk.resources import Resource
+
+provider = TracerProvider()
+otelendpoint=os.environ.get('OTEL_ENDPOINT')
+processor = BatchSpanProcessor(OTLPSpanExporter(endpoint=otelendpoint,insecure=True))
+provider.add_span_processor(processor)
+
+# Sets the global default tracer provider
+trace.set_tracer_provider(provider)
+
+# Creates a tracer from the global tracer provider
+tracer = trace.get_tracer("caikit.runtime.servicers.global_predict_service")
+
+########################
+
 # Local
 from caikit import get_config
 from caikit.core import MODEL_MANAGER, ModuleBase, TaskBase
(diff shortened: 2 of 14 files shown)


gabe-l-hart:
Instructions for launching Jaeger locally:

```sh
docker run --rm \
  -p 16686:16686 \
  -p 4317:4317 \
  -p 4318:4318 \
  jaegertracing/all-in-one:latest
```

caikit/config/config.yml
```diff
@@ -176,6 +176,22 @@ runtime:
         enabled: true
         port: 8086
 
+    # Configuration for the trace push client
+    trace:
+        enabled: false
+        # For HTTP default, use "http(s)://localhost:4318/v1/traces"
```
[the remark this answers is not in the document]
gabe-l-hart: Yep, `4317` is the default for `grpc`. `4318` is the default for HTTP

caikit/config/config.yml
```diff
@@ -176,6 +176,22 @@ runtime:
...
+        endpoint: "localhost:4317"
+        protocol: grpc
```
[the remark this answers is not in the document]
gabe-l-hart: Right. The decision of protocol is based on the collection server, thus I made the client side configurable here.

gkumbhat (commented):
Couple of questions / notes:
1. In this implementation, we are not adding a way to add response attributes, like `input_tokens` etc into the metrics yet ? Like by adding a function to task etc right? we can add those in future 
2. I don't love that there are 2 different ways of getting metrics, both giving out different kind of metrics. So a user would have to have both of those configured to get all the things. I was wondering if there is a way to combine the metrics into 1. Doesn't need to be in this PR, but we can open a story to investigate that for future?

gabe-l-hart:
>In this implementation, we are not adding a way to add response attributes, like input_tokens etc into the metrics yet ? Like by adding a function to task etc right? we can add those in future

That's correct. I have thoughts on how we can do this, but I want to defer to a follow-on PR since this one got big already.

>I don't love that there are 2 different ways of getting metrics, both giving out different kind of metrics. So a user would have to have both of those configured to get all the things. I was wondering if there is a way to combine the metrics into 1. Doesn't need to be in this PR, but we can open a story to investigate that for future?

I _totally_ agree. @hickeyma had [taken a stab](https://github.com/caikit/caikit/pull/499) at converting all metrics to otel. The only thought I have is that right now, the two sets of "things" are actually not overlapping: `metrics` and `traces`. The `metrics` are more individual measurements of interesting things whereas the `traces` are meant to provide "stack tracing" through the larger system (debuggability).

gabe-l-hart:
Ok, the trace propagation was pretty gnarly to get right and test. The unit tests don't _fully_ vet the specifics of how trace propagation happens because the `trace.get_trace_context` only actually yields a propagation context that will do anything with `inject` when it contains the right headers that are pretty hard to simulate.

I did, however, manage to simulate all of this using a dummy client script:

```py
import requests
from opentelemetry import propagate

from caikit.runtime import trace as caikit_trace

body = {
    "model_id": "foobar",
    "inputs": {
        "name": "Gabe"
    }
}

caikit_trace.configure()

tracer = caikit_trace.get_tracer("test")
with tracer.start_as_current_span("some_trace") as span:
    headers = {}
    ctx = propagate.inject(carrier=headers)
    print(headers)
    res = requests.post(
        url="http://localhost:8080/api/v1/task/sample",
        headers=headers,
        json=body,
    )
    print(res.json())

```

With this and `jaeger` running, you can see the nested span for the prediction nested inside of `"some_span"`.

**NOTE**: A raw `curl` call with the `traceparent`/`tracestate` headers will NOT work nicely because the `collector` (`jaeger`) never learns about the parent span, so it blows up when it fields the push request with the nested span and an unknown parent.

caikit/runtime/servicers/global_predict_servicer.py
```diff
@@ -265,14 +281,32 @@ def predict_model(
...
+            trace_span_name,
+            context=trace_context,
+        ) as trace_span:
+
+            # Set trace attributes available before checking anything
+            trace_span.set_attribute("calling", trace_span_name)
+            trace_span.set_attribute("model_id", model_id)
+            trace_span.set_attribute("request_name", request_name)
+            trace_span.set_attribute("task", getattr(task, "__name__", str(task)))
+            if request_id:
```
gkumbhat, line 307: at this point, we would always have `request_id` right ?  Since each of functions, http server and grpc servicer are creating it ? So shall we make request_id as required field and remove this `if` condition ?

caikit/runtime/servicers/global_predict_servicer.py
```diff
@@ -202,6 +206,16 @@ def Predict(
...
                         inference_signature,
                     )
+
+            # Get a unique ID from this request or make one up
+            if (
+                request_id := get_metadata(
+                    context, REQUEST_ID_HEADER_KEY, required=False
+                )
+            ) is None:
+                request_id = str(uuid.uuid4())
```
gkumbhat, line 216: suggestion: for cases where we are creating a local request_id, for ease of trackability, can we add a `local-` prefix to the `request_id` that way, when we are looking at the request_id, we know if this is something we got from downstream or we generated it locally

tests/runtime/conftest.py
```diff
@@ -633,3 +642,25 @@ def deploy_good_model_files():
...
+    after tests complete.
+    """
+    sys_mod_copy = sys.modules.copy()
+    # NOTE: There is a strange import error in a circular import in
+    #   opentelemetry.metrics if we mock.patch sys.modules with the copy, so
+    #   instead we let the imports work with the real sys.modules and then prune
+    #   after the test. This is less robust to parallelism, but we don't run
+    #   tests in parallel for now anyway.
+    try:
+        with mock.patch.object(trace, "_TRACE_MODULE", None):
```
gkumbhat, line 660: 🌶️

gkumbhat (commented):
Left some small suggestions, but otherwise looks good to me. Thanks for the integration and figuring out all the corners around testing these.

caikit/runtime/servicers/global_predict_servicer.py
gabe-l-hart: Nice, this seems like a good idea

gabe-l-hart: Now that I think more about it, all of this `request_id` stuff should be stripped out. This was my original thought on how to do span propagation, but was not at all aligned with how otel does it.

One of the big follow ups I'd like to do from this PR is to create an _actual_ `RuntimeServerContextType` class that does the shimming between `grpc` and `fastapi` so that we don't need to keep writing optional dependency checks and type checks. If we do this, I think this would be a fine place to reinstate this `request_id` logic so that the id becomes available to any place downstream that wants to have it for logging purposes.

As it stands, though, the _only_ use of this `request_id` is to attach it to the trace which is useless because it will always have a unique ID for the `span` anyway.

gkumbhat (approved):
Looks good to me!

PR #732, opened 2024-06-04
add debug logging around limit concurrency

swith004:
This is needed to add server logging around limit concurrency to verify proper server configurations are being set during debugging

**If applicable**:

### DIFF
--- a/caikit/runtime/http_server/http_server.py
+++ b/caikit/runtime/http_server/http_server.py
@@ -267,6 +267,13 @@ class RuntimeHTTPServer(RuntimeServerBase):
                     "Limiting HTTP server concurrency to %d",
                     concurrency_limit,
                 )
+            # limit concurrency is set to a specific value, so log the value
+            else:
+                log.info(
+                    "<RUN57106695I>",
+                    "HTTP server concurrency set to %d",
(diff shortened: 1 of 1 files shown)


caikit/runtime/http_server/http_server.py
```diff
@@ -267,6 +267,13 @@ async def validation_exception_handler(_, exc: ResponseValidationError):
                     "Limiting HTTP server concurrency to %d",
                     concurrency_limit,
                 )
+            # limit concurrency is set to a specific value, so log the value
+            else:
+                log.info(
+                    "<RUN57106695I>",
+                    "HTTP server concurrency set to %d",
+                    concurrency_limit,
+                )
```
gkumbhat, line 276: actually instead of adding logs for particularly the `concurrency_limit`, can we add `log.debug` for `server_config` dict itself, that way we get view of all the configurations

gkumbhat (changes requested):

caikit/runtime/http_server/http_server.py
swith004: Makes sense; added debug log around the server_config dict

gabe-l-hart (approved):

PR #733, opened 2024-06-05
Add .complete file, env var for termination log

kellyaa:
Adds .complete file to indicate to client who are watching the filesystem that we are done writing all of the files to disk. Also makes the termination log path configurable by environment variable.

### DIFF
--- a/caikit/runtime/train.py
+++ b/caikit/runtime/train.py
@@ -313,6 +313,8 @@ def main() -> int:
             future = train(module, wait=True, **train_kwargs)
             info = future.get_info()
             if info.status == TrainingStatus.COMPLETED:
+                # The .complete file indicates all model files have completed
+                # being written to the filesystem
                 complete_path = os.path.join(args.save_path, ".complete")
                 log.info(f"Creating completion file at: {complete_path}")
                 Path(complete_path).touch()
--- a/caikit/runtime/train.py
+++ b/caikit/runtime/train.py
@@ -94,10 +94,10 @@ def exit_complete(
     if save_path:
         try:
             complete_path = os.path.join(save_path, ".complete")
-            log.info(f"Creating completion file at: {complete_path}")
+            log.info("Creating completion file at: %s", complete_path)
             Path(complete_path).touch()
         except Exception as e:
-            log.warning(f"Unable to write completion file due to execption: {e}")
+            log.warning("Unable to write completion file due to execption: %s", e)
 
     exit(exit_code)
 
--- a/caikit/runtime/train.py
+++ b/caikit/runtime/train.py
@@ -84,6 +84,12 @@ def main() -> int:
     """Main entrypoint for running training jobs"""
     parser = TrainArgumentParser(description=__doc__)
 
+    # Set default values for termination log incase parsing the arguments fail later on
+    enable_termination_log = os.environ.get("ENABLE_TERMINATION_LOG", True)
+    termination_log_file = os.environ.get(
+        "TERMINATION_LOG_FILE", "/dev/termination-log"
+    )
+
     # Required Args
     parser.add_argument(
         "--training-kwargs",
@@ -133,17 +139,14 @@ def main() -> int:
     parser.add_argument(
         "--termination-log-file",
         "-f",
+        default=termination_log_file,
         help="Location of where to write a termination error message",
     )
     parser.add_argument(
         "--enable-termination-log",
         "-e",
+        default=enable_termination_log,
         help="Whether to enable writing to termination log when training fails",
-    )
-    # Set default values for termination log incase parsing the arguments fail later on
-    enable_termination_log = os.environ.get("ENABLE_TERMINATION_LOG", True)
-    termination_log_file = os.environ.get(
-        "TERMINATION_LOG_FILE", "/dev/termination-log"
     )
 
     try:
@@ -311,13 +314,15 @@ def main() -> int:
             args.model_name,
         ):
             future = train(module, wait=True, **train_kwargs)
+
+            # The .complete file indicates all files have completed
+            # being written to the filesystem
+            complete_path = os.path.join(args.save_path, ".complete")
+            log.info(f"Creating completion file at: {complete_path}")
+            Path(complete_path).touch()
+
             info = future.get_info()
             if info.status == TrainingStatus.COMPLETED:
-                # The .complete file indicates all model files have completed
-                # being written to the filesystem
-                complete_path = os.path.join(args.save_path, ".complete")
-                log.info(f"Creating completion file at: {complete_path}")
-                Path(complete_path).touch()
                 log.info("Training finished successfully")
                 return 0
             else:
--- a/caikit/runtime/train.py
+++ b/caikit/runtime/train.py
@@ -66,8 +66,9 @@ class TrainArgumentParser(argparse.ArgumentParser):
         raise ArgumentParserError(f"{self.prog}: error: {message}")
 
 
-def write_termination_log(text: str):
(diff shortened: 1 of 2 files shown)


caikit/runtime/train.py
```diff
@@ -65,7 +66,8 @@ def error(self, message):
         raise ArgumentParserError(f"{self.prog}: error: {message}")
 
 
-def write_termination_log(text: str, log_file="/dev/termination-log"):
+def write_termination_log(text: str):
+    log_file = os.environ.get("TERMINATION_LOG_FILE", "/dev/termination-log")
```
gabe-l-hart, line 70: I probably missed this in a previous PR, but I'd like to make the termination log stuff configurable (and optional). This should be managed via a command line arg like the rest of the configuration. I'd also support something more clever that allows the CLI args to be set via env var as well, but I'd like to avoid `os.environ` as much as possible unless it's systematic.

caikit/runtime/train.py
```diff
@@ -297,6 +299,9 @@ def main() -> int:
             future = train(module, wait=True, **train_kwargs)
             info = future.get_info()
             if info.status == TrainingStatus.COMPLETED:
+                complete_path = os.path.join(args.save_path, ".complete")
```
gabe-l-hart, line 302: It looks like the semantics here are only to have this file when the job completes successfully, right? That should be made clear since "complete" could also be interpreted as "done one way or another"

gabe-l-hart (changes requested):
One change on how we configure the termination log

caikit/runtime/train.py
kellyaa: Got you. I did it this was since this was also how we are doing it for fine tuning, so I was trying to keep the configs consistent. Open to changing it to be an argument like the other parameters.
If it can't write to the termination log, it just catches the exception and moves on. That's kind of how I made it "optional" but can make it more explicitly optional if that makes sense. Or at least turned on by default and you can disable it if you like.

caikit/runtime/train.py
kellyaa: That's correct. Made clear as in a code comment or did you have something else in mind?

caikit/runtime/train.py
gabe-l-hart: You can do the env + cli thing as follows:

```
parser.add_argument("-f", "--foo", default=os.getenv("FOO", "default-foo"), ...)
```

caikit/runtime/train.py
gabe-l-hart: I think a code comment would be needed for sure. I might suggest that we call the file `.success` instead of `.complete` since `.complete` is ambiguous.

caikit/runtime/train.py
```diff
@@ -130,15 +133,29 @@ def main() -> int:
     parser.add_argument(
         "--termination-log-file",
         "-f",
-        action="store_true",
-        default="/dev/termination-log",
         help="Location of where to write a termination error message",
     )
+    parser.add_argument(
```
gabe-l-hart, line 138: I think we can do a little better here by doing the `os.environ.get` before `parser.add_argument` and then using the value as the `default` in `parser.add_argument` above. This will allow the user to specify a value on the command line that would take precedence over the env (standard default -> env -> cli precedence)

caikit/runtime/train.py
gabe-l-hart: What do others think on `.complete` vs `.success`? I have a bit of a preference for `.success` for the clarity of the word.

gabe-l-hart (changes requested):
One more requested change on the CLI/env configs, and I'd like to resolve `.complete` vs `.success` (low Cares About from me, but we should discuss).

caikit/runtime/train.py
kellyaa: I originally put `.complete` because it is an indicator of when files are done being written to the filesystem, not necessarily an indicator of job success (that would be the container exit code). We also have fms-hf-trainer using .complete, so we'll want to keep them consist it (so either we leave it as `.complete` here, or we change it to `.success` in both places.

caikit/runtime/train.py
kellyaa: Ah yeah makes sense.

caikit/runtime/train.py
gabe-l-hart: Ok, I guess that then brings up my other question about this only being written on successful completion then. If this is meant to indicate "files are done being written," then I think we should do it on failure as well since there may still be some files (minimally, the termination log, but possibly also intermediate checkpoints and/or log files).

kellyaa: hmmmm good point.

caikit/runtime/train.py
kellyaa: So these lines effectively end up doing that (allowing the parser values to override this variable indicating the termination log variables). It's not as elegant as I'd like, but I needed to store the values in a variable that wasn't pulled directly from the parser - so in the cases where the parser threw errors, I'd still be able to access the termination log values.

kellyaa: These lines: https://github.com/caikit/caikit/blob/c7e00e9742fed5b4605713e289cffb088b3f6822/caikit/runtime/train.py#L153-L157

gabe-l-hart: Ah, makes sense!

caikit/runtime/train.py
```diff
@@ -295,6 +314,13 @@ def main() -> int:
             args.model_name,
         ):
             future = train(module, wait=True, **train_kwargs)
+
+            # The .complete file indicates all files have completed
```
gabe-l-hart, line 318: I think this should probably be done after the termination log too. You could probably write a special `exit_complete` function that replaces all the places you call `exit` and does this `.complete` file then exits. Since it's supposed to be "all done touching disk" and since the termination log may be an important artifact for the job's parent to parse, I think it should be the very last thing that mutates the disk.

gabe-l-hart (changes requested):
One more thought on when to write `.complete`

caikit/runtime/train.py
```diff
@@ -80,6 +80,28 @@ def write_termination_log(text: str, log_file: str, enabled: bool):
...
+    termination_log_file: str,
+    enable_termination_log: bool,
+):
+    if exit_code != 0:
+        write_termination_log(message, termination_log_file, enable_termination_log)
+
+    if save_path:
+        try:
+            complete_path = os.path.join(save_path, ".complete")
+            log.info(f"Creating completion file at: {complete_path}")
```
gabe-l-hart, line 97: Lazy log interpolation here

```suggestion
            log.info("Creating completion file at: %s", complete_path)
```

caikit/runtime/train.py
```diff
@@ -80,6 +80,28 @@ def write_termination_log(text: str, log_file: str, enabled: bool):
...
+            Path(complete_path).touch()
+        except Exception as e:
+            log.warning(f"Unable to write completion file due to execption: {e}")
```
gabe-l-hart, line 100: Lazy log interpolation here

```suggestion
            log.warning("Unable to write completion file due to execption: %s", e)
```

gabe-l-hart (changes requested):
Two little spots for lazy log interpolation

caikit/runtime/train.py
```diff
@@ -77,10 +80,38 @@ def write_termination_log(text: str, log_file="/dev/termination-log"):
...
```
gabe-l-hart, line 100: Lazy log interpolation

```suggestion
            log.info("Creating completion file at: %s", complete_path)
            Path(complete_path).touch()
        except Exception as e:
            log.warning("Unable to write completion file due to execption: %s", e)
```

PR #734, opened 2024-06-06
Add explicit Retry Options to Remote Runtimes

HonakerM:
This PR adds support for retrying Remote inference requests. This functionality could be achieved by messing with the `options` arg but this new system is more explicit. The current defaults retain the existing functionality

**If applicable**:

### DIFF
--- a/caikit/runtime/client/utils.py
+++ b/caikit/runtime/client/utils.py
@@ -19,6 +19,7 @@ from typing import Dict, List, Optional, Tuple
 
 # Third Party
 from requests import Session
+from requests.adapters import HTTPAdapter, Retry
 
 # Third party
 import grpc
@@ -31,8 +32,18 @@ def construct_grpc_channel(
     target: str,
     options: Optional[List[Tuple[str, str]]] = None,
     tls: Optional[ConnectionTlsInfo] = None,
+    retries: Optional[int] = None,
+    retry_options: Optional[Dict[str, str]] = None,
 ) -> grpc.Channel:
     """Helper function to construct a grpc Channel with the given TLS config"""
+    # Add retry option if one was provided
+    if retries and "grpc.enable_retries" not in options:
+        options["grpc.enable_retries"] = retries
+    # if retry_options:
+    #     for option_name, option_value in retry_options.items():
+    #         if option_name not in options:
+    #             options[option_name] = option_value
+        
     if tls and tls.enabled:
         grpc_credentials = grpc.ssl_channel_credentials(
             root_certificates=tls.ca_data,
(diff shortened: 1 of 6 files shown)


HonakerM:
@gabe-l-hart giving you a slight nudge on this one. We're starting to have some transient issues in our project so this retry logic would help

caikit/runtime/client/utils.py
```diff
@@ -31,8 +33,45 @@ def construct_grpc_channel(
...
+                break
+
+        if not options_contain_service_config:
+            service_config = {
+                "methodConfig": [
+                    {
+                        "name": [{}],
+                        "retryPolicy": {
+                            "maxAttempts": retries,
+                            "initialBackoff": "0.1s",
```
gabe-l-hart, line 59: It looks like this is hard-coded here, but overridable if the user provides these advanced configs in `options`, right?

caikit/runtime/client/utils.py
```diff
@@ -79,4 +120,16 @@ def construct_requests_session(
     if timeout:
         session.params["timeout"] = timeout
 
+    # Mount retry object if options were provided
+    if retries:
+        default_status_codes = list(Retry.RETRY_AFTER_STATUS_CODES) + [500, 502, 504]
```
gabe-l-hart, line 125: Probably not for this PR, but I do see that the builtin `http` package holds a `http.HTTPStatus` enum that we should probably start using everywhere.

https://docs.python.org/3/library/http.html#http.HTTPStatus

tests/fixtures/sample_lib/modules/sample_task/sample_implementation.py
```diff
@@ -50,6 +52,8 @@ def run(
         sample_input: SampleInputType,
         throw: bool = False,
         error: Optional[str] = None,
+        request_id: Optional[str] = None,
```
gabe-l-hart, line 55: NIT: This is a test, so it's fine, but we should probably add these (and the others) to the docstring

caikit/runtime/client/utils.py
```diff
@@ -31,8 +33,45 @@ def construct_grpc_channel(
     target: str,
     options: Optional[List[Tuple[str, str]]] = None,
     tls: Optional[ConnectionTlsInfo] = None,
+    retries: Optional[int] = None,
```
gabe-l-hart, line 36: We don't have the args to this function in the `docstring`, but I think it might be good to add them since things like `options` and `retry_options` require specific formats that could use citations/docs.

gabe-l-hart (approved):
A couple of little NITs/thoughts and places where `docstrings` could be updated, but it's also good to merge any time based on the code and tests. Let me know if you have time to handle the NITs and if not I can merge as is.

caikit/runtime/client/utils.py
HonakerM: Yep exactly! Anything the user provides in `retry_options` will replace any of these static keys. Including `maxAttempts` if they supply it

caikit/runtime/client/utils.py
HonakerM: Oh I didn't realize that was in the standard library! It would be neat to refactor the code to use them

HonakerM:
@gabe-l-hart I updated the doc strings!

gabe-l-hart (approved):
Ship it!

PR #735, opened 2024-06-18
Module backend context update

gabe-l-hart:

The [caikit.core.module_backends](https://github.com/caikit/caikit/blob/main/docs/adrs/018-shared-backends.md) are designed to allow individual models to be executed in multiple ways, including proxying to remote services (e.g. [caikit-tgis-backend](https://github.com/caikit/caikit-tgis-backend)).

When a backend performs remote network operations, it may need to be aware of meta-content in runtime request headers (e.g. [x-route-info](https://github.com/caikit/caikit-nlp/pull/363)). Furthermore, a backend may be used during lazy model discovery (e.g. [TGISAutoFinder](https://github.com/caikit/caikit-nlp/blob/main/caikit_nlp/model_management/tgis_auto_finder.py)).

This combination of backend functionality means that the backend needs to be made aware of the server context _before_ attempting to initialize the model. To accomplish this, this PR adds the following:

* Add `handle_runtime_context` to `caikit.core.module_backends.base.BackendBase` which can be overridden by derived backends
* Add `caikit.core.MODEL_MANAGER.get_module_backends` which exposes an easy way to get at the configured module backends
* In `GlobalPredictServicer` add `notify_backends_with_context` to call `handle_runtime_context` on all configured backends found using `MODEL_MANAGER.get_module_backends()`
* Call `notify_backends_with_context` before calls to `retrieve_model` in both `GlobalPredictServicer.Predict` (the central `grpc` request handler) and the `http_server` unary and stream handlers.

This PR does NOT fix the situation with `TGISAutoFinder` + `TGISBackend` yet. To support lazy auto finding for TGIS models with `x-route-info`, we will need to implement `TGISBackend.handle_runtime_context` to perform similar logic to the handling of `x-route-info` in `caikit-nlp`. This will be a change of responsibility with some strange version compatibility implications (do we remove the `x-route-info` parsing from `caikit-nlp`?).

**If applicable**:

### DIFF
--- a/caikit/core/module_backends/base.py
+++ b/caikit/core/module_backends/base.py
@@ -24,6 +24,9 @@ import abc
 # First Party
 import aconfig
 
+# Local
+from ..data_model.runtime_context import RuntimeServerContextType
+
 
 class BackendBase(abc.ABC):
     """Interface for creating configuration setup for backends"""
@@ -66,3 +69,21 @@ class BackendBase(abc.ABC):
     def start_lock(self):
         with self._start_lock:
             yield
+
+    def handle_runtime_context(
+        self,
+        model_id: str,
+        runtime_context: RuntimeServerContextType,
+    ):
+        """Update backend state for the given model based on a runtime request.
+
+        Some backends may need to handle runtime context information for the
+        target model in order to correctly configure the backend before finding
+        and loading the model. By default, this is a No-Op.
+
+        Args:
+            model_id (str): The unique ID of the model that is referenced by the
+                runtime context
+            runtime_context (RuntimeServerContextType): The context for the
+                given runtime request
+        """
--- a/caikit/core/module_backends/base.py
+++ b/caikit/core/module_backends/base.py
@@ -70,7 +70,7 @@ class BackendBase(abc.ABC):
         with self._start_lock:
             yield
 
-    def handle_runtime_context(
+    def handle_runtime_context(  # noqa: B027
         self,
         model_id: str,
         runtime_context: RuntimeServerContextType,
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -4,7 +4,7 @@ This sets up global test configs when pytest starts
 
 # Standard
 from contextlib import contextmanager
-from typing import Callable, List, Union
+from typing import Callable, List, Optional, Union
 from unittest.mock import patch
 import copy
 import importlib
@@ -19,6 +19,7 @@ import pytest
 import semver
 
 # First Party
+import aconfig
 import alog
 
 # Local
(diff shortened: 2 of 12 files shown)


caikit/core/module_backends/base.py
```diff
@@ -66,3 +69,21 @@ def stop(self):
     def start_lock(self):
         with self._start_lock:
             yield
+
+    def handle_runtime_context(  # noqa: B027
+        self,
+        model_id: str,
+        runtime_context: RuntimeServerContextType,
```
joerunde, line 76: Looks like there's an assumption baked in here that the information in any given context will be the same for the same model id. If multiple concurrent requests come in for the same model this won't be able to disambiguate, right?

Seems fine for the use case where a given model just needs some extra metadata to be loaded and it'll be the same on every request, but not so fine if somebody tries to use this to add tracing info to inference calls

gabe-l-hart: Ah, yeah, that is definitely true. The `context_arg` stuff that @HonakerM added is the right way to do per-call tracing (WIP at the moment).

tests/conftest.py
```diff
@@ -209,6 +210,33 @@ def temp_config(config_overrides: dict, merge_strategy="override"):
             yield get_config()
 
 
+def get_mutable_config_copy(base_config: Optional[aconfig.ImmutableConfig] = None):
```
joerunde, line 213: ooof :D

joerunde:
The tangles of the web thicken!

This looks reasonably simple and fine, I'm just sad that there's not a good tool in python (at least not that I know of) that can make this a bit more transparent.

A probably silly suggestion but if we're doing something that already assumes 1 current context per model id, what about simplifying this to store those in memory with caikit?

```
# in runtime server
def Predict(request, context):
  model_id = ...
  caikit.set_request_context(model_id, context)
  ...

# in backends
def some_fn_that_can_use_context(model_id):
  context = caikit.get_request_context(model_id)
  if context:
    # do something with it
```

tests/conftest.py
gabe-l-hart: yerp. This should really be added to `aconfig`

gkumbhat (approved):
Thanks Gabe. The changes LGTM.

gabe-l-hart:
(Sidebar: I _hate_ that you can't respond to the top-level comment on a PR!)

@joerunde to your point, yes I 100% agree that this is gross. I did think about exactly what you're suggesting with a per-thread context map. It would certainly work and could accomplish the same thing, but I've always been worried about relying on the thread ID matching up inside a plugin. We don't currently support using `multiprocessing` to handle requests or doing process-level isolation for inference, but those wouldn't necessarily be out of the realm of possibility and would badly break the idea of a thread-local `get_current_context()` function.

The `context_arg` change and this change make it explicit where the request can be accessed and guarantee that it will be consistent (at the cost of more complex plumbing).

caikit/core/module_backends/base.py
gabe-l-hart: >If multiple concurrent requests come in for the same model this won't be able to disambiguate, right?

Yes, this would definitely fall to the backend itself to handle

PR #744, opened 2024-06-28
:sparkles: Add initial metadata

evaline-ju:

For: https://github.com/caikit/caikit/issues/745

**If applicable**:


caikit/runtime/interceptors/caikit_runtime_server_wrapper.py
```diff
@@ -37,6 +37,8 @@
     ["rpc_name"],
 )
 
+ACK_STRING = "acknowledgement"
```
gabe-l-hart, line 40: I think this should be in `caikit.runtime.names` to consolidate constants

gabe-l-hart (changes requested):
One small request for relocating the constant. It would also be great to get at least one unit test on this!

gabe-l-hart (approved):
Merging as is with unit tests coming as a follow up

PR #751, opened 2024-07-25
Implement Prediction Job

HonakerM:
This PR refactors the model trainer abstraction to be a more generic "job" abstraction. With this job abstraction we are able to implement both inferences and trainings in the background and poll their status/results. 

This type of functionality has multiple use-cases inside Caikit with a few listed below:
- Executing a long running inference which could be impacted by network issues 
- Tracking and managing inferences while they're running e.g. polling the status of an inference or cancelling a request.
- Better support coordination between pods to support things like load distribution 

This PR creates new http and grpc endpoints for the job predictions:

HTTP
```
POST /api/v1/task/<task>/job - create
GET /api/v1/task/<task>/job - get status
GET /api/v1/task/<task>/job/results - get results
DELETE /api/v1/task/<task>/job - cancel
```

GRPC:
```
caikit.runtime.SampleLib.SampleLibJobService.<task>StartPredictionJob - create job
caikit.runtime.SampleLib.SampleLibJobService.<task>PredictionJobStatus - get status
caikit.runtime.SampleLib.SampleLibJobService.<task>GetPredictionJobResult - get results
caikit.runtime.SampleLib.SampleLibJobService.<task>CancelPredictionJob - cancel
```

**If applicable**:


caikit/interfaces/runtime/data_model/package.py
```diff
@@ -19,3 +19,4 @@
 from caikit.core.data_model import CAIKIT_DATA_MODEL
 
 RUNTIME_PACKAGE = f"{CAIKIT_DATA_MODEL}.runtime"
+RUNTIME_JOB_PACKAGE = f"{RUNTIME_PACKAGE}.job"
```
HonakerM, line 22: This was required due to a [py-to-proto](https://github.com/IBM/py-to-proto/issues/66) bug with protobuf 3

caikit/core/data_model/job.py
```diff
@@ -0,0 +1,49 @@
...
+            self.__class__.COMPLETED,
+            self.__class__.CANCELED,
+            self.__class__.ERRORED,
+        ]
+
+
+class JobType(Enum):
+    """Enum to track the different types of Jobs"""
+
+    TRAINING = (1,)
```
tozhangkai, line 48: @HonakerM why `TRAINING` is a tuple but `PREDICTION` is an integer?

gabe-l-hart:
One thought on naming:

> GRPC:
> 
> caikit.runtime.SampleLib.SampleLibJobService.<task>StartPredictionJob - create job
> caikit.runtime.SampleLib.SampleLibJobService.<task>PredictionJobStatus - get status
> caikit.runtime.SampleLib.SampleLibJobService.<task>GetPredictionJobResult - get results
> caikit.runtime.SampleLib.SampleLibJobService.<task>CancelPredictionJob - cancel

I think we should be consistent about whether the RPCs are prefixed with the logical verb. Here, `PredicitonJobStatus` is the odd-one-out I think, so I'd suggest we do `GetPredictionJobStatus`.

HonakerM:
@gabe-l-hart good call on naming! That definitely makes more sense. Let me push a commit.

caikit/core/model_management/job_predictor_base.py
```diff
@@ -0,0 +1,64 @@
...
+
+class JobPredictorInfo(JobInfo):
+    """JobPredictorInfo isa remap of JobInfo but for predictors"""
+
+    pass
+
+
+class JobPredictorBase(JobBase):
+    __doc__ = __doc__
+    ModelFutureBase = JobFutureBase
```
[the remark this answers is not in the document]
gabe-l-hart: I see in the implementation, the `result` is set up to return `DataObjectBase`. That would be another nice thing to specify in an inherited subclass.

caikit/core/model_management/local_job_predictor.py
```diff
@@ -0,0 +1,221 @@
...
+    CaikitCoreException,
+    CaikitCoreStatusCode,
+)
+
+log = alog.use_channel("LOC-TRNR")
+error = error_handler.get(log)
+
+# Constant for the result file name. This is exposed here to better support
+# subclassing
+RESULT_FILE_NAME = "result.bin"
```
gabe-l-hart, line 55: Is there a reason this isn't configurable? Seems like a user may want to be able to avoid overwriting and/or name things to fit their needs.

caikit/core/model_management/local_job_predictor.py
```diff
@@ -0,0 +1,221 @@
...
+            subclasses implement their own save method like S3"""
+            save_path_pathlib = Path(self.save_path)
+            log.debug("Saving inference %s to %s", self.id, self.save_path)
+            save_path_pathlib.parent.mkdir(exist_ok=True, parents=True)
+            with alog.ContextTimer(
+                log.debug, "Inference %s saved in: ", self.id
+            ) and save_path_pathlib.open("wb") as output_file:
+                output_file.write(result.to_binary_buffer())
+
+    LocalModelFuture = LocalJobPredictorFuture
```
gabe-l-hart, line 138: This seems misplaced here? I think `LocalModelFuture` should only be in the trainer?

caikit/core/model_manager.py
```diff
@@ -180,28 +189,96 @@ def train(
...
+                model_future.wait()
+
+        # Return a handle to the future
+        return model_future
+
     def get_model_future(
         self,
-        training_id: str,
-    ) -> ModelTrainerBase.ModelFutureBase:
+        future_id: str = None,
```
gabe-l-hart, line 244: I think that the support for fetching prediction jobs should be a separate function since `ModelFuture` is strictly for training

caikit/interfaces/runtime/data_model/package.py
gabe-l-hart: Yuck! I wonder if we should just fix that and bump the `py-to-proto` dependency here? That might be less painful than the inevitable breaking change when we want to consolidate the package naming.

caikit/interfaces/runtime/data_model/prediction_job_management.py
```diff
@@ -0,0 +1,49 @@
...
+# Local
+from ....core.data_model import DataObjectBase, JobStatus, dataobject
+from .package import RUNTIME_JOB_PACKAGE
+
+
+@dataobject(RUNTIME_JOB_PACKAGE)
+class PredictionJobInfoRequest(DataObjectBase):
+    """DataModel to request information about a PredictionJob"""
+
+    job_id: str
```
gabe-l-hart, line 31: Let's add `FieldNumber`s on all of these even though it's a single field just for the sake of consistency

caikit/interfaces/runtime/data_model/training_management.py
```diff
@@ -43,7 +43,7 @@ class ModelPointer(DataObjectBase):
 @dataobject(RUNTIME_PACKAGE)
 class TrainingStatusResponse(DataObjectBase):
     training_id: Annotated[str, FieldNumber(1)]
-    state: Annotated[TrainingStatus, FieldNumber(2)]
+    state: Annotated[JobStatus, FieldNumber(2)]
```
gabe-l-hart, line 46: This will revert if we do the inheritance thing which I think would be good. I think this is _technically_ an API breaking change as is since the name of the message is changing.

caikit/runtime/grpc_server.py
```diff
@@ -103,6 +114,31 @@ def __init__(
                 self.inference_service.service, self.server
             )
 
+        if self.enable_inference_jobs:
```
gabe-l-hart, line 117: NIT: Should this be `enable_prediction_jobs`?

caikit/runtime/grpc_server.py
```diff
@@ -103,6 +114,31 @@ def __init__(
...
+            # Initialize the job management servicer and create a function wrapper
+            self._prediction_job_management_servicer = (
+                PredictionJobManagementServicerImpl()
+            )
+            self.server = CaikitRuntimeServerMultiFuncWrapper(
```
gabe-l-hart, line 122: I haven't made it far enough yet, but I'm not understanding the need for this `MultiFuncWrapper` stuff yet...

caikit/runtime/grpc_server.py
gabe-l-hart: I guess we say `enable_inference`, so it's probably fine as-is

caikit/runtime/http_server/http_server.py
```diff
@@ -1002,7 +1188,7 @@ def _get_response_openapi(
             for def_name, schema in json_schema.pop("$defs", {}).items():
                 self._openapi_defs[def_name] = schema
 
-            response_schema = {"application/json": json_schema}
+            response_schema = {"application/json": {"schema": json_schema}}
```
gabe-l-hart, line 1191: Interesting. What was this line about?

caikit/runtime/interceptors/caikit_runtime_server_wrapper.py
```diff
@@ -379,3 +384,63 @@ def _get_handler_fn(handler: RpcMethodHandler) -> Callable:
         if handler.stream_unary:
             return handler.stream_unary
         return handler.stream_stream
+
+
+class CaikitRuntimeServerMultiFuncWrapper(CaikitRuntimeServerWrapper):
+    """This class has the same functionality of the CaikitRuntimeServerWrapper
```
gabe-l-hart, line 390: Why not extend the existing `CaikitRuntimeServerWrapper` to allow this?

gabe-l-hart (changes requested):
I'm part way through with some NITs and a few rearranging suggestions. I haven't made it through the full `runtime` plumbing, but I'm going to submit now so the `core` things can be looked at first.

caikit/core/data_model/training_status.py
```diff
@@ -14,27 +14,8 @@
...
 
-    @property
-    def is_terminal(self):
-        return self in [
-            self.__class__.COMPLETED,
-            self.__class__.CANCELED,
-            self.__class__.ERRORED,
-        ]
+# Copy JobStatus as training status for now
+TrainingStatus = JobStatus
```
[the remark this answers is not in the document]
HonakerM: Oh I didn't realize that! Good call. I'll update

caikit/interfaces/runtime/data_model/package.py
HonakerM: So I initially looked to fix `py-to-proto` but I had no idea where it lived 😅 . I'll look into that once I resolve the other comments

caikit/core/data_model/predict_status.py
```diff
@@ -0,0 +1,21 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+Common data model enum used for reporting training status
+"""
+# Local
+from .job import JobStatus
+
+# Copy JobStatus as prediction status for now
+PredictJobStatus = JobStatus
```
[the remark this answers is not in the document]
HonakerM: Yeah that probably makes more sense!

caikit/core/model_management/job_predictor_base.py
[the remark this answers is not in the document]
HonakerM: Good call, that is nice to specify in the base class. I restructured the file to mimic the model trainer class

caikit/core/model_management/local_job_predictor.py
HonakerM: So this is just the sub file name inside a unque folder for that id e.g. `<result_dir>/<future id>/result.bin`. Initially I had `result.bin` hard coded in the `__init__` but I moved it to a constant for better subclassing. I don't think it needs to be user configurable since the `result_dir` already is but will defer to your judgement. This is a low CA for me

caikit/core/model_management/local_job_predictor.py
HonakerM: Yep! Darn copy/pasta

caikit/core/data_model/training_status.py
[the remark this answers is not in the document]
HonakerM: So it turns out Enum's specifically aren't subclassable[ in python](https://stackoverflow.com/questions/33679930/how-to-extend-python-enum). Let me see if I just use `DataObjectBase` that would work better.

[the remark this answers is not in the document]
HonakerM: DataObjectBase didn't work as well since it breaks the fields. I'll revert my change and will swing back around to this one.

caikit/interfaces/runtime/data_model/training_management.py
HonakerM: So inheritance thing is a no go until we fix the py-to-proto bug (I'll look into it later today). I'll leave it for now until that's fixed.

caikit/runtime/grpc_server.py
HonakerM: yeah that was a big pain point for me. I think enable_prediction_jobs is the correct answer but since we already use enable_inference I thought we should retain consistency

caikit/runtime/http_server/http_server.py
HonakerM: Oh this is a change that should probably be its own PR but I snuck it in here..... all of our openapi generation is broken rn since it has the incorrect schema format 😅

caikit/runtime/interceptors/caikit_runtime_server_wrapper.py
HonakerM: Uhhh good question, I didn't want to break anything 😅. It would probably be a pretty straightforward change so let me refactor

caikit/core/data_model/training_status.py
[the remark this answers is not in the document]
HonakerM: Shucks, this might be a limitation of protobuf: https://stackoverflow.com/questions/68134072/different-enums-same-value-in-protobuf. Any ideas?

caikit/runtime/http_server/http_server.py
```diff
@@ -679,6 +749,122 @@ async def _handler(
...
+                    model_id=model_id,
+                    request_name=rpc.request.name,
+                    task=rpc.task,
+                    context=context,
+                    **request_params,
+                )
+                result = await loop.run_in_executor(self.thread_pool, call)
+                log.debug4("Job started from model %s with id", model_id, result.job_id)
+
+                return Response(content=result.to_json(), media_type="application/json")
```
tozhangkai, line 858: @HonakerM it seems the async job submission returns a regular 200 success response. By convention, async API should return 202 accepted. In most use cases it doesn't make much difference but I know some applications have different behaviors on 202 response. (WxO for example.) Is it possible to 202?

HonakerM: @tozhangkai good idea, 202 definitely makes more sense logically. I just pushed a commit to update the prediction job and training job to return 202s. 

@gabe-l-hart what are your thoughts on this? Updating the train endpoint to 202 will be a breaking change but it does make more sense logically

caikit/interfaces/runtime/data_model/training_management.py
gabe-l-hart: Ok, yeah, I do want to be careful that we don't break the API by accident here

caikit/runtime/grpc_server.py
gabe-l-hart: Yeah, ug, sorry for the inconsistency!

caikit/runtime/http_server/http_server.py
gabe-l-hart: Deal. You _might_ want to split this into its own PR so it can be merged quickly

caikit/runtime/http_server/http_server.py
```diff
@@ -442,6 +457,56 @@ def _cancel_training(self, training_id: Annotated[str, Query]) -> Response:
                 )
             raise
 
+    def _get_prediction_job_result(self, job_id: Annotated[str, Query]) -> Response:
+        """GET handler for fetching a training"""
```
gabe-l-hart, line 461: NIT:
```suggestion
        """GET handler for fetching a prediction job result"""
```

caikit/runtime/http_server/http_server.py
```diff
@@ -442,6 +457,56 @@ def _cancel_training(self, training_id: Annotated[str, Query]) -> Response:
...
+        except Exception as err:
+            if error_content := self._handle_exception(err):
+                return Response(
+                    content=json.dumps(error_content),
+                    status_code=error_content["code"],
+                )
+            raise
+
+    def _get_prediction_job_status(self, job_id: Annotated[str, Query]) -> Response:
+        """GET handler for fetching a training"""
```
gabe-l-hart, line 479:
```suggestion
        """GET handler for fetching a prediction job status"""
```

caikit/runtime/http_server/http_server.py
```diff
@@ -442,6 +457,56 @@ def _cancel_training(self, training_id: Annotated[str, Query]) -> Response:
...
+        except Exception as err:
+            if error_content := self._handle_exception(err):
+                return Response(
+                    content=json.dumps(error_content),
+                    status_code=error_content["code"],
+                )
+            raise
+
+    def _cancel_prediction_job(self, job_id: Annotated[str, Query]) -> Response:
+        """DELETE handler for undeploying a model"""
```
gabe-l-hart, line 495:
```suggestion
        """DELETE handler for canceling a prediction job"""
```

caikit/runtime/http_server/http_server.py
```diff
@@ -558,9 +628,9 @@ def _train_add_unary_input_unary_output_handler(self, rpc: ModuleClassTrainRPC):
         pydantic_response = dataobject_to_pydantic(response_data_object)
 
         @self.app.post(
-            get_http_route_name(rpc.name),
+            get_http_route_name(rpc_name=rpc.name),
             responses=self._get_response_openapi(
-                response_data_object, pydantic_response
+                response_data_object, pydantic_response, 202
```
gabe-l-hart, line 633: I think we've discussed usage of 202 vs 200 in the past. You're right that 202 is the right code for creating a job that is accepted by the server. On the flip side, I worry that using 202 could break `if response.code != 200` style checks on the client side (which are NOT correct, but almost certainly exist in places).

gabe-l-hart: Furthermore, I think this would be an API breaking change, so I think we should not move to 202 at this point.

caikit/runtime/http_server/http_server.py
```diff
@@ -679,6 +753,126 @@ async def _handler(
...
+        self.app.get(
+            get_http_prediction_job_route_name(rpc_name=rpc.name),
+            responses=self._get_response_openapi(
+                get_dataobject_response, get_pydantic_response
+            ),
+            response_class=Response,
+            include_in_schema=rpc.task.get_visibility(),
+        )(self._get_prediction_job_status)
+
+        # Bind DELETE to cancel a training
```
gabe-l-hart, line 772:
```suggestion
        # Bind DELETE to cancel a prediction job
```

caikit/runtime/http_server/http_server.py
```diff
@@ -679,6 +753,126 @@ async def _handler(
...
+        self.app.delete(
+            get_http_prediction_job_route_name(rpc_name=rpc.name),
+            responses=self._get_response_openapi(
+                cancel_dataobject_response, cancel_pydantic_response
+            ),
+            response_class=Response,
+            include_in_schema=rpc.task.get_visibility(),
+        )(self._cancel_prediction_job)
+
+        # Bind GET at the `/result` subpath to get the result of a training
```
gabe-l-hart, line 785:
```suggestion
        # Bind GET at the `/result` subpath to get the result of a prediction job
```

caikit/runtime/http_server/http_server.py
```diff
@@ -679,6 +753,126 @@ async def _handler(
...
+                    context=context,
+                    **request_params,
+                )
+                result = await loop.run_in_executor(self.thread_pool, call)
+                log.debug4("Job started from model %s with id", model_id, result.job_id)
+
+                return Response(
+                    content=result.to_json(),
+                    media_type="application/json",
+                    status_code=status.HTTP_202_ACCEPTED,
```
gabe-l-hart, line 865: Same point here about 202. This isn't an API breaking change since it's a new endpoint, but we would want consistency.

caikit/runtime/interceptors/caikit_runtime_server_wrapper.py
```diff
@@ -47,18 +52,26 @@ class CaikitRuntimeServerWrapper(grpc.Server):
...
+        service_type: ServiceType,
+    ):
         """Initialize a new CaikitRuntimeServerWrapper
 
         Args:
             server(grpc.Server): The server that is being wrapped
-            global_predict(function): A function that will accept an arbitrary
-                gRPC request message and a grpc.ServicerContext, and return
-                a suitable gRPC response message
+            rpc_callable(Union[Callable, Dict[CaikitRPCBase, Callable]]): Either a function that
```
gabe-l-hart, line 66: Nice, this is way cleaner!

caikit/runtime/servicers/global_predict_servicer.py
```diff
@@ -229,6 +234,179 @@ def Predict(
...
+                task=caikit_rpc.task,
+                prediction_func_name=inference_signature.method_name,
+                context=context,
+                context_arg=inference_signature.context_arg,
+                model=model,
+                wait=False,
+                **caikit_library_request,
+            )
+
+            # Don't measure
```
gabe-l-hart, line 319: I'm not sure what this comment means here?

caikit/runtime/servicers/global_predict_servicer.py
```diff
@@ -229,6 +234,179 @@ def Predict(
...
+            response_proto = build_proto_response(response)
+            return response_proto
+
+    def run_prediction_job(
```
gabe-l-hart, line 323: This function looks like it has a lot of duplicate code with `run_training_job`. Is there a refactor that could slim this down?

caikit/runtime/servicers/prediction_job_management_servicer.py
```diff
@@ -0,0 +1,214 @@
...
+        """Cancel a prediction job."""
+        return self.cancel_job(request.job_id).to_proto()
+
+    ####################################
+    ## Interface-agnostic entrypoints ##
+    ####################################
+
+    def get_job_result(self, job_id: str) -> DataObjectBase:
+        """Get the result of a job by ID"""
+        model_future: JobFutureBase | None = self._get_model_future(
```
gabe-l-hart, line 93: I don't think we can use the `| None` syntax yet without dropping `3.9`

caikit/runtime/servicers/prediction_job_management_servicer.py
```diff
@@ -0,0 +1,214 @@
...
+
+            return PredictionJobStatusResponse(
+                job_id=model_future.id,
+                state=job_info.status,
+                reasons=reasons,
+            )
+        except CaikitCoreException as err:
+            # In the case that we get a `NOT_FOUND`, we assume that the job was canceled.
+            # This is to handle stateful trainers that implement `cancel` by fully deleting
+            # the training.
```
gabe-l-hart, line 170:
```suggestion
            # This is to handle stateful predictors that implement `cancel` by fully deleting
            # the prediction job.
```

caikit/runtime/servicers/prediction_job_management_servicer.py
```diff
@@ -0,0 +1,214 @@
...
+                    job_id,
+                ),
+            ) from err
+
+    ############################
+    ## Implementation Details ##
+    ############################
+
+    @staticmethod
+    def _get_model_future(job_id: str, operation: str):
```
gabe-l-hart, line 195: Should this be `_get_prediction_future`?

caikit/runtime/servicers/training_management_servicer.py
```diff
@@ -131,7 +131,7 @@ def _get_model_future(training_id: str, operation: str):
         Wrapped here so that we only catch errors directly in the `trainer.get_model_future` call
         """
         try:
-            return MODEL_MANAGER.get_model_future(training_id=training_id)
+            return MODEL_MANAGER.get_model_future(training_id)
```
gabe-l-hart, line 134: Why remove the keyword arg here? I think this may have been residual from them being merged into `get_model_future`, but I'd rather keep the explicit kwarg here

caikit/runtime/utils/servicer_util.py
```diff
@@ -292,16 +292,15 @@ def validate_data_model(
                     )
                     continue
 
-                field_message_type = input_proto_msg.fields_by_name[
-                    field.name
-                ].message_type
+                field_descriptor = input_proto_msg.fields_by_name[field.name]
```
gabe-l-hart, line 295: I think this same better check could be done without changing `field_message_type` and simply adding `field_message_type` to the first clause of the `if` (slightly smaller delta)

tests/core/helpers.py
```diff
@@ -130,15 +135,19 @@ class TestModelFuture(ModelTrainerBase.ModelFutureBase):
 
         def __init__(self, parent, trained_model, save_path, save_with_id):
             super().__init__(
-                trainer_name=parent.instance_name,
-                training_id=str(uuid.uuid4()),
-                save_path=save_path,
+                future_name=parent.instance_name,
+                future_id=str(uuid.uuid4()),
```
gabe-l-hart, line 139: I think this should be reverted since we don't share future types. You could yank this class into a generic `JobFuture` and then subclass with the naming override?

caikit/runtime/http_server/http_server.py
gabe-l-hart, line 865: Same point about 202

caikit/runtime/servicers/training_management_servicer.py
gabe-l-hart, line 134: Why remove this kwarg? I think this may be residual from the earlier code where they were merged, but probably better to revert this now that they're split.

caikit/runtime/utils/servicer_util.py
gabe-l-hart, line 295: I think this can be done with a smaller delta by just adding `if field_message_type` to the beginning of the `if`

tests/core/helpers.py
gabe-l-hart, line 139: This is interesting. I hadn't noticed that switching to have `ModelFutureBase` inherit from `JobFutureBase` caused this kwarg naming change. I think this could _technically_ be considered an API breaking change since authors of `Trainers` may have explicitly called `training_id=` when initializing the parent. I think the right way to handle this is to add name aliases in `__init__` in the subclasses of `JobFutureBase`.

tests/core/helpers.py
```diff
@@ -180,6 +189,68 @@ def get_model_future(self, training_id: str) -> ModelTrainerBase.ModelFutureBase
...
+        self.instance_name = instance_name
+        self.canned_status = config.get("canned_status", PredictionJobStatus.RUNNING)
+        self._futures = {}
+
+    class TestJobFuture(JobPredictorFutureBase):
+        __test__ = False
+
+        def __init__(self, parent, result):
+            super().__init__(
+                future_name=parent.instance_name,
```
gabe-l-hart, line 207: If we do the name aliases, this will also need to change

tests/core/model_management/test_local_job_predictor.py
```diff
@@ -0,0 +1,304 @@
...
+
+
+## Tests #######################################################################
+
+
+def test_predict_and_get_info(predictor_type_cfg):
+    """Test that running a prediction can fetch status correctly"""
+    predictor = local_predictor(**predictor_type_cfg)
+
+    # Launch the training and force it to wait
```
gabe-l-hart, line 74:
```suggestion
    # Launch the prediction and force it to wait
```

gabe-l-hart: I'm guessing we're going to find a lot of these, so consider this a blanket suggestion to check for copy-paste comments!

tests/core/model_management/test_local_job_predictor.py
```diff
@@ -0,0 +1,304 @@
...
+
+    model_future = predictor.predict(
+        SampleModule(),
+        "run",
+        sample_input=SampleInputType("hello"),
+        throw=True,
+        error="Error",
+    )
+    # assert model_future.get_info().status == TrainingStatus.RUNNING
+    # assert not model_future.get_info().status.is_terminal
```
gabe-l-hart, line 244: These should probably be uncommented or removed

tests/core/model_management/test_local_job_predictor.py
```diff
@@ -0,0 +1,304 @@
...
+
+def test_predict_and_get_info(predictor_type_cfg):
+    """Test that running a prediction can fetch status correctly"""
+    predictor = local_predictor(**predictor_type_cfg)
+
+    # Launch the training and force it to wait
+    # NOTE: Data stream passed by positional arg to ensure it is passed through
+    #   correctly by position
+    wait_event = threading.Event()
+    model_future = predictor.predict(
```
gabe-l-hart, line 78: Another copy-paste NIT, but I'd like to not call these `model_future` to avoid confusion.

tests/core/model_management/test_local_job_predictor.py
```diff
@@ -0,0 +1,304 @@
...
+        model_future.result()
+
+    # Second time should succeed
+    model_future = predictor.predict(
+        SampleModule(),
+        "run",
+        external_prediction_id=predictor_id,
+        sample_input=SampleInputType("hello"),
+    )
+    assert model_future.result()
```
gabe-l-hart, line 279: Since the `result` of the prediction job is a bit more important content-wise than the training job, I think we also need at least one test that makes sure the `result()` is the correct proto object (that might exist in the server tests below that I haven't gotten to yet)

tests/core/test_model_manager.py
```diff
@@ -727,3 +739,36 @@ def test_train_with_save_path(reset_globals):
             train_future.wait()
             assert train_future.get_info().status == TrainingStatus.COMPLETED
             assert os.path.exists(save_path)
+
+
+def test_job_predictor_with_model(reset_globals):
+    """Make sure training can be accessed through the central train function
```
gabe-l-hart, line 745: Looks like some more copy-pasta comments!

tests/fixtures/sample_lib/modules/sample_task/sample_implementation.py
```diff
@@ -68,19 +70,27 @@ def run(
...
                 for before being successful. Defaults to None.
+            sleep_time float: How long to sleep before returning a result. Defaults to 0.
+            sleep_increment float: How large of increments to sleep in.
             context (Optional[RuntimeServerContextType]): The context for the runtime server request
         Returns:
             SampleOutputType: The output
         """
         span_name = f"{__name__}.{type(self).__name__}.run"
         with trace.start_child_span(context, span_name):
+            if sleep_time:
```
gabe-l-hart, line 81: I know @joerunde will not be pleased to introduce additional sleeps in unit tests. Is it possible to get the same behavior using an `Event` instead?

tests/runtime/conftest.py
```diff
@@ -100,6 +100,19 @@ def sample_inference_service(render_protos) -> ServicePackage:
     return inference_service
 
 
+@pytest.fixture(scope="session")
+def sample_job_inference_service(render_protos) -> ServicePackage:
```
gabe-l-hart, line 104: NIT: Should this be `sample_inferece_job_service`?

tests/runtime/http_server/test_http_server.py
```diff
@@ -198,47 +200,59 @@ def test_mutual_tls_server_with_wrong_cert(open_port):
...
-            assert (
-                server.global_train_servicer
-                and server.training_management_servicer
-                and enable_training
-            ) or (
-                server.global_train_servicer is None
-                and server.training_management_servicer is None
-                and not enable_training
-            )
+        error_context = (
```
gabe-l-hart, line 226: Nice, this is pretty slick!

tests/runtime/http_server/test_http_server.py
```diff
@@ -198,47 +200,59 @@ def test_mutual_tls_server_with_wrong_cert(open_port):
...
+            pytest.raises(CaikitRuntimeException) if error else nullcontext()
+        )
+        with error_context:
+            with runtime_http_test_server(open_port) as server:
+                # start a non-blocking http server with basic tls
```
gabe-l-hart, line 231: I know you didn't write it, but I don't _think_ this comment is accurate (or useful). Can you remove it?

tests/runtime/http_server/test_http_server.py
```diff
@@ -1060,6 +1074,87 @@ def test_single_models_info_ok(client, sample_task_model_id):
...
+        json_response = response.json()
+        assert response.status_code == 200, json_response
+        time.sleep(0.001)
+    assert json_response["state"] == "COMPLETED"
+
+    # Validate the results
+    response = client.get(f"/api/v1/task/sample/job/results", params={"job_id": id})
+    json_response = response.json()
+    assert response.status_code == 200, json_response
+    assert json_response["greeting"] == "Hello world"
```
gabe-l-hart, line 1108: Nice, I think this satisfies checking the actual content of the response

tests/runtime/http_server/test_http_server.py
```diff
@@ -1079,7 +1174,7 @@ def test_train_sample_task(client, runtime_http_server):
 
     # assert training response
     training_json_response = training_response.json()
-    assert training_response.status_code == 200, training_json_response
+    assert training_response.status_code == 202, training_json_response
```
gabe-l-hart, line 1177: This is what I'm worried about as a breaking change

tests/runtime/servicers/test_training_management_servicer.py
```diff
@@ -258,7 +258,7 @@ def test_training_status_incorrect_id(training_management_servicer):
 
     assert context.value.status_code == grpc.StatusCode.NOT_FOUND
     assert (
-        f"Unknown training_id: some_random_id" in context.value.message
+        f"Unknown background_id: some_random_id" in context.value.message
```
gabe-l-hart, line 261: I think these need to be reverted. If they are needed to get the tests to pass, I think the source error message needs to be reverted!

tests/runtime/test_grpc_server.py
```diff
@@ -322,6 +333,98 @@ def test_predict_sample_module_error_response(
...
+def test_job_predict_sample_module_cancel_request(
+    sample_task_model_id, runtime_grpc_server, sample_job_inference_service
+):
+    """Test that a grpc prediction job can be cancelled"""
+    # start a prediction that sleeps for a long time, so I can cancel
+    stub = sample_job_inference_service.stub_class(
+        runtime_grpc_server.make_local_channel()
+    )
+    predict_request = get_inference_request(SampleTask)(
+        sample_input=HAPPY_PATH_INPUT_DM, sleep_time=10
```
gabe-l-hart, line 387: Ah, this at least shouldn't introduce a big sleep since it gets cancelled

gabe-l-hart (changes requested):
A pile of NITs still (mostly copy-pasta) and a few structural things I still think need to be addressed to avoid API breaking changes:

1. Preserve kwarg names for `ModelFuture` to avoid breaking existing trainers
2. Don't switch 200 -> 202 (I kind of hate this one, but I think it's important)
3. Let's see if we can solve the `py-to-proto` duplicate import issue too!

caikit/core/model_manager.py
```diff
@@ -180,29 +188,102 @@ def train(
...
+                model_future.wait()
+
+        # Return a handle to the future
+        return model_future
+
     def get_model_future(
         self,
-        training_id: str,
-    ) -> ModelTrainerBase.ModelFutureBase:
+        training_id: str = None,
```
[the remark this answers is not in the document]
HonakerM: Uhhh great question! It should definitely be removed

caikit/runtime/servicers/global_predict_servicer.py
HonakerM: Not really since most of the boiler plate is in the tracing/future creation. They both have their own validation and content of args/kwargs

caikit/runtime/utils/servicer_util.py
HonakerM: Good call! I'm not sure I completely understand but I updated to reduce the delta

tests/core/helpers.py
HonakerM: Hmmm, TestModelFuture and TestJobFuture share a couple of implementations but they have enough different to where each class would need a good amount of overrides (like each having to still implement get_info, result, and save)

HonakerM: Good call! I just updated `ModelTrainerFutureBase`

caikit/interfaces/runtime/data_model/package.py
HonakerM: I was able to resolve it! Its actually a bug in the underlying protobuf code so I implemented a fix in caikit

tests/core/helpers.py
```diff
@@ -132,13 +137,17 @@ def __init__(self, parent, trained_model, save_path, save_with_id):
...
                 save_with_id=save_with_id,
             )
             self._parent = parent
             self._trained_model = trained_model
             self._canceled = False
             self._completed = False
+            self._save_path = save_path
+
+        @property
+        def save_path(self):
```
HonakerM, line 149: I think this change makes sense because `save_path` is really only relevant to the local model trainer but want to double check.

caikit/interfaces/runtime/data_model/prediction_job_management.py
```diff
@@ -0,0 +1,49 @@
...
+# Local
+from ....core.data_model import DataObjectBase, JobStatus, dataobject
+from .package import RUNTIME_PACKAGE
+
+
+@dataobject(RUNTIME_PACKAGE)
+class PredictionJobInfoRequest(DataObjectBase):
+    """DataModel to request information about a PredictionJob"""
+
+    job_id: Annotated[str, FieldNumber(1)]
```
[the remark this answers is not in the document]
HonakerM: That's a good point. I'll make the update. I originally thought of making it more generic but since training is specific predictions should be as well

caikit/interfaces/runtime/data_model/training_management.py
HonakerM: Since protobuf sends enums as their values and not types it won't break anything over the wire; however, it will break people who are accessing the enum via lookups (like `DataBase.get_class_by_name` or people doing type checks (though the mapping should help this case). IMO it's probably worth a v27

caikit/core/model_manager.py
```diff
@@ -180,29 +188,102 @@ def train(
...
+    ) -> JobPredictorFutureBase:
+        """Get the future handle to an in-progress prediction job
+
+        Args:
+            future_id (str): The ID string from the original prediction
+                submission's ModelFuture
+
+        Returns:
+            model_future (JobPredictorFutureBase): The future handle
+                to the job which holds the status of the in-flight prediction.
```
gabe-l-hart, line 275:
```suggestion
        Args:
            prediction_id (str): The ID string from the original prediction
                submission's ModelFuture

        Returns:
            prediction_future (JobPredictorFutureBase): The future handle
                to the job which holds the status of the in-flight prediction.
```

gabe-l-hart (changes requested):
ONE. MORE. NIT!

gabe-l-hart (approved):
Ship it! I agree that we should call this an API breaking change even though it's very minimal. We should spell out exactly what would break in the release notes.

PR #752, opened 2024-07-26
Fix bug when fetching model info of partially loaded model

HonakerM:
This PR fixes a minor bug where the model_info endpoint would try to read info from a model that was still being loaded. 

**If applicable**:

### DIFF
--- a/caikit/interfaces/runtime/data_model/info.py
+++ b/caikit/interfaces/runtime/data_model/info.py
@@ -59,15 +59,15 @@ class ModelInfo(DataObjectBase):
     """Information regarding a specific Model instance"""
 
     # Model information
-    model_path: Annotated[str, FieldNumber(1)]
+    model_path: Annotated[Optional[str], FieldNumber(1)]
     name: Annotated[str, FieldNumber(2)]
-    size: Annotated[int, FieldNumber(3)]
-    metadata: Annotated[JsonDict, FieldNumber(4)]
+    size: Annotated[Optional[int], FieldNumber(3)]
+    metadata: Annotated[Optional[JsonDict], FieldNumber(4)]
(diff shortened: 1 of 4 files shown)


HonakerM:
@gabe-l-hart this is a NIT pr but would be nice to have soonish ™️

gabe-l-hart (changes requested):
I think the issue here is a little bit more nuanced than this. The real issue is the line directly below here that calls `model_instance = loaded_model.model()` which blocks on loading. I think rather than just skipping non-loaded models, we should append a `ModelInfo` with some of the fields empty.

(Also tests are all failing, so we'll need to get those working before merging)

HonakerM:
> (Also tests are all failing, so we'll need to get those working before merging)

I misspelled a variable name 😅 

> I think rather than just skipping non-loaded models, we should append a ModelInfo with some of the fields empty.

I'm not sure I agree with this. The `ModelInfo` gets the majority of its info from the module instance and class so we'd only be able to fill in the name of the remote model

gabe-l-hart:
>I'm not sure I agree with this. The ModelInfo gets the majority of its info from the module instance and class so we'd only be able to fill in the name of the remote model

True, but this "just the model name plus non-loaded indicator" is very valuable for non-remote cases where a client needs to check and see if a model has been registered with the system, even if it's not done loading yet.

HonakerM:
@gabe-l-hart sounds good! I just added the response and updated the remote finder to ignore them

gabe-l-hart (approved):
Looks good!

gabe-l-hart:
Looks like tests are failing

HonakerM:
Rats! I thought I fixed those, just pushed a new commit 🤞

gabe-l-hart (approved):

Issue #754, opened 2024-08-05
Server streaming value check failures result in unknown grpc errors


evaline-ju:
## Describe the bug

On server streaming GRPC calls, failed value checks result in `Unknown` GRPC errors. It would be more helpful for clients to process `InvalidArgument` errors

## Sample Code

Basically following the sample_lib [README here](https://github.com/caikit/caikit/blob/main/examples/sample_lib/README.md) 
Running a sample server with
```
python3 -m examples.sample_lib.start_runtime_with_sample_lib
```

I created a model then purposely tried to invoke the `ValueError`

```
grpcurl -plaintext -H "mm-model-id: my-model" -d '{
  "sampleInput": {},
  "errStream": true
}' localhost:8085 caikit_sample_lib.SampleLibService.ServerStreamingSampleTaskPredict
```

This results in the error seen
```
Unknown (2)
Exception iterating responses: raising a ValueError
```

## Expected behavior

It would be more helpful if a GRPC `InvalidArgument` error were to be raised on failed value or type checks, to allow clients to better process errors

## Observed behavior

`Unknown` grpc errors are surfaced

## Additional context

The error handling in [build_proto_stream](https://github.com/caikit/caikit/blob/main/caikit/runtime/utils/servicer_util.py#L86) will have to be improved, since iterating over `caikit_library_response` raises an error. Similar error handling as the unary case (like with `context.abort`) will likely have to happen.

PR #757, opened 2024-08-07
:goal_net::bug: Handle server streaming grpc errors

evaline-ju:
Follow-up to #755 / #754 to handle other exception cases besides `ValueError` and `TypeError`. A case is seen in caikit-nlp when only `token_logprobs` is requested as True without `generated_tokens` or `input_tokens` set as True

This partially refactors the mapping of caikit core status codes to grpc codes so that this information can be reused whether on stream iteration or other errors (e.g. stream creation)

**If applicable**:


gkumbhat (approved):
left small nit, but otherwise lgtm

PR #761, opened 2024-08-27
Add support for setting a maximum session age in RemoteModuleBase

HonakerM:
This PR adds support for setting a maximum session/channel age in the RemoteModuleBase. This is primarily useful for grpc connections which only poll dns once on creation. This means that if you add a downstream server (like Kubernetes HPA) to the DNS pool existing grpc channels won't "discover" the new replicas until the pod has been restarted.

**If applicable**:

### DIFF
--- a/caikit/interfaces/common/data_model/remote.py
+++ b/caikit/interfaces/common/data_model/remote.py
@@ -146,6 +146,9 @@ class ConnectionInfo(DataObjectBase):
     retries: Optional[int] = 1
     # Runtime specific retry options
     retry_options: Optional[JsonDict] = field(default_factory=dict)
+    # Maximum age for a client channel. Values less then 0 are infinite while 0 means new
+    # channel/session for every request
+    max_session_age: Optional[float] = -1
 
     def __post_init__(self):
         """Post init function to verify field types and set defaults"""
@@ -172,6 +175,12 @@ class ConnectionInfo(DataObjectBase):
             port=self.port,
             timeout=self.timeout,
             retries=self.retries,
+        )
+        error.type_check(
+            "<COR730224567E>",
+            float,
+            int,
+            max_session_age=self.max_session_age,
         )
 
         if self.options:
--- a/caikit/runtime/client/remote_module_base.py
+++ b/caikit/runtime/client/remote_module_base.py
@@ -487,6 +487,7 @@ class RemoteModuleBase(ModuleBase):
             # If the current time is greater than the conn channel timeout then construct a
             # new channel
             elif datetime.now() - self._current_conn_time > self._max_conn_delta:
+                log.debug2("Creating new client channel due to max_session_age value")
                 self._current_conn_time = datetime.now()
                 self._conn_channel = construction_fn()
 
--- a/caikit/runtime/client/remote_module_base.py
+++ b/caikit/runtime/client/remote_module_base.py
@@ -18,8 +18,9 @@ design this class/factory does not use any references to the original Module cla
 """
 # Standard
 from collections import OrderedDict
+from datetime import datetime, timedelta
 from threading import Lock
-from typing import Any, Callable, Dict, Generator, List, Type, Union
+from typing import Any, Callable, Dict, Generator, List, Optional, Type, Union
 import copy
 import inspect
 import json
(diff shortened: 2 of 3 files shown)


caikit/runtime/client/remote_module_base.py
```diff
@@ -83,7 +84,9 @@ def __init__(
 
         # Configure GRPC variables and threading lock
         self._channel_lock = Lock()
-        self._conn_channel: Union[grpc.Channel, Session] = None
+        self._conn_channel: Optional[Union[grpc.Channel, Session]] = None
+        self._current_conn_time = None
+        self._max_conn_delta = timedelta(seconds=self._connection.max_session_age)
```
gabe-l-hart, line 89: Do we need a check on whether `max_session_age` is valid?

caikit/interfaces/common/data_model/remote.py
```diff
@@ -146,6 +146,9 @@ class ConnectionInfo(DataObjectBase):
     retries: Optional[int] = 1
     # Runtime specific retry options
     retry_options: Optional[JsonDict] = field(default_factory=dict)
+    # Maximum age for a client channel. Values less then 0 are infinite while 0 means new
+    # channel/session for every request
+    max_session_age: Optional[float] = -1
```
gabe-l-hart, line 151: Why is this marked `Optional` if the default value is not `None`? I think setting it to `None` would cause the `__post_int__` check to fail.

caikit/runtime/client/remote_module_base.py
```diff
@@ -424,54 +427,72 @@ def output_stream_parser():
...
+                self._current_conn_time = datetime.now()
+                self._conn_channel = construction_fn()
+
+            # If the max session age is less then 0 then reuse the existing con channel
+            elif self._connection.max_session_age < 0:
+                pass
+
+            # If the current time is greater than the conn channel timeout then construct a
+            # new channel
+            elif datetime.now() - self._current_conn_time > self._max_conn_delta:
```
gabe-l-hart, line 489: The `elif...pass` is a bit confusing. I'd suggest merging these into a single `elif`

tests/runtime/client/test_remote_model_initializer.py
```diff
@@ -521,3 +523,90 @@ def test_remote_initializer_retry(sample_task_model_id, open_port, protocol):
                 request_id=random_test_id(),
                 throw_first_num_requests=5,
             )
+
+
+@pytest.mark.parametrize("protocol", ["grpc", "http"])
+def test_remote_initializer_always_new_channel(
+    sample_task_model_id, open_port, protocol
+):
+    """Test to ensure RemoteModule Initializer negative max connection age always creates
```
gabe-l-hart, line 532:
```suggestion
    """Test to ensure RemoteModule Initializer zero max connection age always creates
```

tests/runtime/client/test_remote_model_initializer.py
```diff
@@ -521,3 +523,90 @@ def test_remote_initializer_retry(sample_task_model_id, open_port, protocol):
...
+    # Set random module_id so tests don't conflict
+    remote_config.module_id = random_test_id()
+
+    with runtime_test_server(open_port, protocol=protocol):
+        # Construct initializer and RemoteModule
+        remote_initializer = RemoteModelInitializer(Config({}), "test")
+        remote_model = remote_initializer.init(remote_config)
+        assert isinstance(remote_model, ModuleBase)
+
+        # Run RemoteModule Request twice to ensure the channel is retained. Run the second one with a small sleep to ensure the third request gets a new object
```
gabe-l-hart, line 596: We need a new definition of "small sleep!" I worry this is a little error prone with sleep-based behavior. Could we use a mock to patch over the `time.time` function to make it appear as "timed out" on the third call?

gabe-l-hart (changes requested):
One typo nit pick and a couple of suggestions

caikit/runtime/client/remote_module_base.py
HonakerM: Nope, that should be checked on `ConnectionInfo` creation

caikit/interfaces/common/data_model/remote.py
HonakerM: Good call! I updated it to `Union[float, int]` to be explicit about the types checked

caikit/runtime/client/remote_module_base.py
HonakerM: Good call, I was trying to be explicit but just made it more confusing

tests/runtime/client/test_remote_model_initializer.py
HonakerM: Good call on mock! Done

caikit/runtime/client/remote_module_base.py
gabe-l-hart: Ok, yeah, with the removal of `Optional` I agree that this is fine

gabe-l-hart (approved):
Looks great, ship it!

PR #762, opened 2024-09-04
:label: Add chunker types

evaline-ju:

"Chunkers" can be thought of as servers that “chunk” the target modality. For text, this can look like tokenizers or sentence splitters. ref. https://github.com/foundation-model-stack/fms-guardrails-orchestrator/blob/main/docs/architecture/adrs/001-orchestrator.md

This PR essentially "open-sources" the chunker stream result and chunker task, to enable chunker implementations at least with the chunker task.

The chunker task and stream result are very similar to the existing "tokenization" task and stream result, with the addition of tracking indices in the stream case. This helps track information in the case of what tokens/text (say from a text generation server) map to which sentences in a chunker result, if the chunker is a sentence splitter.

**If applicable**:

### DIFF
--- a/caikit/interfaces/nlp/data_model/text.py
+++ b/caikit/interfaces/nlp/data_model/text.py
@@ -70,5 +70,6 @@ class ChunkerTokenizationStreamResult(TokenizationStreamResult):
     # Below 2 represent pointer from the input stream.
     # These are different from start and processed_index index returned from
     # TokenizationStreamResult, which refers to the char span
-    input_start_index: Annotated[int, FieldNumber(5)]
-    input_end_index: Annotated[int, FieldNumber(6)]
+    # Field numbers 20 and 21 used in case parent classes have added fields
+    input_start_index: Annotated[int, FieldNumber(20)]
+    input_end_index: Annotated[int, FieldNumber(21)]


caikit/interfaces/nlp/data_model/text.py
```diff
@@ -59,3 +59,16 @@ class TokenizationStreamResult(TokenizationResults):
...
+@dataobject(package=NLP_PACKAGE)
+class ChunkerTokenizationStreamResult(TokenizationStreamResult):
+    """
+    Streaming tokenization result that provides pointer to the input chunk processed
+    """
+
+    # Below 2 represent pointer from the input stream.
+    # These are different from start and processed_index index returned from
+    # TokenizationStreamResult, which refers to the char span
+    input_start_index: Annotated[int, FieldNumber(5)]
```
gabe-l-hart, line 73: Small thought: It might be good to use a fully different set of field numbers for derived messages like this so that if additional fields are added to the parent, they don't need to be aware of all children. I'd suggest going with something like `20` and `21` or something like that.

gabe-l-hart (commented):
One little future-proofing thought on field numbers

caikit/interfaces/nlp/data_model/text.py
gkumbhat: good point.. otherwise, there may be cases where parent adds new field and now things are overlapping

gkumbhat (approved):
Looks good to me!

PR #763, opened 2024-09-04
Fix Failed Loaded Modules being considered Ready

HonakerM:
There is a bug in Caikit when a model failed to load but its `LoadedModel` instance is still marked as "loaded" because the `loaded()` check only considers if the future finished not if the model is actually loaded. This PR fixes this by adding an additional parameter `require_instance` which ensures that a local instance is actually available. 

**If applicable**:

### DIFF
--- a/caikit/runtime/model_management/loaded_model.py
+++ b/caikit/runtime/model_management/loaded_model.py
@@ -124,9 +124,10 @@ class LoadedModel:  # pylint: disable=too-many-instance-attributes
         @param required_instance (bool): If true, require that the model has loaded successfully
         
         @returns loaded (bool): Whether or not the model is finished loading
-        if require_instance:
-            return bool(self._model)
-        return bool(self._model or self._caikit_model_future.done())
+        return (
+            bool(self._model) or
+            (not require_instance and self._caikit_model_future.done())
+        )
 
     def wait(self):
         if self._model is None:
--- a/caikit/runtime/model_management/loaded_model.py
+++ b/caikit/runtime/model_management/loaded_model.py
@@ -118,7 +118,11 @@ class LoadedModel:  # pylint: disable=too-many-instance-attributes
         self.wait()
         return self._model
 
-    def loaded(self) -> bool:
+    def loaded(self, require_instance: bool = False) -> bool:
+        # If require_instance is true then require that the loaded model has a constructed instance
+        # attached
+        if require_instance:
+            return bool(self._model)
         return bool(self._model or self._caikit_model_future.done())
 
     def wait(self):
--- a/caikit/runtime/model_management/loaded_model.py
+++ b/caikit/runtime/model_management/loaded_model.py
@@ -119,8 +119,11 @@ class LoadedModel:  # pylint: disable=too-many-instance-attributes
         return self._model
 
     def loaded(self, require_instance: bool = False) -> bool:
-        # If require_instance is true then require that the loaded model has a constructed instance
-        # attached
+        """Determine if the model is has finished loading
+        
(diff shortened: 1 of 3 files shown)


caikit/runtime/model_management/loaded_model.py
```diff
@@ -118,7 +118,11 @@ def model(self) -> ModuleBase:
         self.wait()
         return self._model
 
-    def loaded(self) -> bool:
+    def loaded(self, require_instance: bool = False) -> bool:
+        # If require_instance is true then require that the loaded model has a constructed instance
+        # attached
```
gabe-l-hart, line 123:
```suggestion
        """Determine if the model is has finished loading
        
        @param required_instance (bool): If true, require that the model has loaded successfully
        
        @returns loaded (bool): Whether or not the model is finished loading
```

caikit/runtime/model_management/loaded_model.py
```diff
@@ -118,7 +118,11 @@ def model(self) -> ModuleBase:
         self.wait()
         return self._model
 
-    def loaded(self) -> bool:
+    def loaded(self, require_instance: bool = False) -> bool:
```
gabe-l-hart, line 121: Why is the default `False`? I guess really, why make this an argument at all? It seems like the old logic was just wrong? Did you trace history to figure out why `or self._caikit_model_future.done()` was added?

caikit/runtime/model_management/loaded_model.py
```diff
@@ -118,7 +118,11 @@ def model(self) -> ModuleBase:
...
+        # If require_instance is true then require that the loaded model has a constructed instance
+        # attached
+        if require_instance:
+            return bool(self._model)
         return bool(self._model or self._caikit_model_future.done())
```
gabe-l-hart, line 126: If we do keep this as an option:

```suggestion
        return (
            bool(self._model) or
            (not require_instance and self._caikit_model_future.done())
        )
```

caikit/runtime/model_management/loaded_model.py
HonakerM: I didn't trace history but logically it seems `loaded` used to see if the `ModelFuture` was generally completed not just successfully completed. When I checked caikit general it doesn't seem like we use `loaded` outside the servicer so assumed it was used in a subclass/other implementation.

---
After looking through the history it seems like it is only used in the info servicer. I'm game to change this to the default

gabe-l-hart (changes requested):
Ok, yeah, I looked through history and it looks like I added this all in one line, so that second clause was not added for a specific reason. The rest of the code seems to clearly indicate that `self._model` is only ever set to `self._caikit_model_future.result()`, so I would say let's just remove that second clause and not have this argument.

Also, linting and passing tests!

HonakerM:
> so I would say let's just remove that second clause and not have this argument.

Sounds good!

> Also, linting and passing tests!

I just committed your changes 👀

gabe-l-hart:
>I just committed your changes 👀

🤦 This is why I still don't really trust suggestions in GH. You can't see how long the lines are!

HonakerM:
@gabe-l-hart it looks like a test is failing because it was using `loaded()` to check if a job completed without calling `model()`. Any suggestions on a solution? Should I change the test or add a new function `completed()` to check if the job completed

gabe-l-hart:
>Any suggestions on a solution? Should I change the test or add a new function completed() to check if the job completed

Hmmm. Yeah, that sounds like the right option. I'd suggest calling it `load_finished` or something that makes it clear that it's just checking the status of the load operation and not the outcome.

gabe-l-hart (approved):
Ship it. Thanks Casey!

PR #765, opened 2024-09-10
Expose GRPC Concurrency Option and Update Retry

HonakerM:
This PR exposes another GRPC server option to allow for greater customization of deployment runtimes. This also updates the remote module connection to better retry requests when people use the above option.
 
**If applicable**:


HonakerM:
@gabe-l-hart not sure if this one needs new tests. It's just exposing an option of the underlying grpc server.

caikit/runtime/grpc_server.py
```diff
@@ -84,6 +84,7 @@ def __init__(
             thread_pool=self.thread_pool,
             interceptors=(PROMETHEUS_METRICS_INTERCEPTOR,),
             options=(self.config.runtime.grpc.options or {}).items(),
+            maximum_concurrent_rpcs=self.config.runtime.grpc.maximum_concurrent_rpcs,
```
gabe-l-hart, line 87: Interesting that this does not go into the `options` arg? Assuming this is correct, then this looks good to me

gabe-l-hart (approved):
Looks good according to the docs: https://grpc.github.io/grpc/python/grpc.html#grpc.server

caikit/runtime/grpc_server.py
HonakerM: Yeah its kinda annoying. TBH I thought this was an option for the longest time

PR #766, opened 2024-09-11
Fix regression in model trainer

HonakerM:
This Pr fixes a regression with caikit 0.27 where the save_path was no longer included in the model trainer future base.

**If applicable**:

### DIFF
--- a/caikit/core/model_management/model_trainer_base.py
+++ b/caikit/core/model_management/model_trainer_base.py
@@ -48,7 +48,6 @@ class ModelTrainerFutureBase(JobFutureBase):
         if "save_path" in kwargs:
             self.save_path = kwargs["save_path"]
         super().__init__(*args, **kwargs)
-        
 
     @abc.abstractmethod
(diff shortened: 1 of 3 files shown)


caikit/core/model_management/model_trainer_base.py
```diff
@@ -44,6 +44,9 @@ def __init__(self, *args, **kwargs):
             kwargs["future_name"] = kwargs["trainer_name"]
         if "training_id" in kwargs:
             kwargs["future_id"] = kwargs["training_id"]
+        # If save path is provided then add it as a property
```
gabe-l-hart, line 47: I think we might want to do this no matter what and just assign it to `None` given that the log line will fire on _any_ trainer?

HonakerM: That's a good point. The save_path should almost always be included in future construction but that does rely on the trainer implementations

gabe-l-hart (approved):
Thanks for the quick fix!

PR #772, opened 2025-07-10
replace removed class/function MessageFactory().GetPrototype()

ZacBranson:

The class MessageFactory() was previously deprecated in protobuf and has been removed entirely in new versions.  This PR replaces the use of MessageFactory().GetPrototype() with the newer GetMessageClass(), which should be a drop in replacement.

**If applicable**:

### DIFF
--- a/caikit/core/data_model/json_dict.py
+++ b/caikit/core/data_model/json_dict.py
@@ -145,4 +145,4 @@ def _get_message_class(
     """
     if hasattr(message_factory, "GetMessageClass"):
         return message_factory.GetMessageClass(desc)  # pragma: no cover
-    return message_factory.MessageFactory().GetPrototype(desc)  # pragma: no cover
+    return message_factory.GetMessageClass(desc)  # pragma: no cover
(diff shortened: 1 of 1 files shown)


caikit/core/data_model/json_dict.py
```diff
@@ -145,4 +145,4 @@ def _get_message_class(
     """
     if hasattr(message_factory, "GetMessageClass"):
         return message_factory.GetMessageClass(desc)  # pragma: no cover
-    return message_factory.MessageFactory().GetPrototype(desc)  # pragma: no cover
+    return message_factory.GetMessageClass(desc)  # pragma: no cover
```
gabe-l-hart, line 148: My only question is whether this is backwards compatible with protobuf 4.x. We recently removed support for 3.x, so if it works with 4.x, this should be good to go.

gabe-l-hart (commented):
One version compatibility check before we fully approve

caikit/core/data_model/json_dict.py
ZacBranson: It looks like the method was [added](https://github.com/protocolbuffers/protobuf/commit/c80e7efac72510a2bc3e9365520055f6d6656c1d#diff-ea4b1f4d75acdd7bb9e6f0ec31ef7bebcb5cd5b0a6ea3bb782992d42a438fbf3) on Jan 13, 2023 and was included in [v22.0](https://github.com/protocolbuffers/protobuf/releases/tag/v22.0), which would be [version 4.22.0](https://pypi.org/project/protobuf/4.22.0/) of the python library.

So 4.21.x would not work

